PR #29, opened 2021-02-17
Finished Schedule Component(list View)

thesakshi:


chenaaron3:
Nice job! The code for the component look great. I noticed you commented out code for fetching the schedule data. Did you have trouble with that?

thesakshi:
> Nice job! The code for the component looks great. I noticed you commented out code for fetching the schedule data. Did you have trouble with that?

Hi, 
No, I did not have any problem with that code. That code was already there when I started working on the component and I assumed that it was/will be required when we use actual data from the API and therefore, I did not remove it.

chenaaron3 (commented):
Looks good. Just try to avoid the !important flags if possible. Next task is to connect to the api.

chenaaron3 (approved):
Approve merge.

PR #68, opened 2021-05-15
Planner stats

iabbaali:
# Description
- Display course and unit count for the overall planner
- Display course and unit count for each year
- Display unit count for each quarter

# Demo

Screen Shot 2021-05-14 at 5 32 09 PM

Video Walkthrough


trishaprile (approved):
Perfect!!

PR #114, opened 2022-01-25
Bump `node-sass` version to be compatible with Node 16+

ryanmohta:
`npm install` for the `site` directory didn’t complete on Node 16.x, but worked fine on Node 14.x. Turns out the issue was the version of the `node-sass` module we were using was only compatible up to Node 14! (image source: [node-sass npm page](https://www.npmjs.com/package/node-sass))

Screen Shot 2022-01-24 at 16 59 32

I bumped the module to the current latest version (7.0.1), and ran the website to make sure changing the version didn’t break anything. Thankfully it didn’t!

UPDATE: turns out the `node-sass` module is deprecated and didn’t pass the GitHub Actions checks for AWS and Heroku, so I replaced it with the newer `sass` module per [this stackoverflow post](https://stackoverflow.com/questions/70281346/node-js-sass-version-7-0-0-is-incompatible-with-4-0-0-5-0-0-6-0-0)

### DIFF
--- a/site/package.json
+++ b/site/package.json
@@ -21,7 +21,6 @@
     "js-cookie": "^2.2.1",
     "lodash": "^4.17.21",
     "moment": "^2.24.0",
-    "node-sass": "^7.0.1",
     "react": "^16.13.1",
     "react-beautiful-dnd": "^13.0.0",
     "react-bootstrap": "^1.5.1",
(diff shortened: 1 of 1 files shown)


chenaaron3 (approved):
looks good to me!

PR #122, opened 2022-02-04
Peters Roadmap - Add Year Popup Default New Year to Last Year

### LINKED ISSUE #98: Default New Year to Last Year on Roadmap
- [x] If there's no existing year, default to current year
- [x] If there's years on the roadmap, default to the last year + 1

ethanwong16:
- Closes #98 

This PR updates Peter's Roadmap's "add year" popup to have a more intuitive placeholder year in the form.  
If users don't have any currently existing years in their planner, the placeholder year defaults to the current year.  Otherwise, the placeholder year is the year following the bottom-most existing year in the user's planner.

This PR also makes minor updates to the roadmapSlice, switching the delete year mechanism to replace usage of `del` with `splice`.  

This PR also updates the years so that they start from 1 instead of 0.

## Description

Changes inside `AddYearPopup.tsx`
- Added props to `AddYearPopup` to take in the default placeholder year
- Added `useEffect` hook to set the default placeholder year every time the default year changes

Changes inside `Planner.tsx`
- Calculate the default year using stored years from `roadmapSlice` and pass to `AddYearPopup`
- If there are no years, pass a new `Date` object representing the current year
- Otherwise, get the last year, increment it by 1, and pass to `AddYearPopup`

Changes inside `roadmapSlice.ts`
- Replace `del` operator with `splice`, as del may not shift elements after deleting and instead leave a null value

Changes inside `Year.tsx`
- Display years starting from 1 instead of 0

## Screenshots

|before|before|
|--|--|
|after|after|

## Todos:
- [ ] Write tests
- [ ] Write documentation


chenaaron3 (approved):
lgtm :D Nice job!

PR #127, opened 2022-02-07
Added favicons for other platforms

patrickanguyen:
# Added favicons for other platforms

## Description
Created favicons for other platforms using https://realfavicongenerator.net/

## Screenshots

No change (or noticeable change) between new and old desktop favicon
image
Left is new and right is old

- Use https://realfavicongenerator.net/favicon_checker to verify that the favicon works on all major platforms

## Notes
- Previously the `href` to the favicon was prepended with `%PUBLIC_URL%/`, I am not sure if this is required since it seems to work locally on my machine.


chenaaron3 (approved):
lgtm 👍

PR #149, opened 2022-02-17
removes duplicate voting for users

AyushSat:
Removes the ability for logged in users to vote on the same review multiple times
## Description
Added a new database entry that stores all review votes and checks any new votes against that data to see if duplicates are occurring.
## Screenshots
More backend, no screenshots.

## Todos:
- Implement colored buttons to indicate already voted


chenaaron3 (approved):
lgtm 💯

PR #150, opened 2022-02-17
Report Reviews

advayanand:
This PR consisted of implementing a report function for every review, where users can report reviews that they feel are harmful or should not belong on the platform. There is also a reports page where admins can view all reports that have been submitted, and choose to accept or deny reports.

## Description

Backend
- Implemented a /reports endpoint in the backend with methods to get all reports, submit a report, and delete a report
- Modified the GET /reviews endpoint to also take in the review ID as a query filter to allow querying reviews by ID
- Implemented a DELETE /reviews endpoint to allow deleting a review

Frontend
- Added a button that users can click to report a review
- Added a report form where users can submit a report along with a reason for reporting
- Added a page where admin users can view all reports that have been submitted and accept or deny reports.
  - Accepting a report deletes both itself and the corresponding review from the database, while denying a report only deletes the report.

## Screenshots

Screen Shot 2022-02-16 at 11 51 52 PM

Screen Shot 2022-02-16 at 11 52 54 PM

https://user-images.githubusercontent.com/22901073/154440898-4139ba35-38c8-4847-9124-a0f29b50ef77.mov

## Todos:
- [ ] Write tests
- [ ] Write documentation
- [ ] Currently all users can view the reports page; this needs to be restricted to only admin users
- [ ] Fix styling on the report form and reports page

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -42,6 +42,7 @@ jobs:
             SESSION_SECRET=$SESSION_SECRET\n \
             GOOGLE_CLIENT=$GOOGLE_CLIENT\n \
             GOOGLE_SECRET=$GOOGLE_SECRET\n \
+            ADMIN_EMAILS=$ADMIN_EMAILS\n \
             PRODUCTION_DOMAIN=$PRODUCTION_DOMAIN\n \
             PORT=$PORT" > .env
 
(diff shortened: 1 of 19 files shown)


chenaaron3 (approved):
Really nice work Advay! Glad you were able to make all these components/endpoints without much assistance. I went ahead and fixed up the isAdmin endpoint you were waiting on so only certain users can access the admin page. Overall really solid work 💯!

PR #169, opened 2022-02-20
Removed about page and files relating to it

patrickanguyen:
# Removed about page and files relating to it

## Description
The About page was removed from the footer in #164 and there are no more references to the About page anymore. However, the About page is still accessible through https://peterportal.org/about. 

This PR removes all files relating to the About page and removes the route to it, so it redirects to a 404 instead. 

Also, can we please change the anteater picture on the 404 page 🙏? It is so out of place compared to the rest of the website and I despise it so much. 

### Before
image

### After
image

- Go to https://peterportal.org/about and verify that it redirects to a 404


chenaaron3:
HAHA yeah I just grabbed a random anteater pic from Google 😂. You got any replacement suggestions?

chenaaron3:
working
They used this one in figma.

patrickanguyen:
Just anything than the anteater 🙏 Even removing the image entirely would be better imo

chenaaron3 (approved):
lgtm

PR #203, opened 2022-03-08
Ayush

AyushSat:
Added functionality to undo votes on reviews, colored vote buttons if user has already voted on the review

## Description
Added a new patch to the review router to get the color for a specific vote for a review and user, used the useEffect React hook to constantly update the className for each vote button.

## Screenshots

Before:
image
After:
image

Go to any course(course used to test COMPSCI 143A) and vote on any of the test reviews there, see how the buttons color when you vote and click again to remove the vote.


api/src/controllers/reviews.ts
```diff
@@ -181,6 +192,35 @@ router.patch("/vote", async function (req, res) {
   }
 });
 
+/**
+ * Get whether or not the color of a button should be colored
+ */
+router.patch("/getVoteColor", async function (req, res) {
```
chenaaron3, line 198: I would make this return a dictionary that provides the votes of a given user (map reviewID to vote status boolean). That way we don't need to make a request for each subreview. Only 1 request will be needed per page to see all vote colors. Could possibly take in a list of reviewIDs too so we don't relay irrelevant reviews.

site/src/component/Review/SubReview.tsx
```diff
@@ -16,12 +16,19 @@ interface SubReviewProps {
...
   const [cookies, setCookie] = useCookies(['user']);
+  const [upvoteClassname, setUpvoteClass] = useState("upvote");
+  const [downvoteClassname, setDownvoteClass] = useState("downvote");
 
   const voteReq = async (vote: VoteRequest) => {
     const res = await axios.patch('/reviews/vote', vote);
     return res.data.deltaScore;
   }
 
+  const getColor = async (vote: VoteColorRequest) => {
```
chenaaron3, line 27: Instead of making an API call in each subreview, we should make that 1 API call in Reviews Component and pass the vote status as a prop.

chenaaron3 (approved):
Implementation works, but not optimal. We don't want to make too many API requests.

api/src/controllers/reviews.ts
```diff
@@ -220,6 +220,44 @@ router.patch("/getVoteColor", async function (req, res) {
...
 
+/**
+ * Get multiple review colors
+ */
+router.patch("/getVoteColors", async function (req, res) {
+  if (req.session.passport != null) {
+    //query of the user's email and the review id
+    let ids = req.body["ids"];
+    let colors = [];
+    for (let i = 0; i < ids.length; i++) {
```
chenaaron3, line 231: You can use the [$in](https://www.mongodb.com/docs/manual/reference/operator/query/in/) operator to do this in one query.

chenaaron3 (commented):
Just 1 feedback on optimizing the query, but everything else looks good!

chenaaron3:
Also, can you rebase with master and resolve the merge conflicts?

PR #220, opened 2022-05-03
Sidebar now closes when switching tabs

ryanmohta:
As the title says, the sidebar now closes when switching between the individual tabs like **Catalog**, **Roadmap**, and the admin views! I used our existing `toggleMenu` function to accomplish this, but renamed it `closeSidebar` so its purpose is more clear (the function is only used once currently and for this purpose).

I tested it out by clicking between all of the sidebar options both when the sidebar was open and collapsed, and it worked in all cases.


chenaaron3 (approved):
lgtm

PR #226, opened 2022-05-22
New Search Component Optimization

advayanand:
## Description

- Optimized search so that it only updates results when user stops typing for 1 second

## Screenshots

https://user-images.githubusercontent.com/22901073/169672603-12ca1348-906c-4867-b489-0e15be4f6523.mov

## Todos:
- [ ] Write tests
- [ ] Write documentation


site/src/component/SearchModule/SearchModule.tsx
```diff
@@ -22,6 +23,7 @@ const SearchModule: FC<SearchModuleProps> = ({ index }) => {
     const dispatch = useAppDispatch();
     const courseSearch = useAppSelector(state => state.search.courses);
     const professorSearch = useAppSelector(state => state.search.professors);
+    const [ pendingRequest, setPendingRequest ] = useState<NodeJS.Timeout | null>(null);
```
chenaaron3, line 26: We are not rendering anything based on the pending request, so this variable does not have to be stateful. We can just instantiate it as a normal variable.

site/src/component/SearchModule/SearchModule.tsx
```diff
@@ -13,6 +13,7 @@ import { searchAPIResults } from '../../helpers/util';
 import { SearchIndex, BatchCourseData, CourseGQLResponse, ProfessorGQLResponse, BatchProfessorData } from '../../types/types';
 
 const PAGE_SIZE = 10;
+const SEARCH_TIMEOUT_MS = 1000;
```
chenaaron3, line 16: Lets change this to 500. I feel like 1000 feels a bit laggy.

chenaaron3 (changes requested):
Added comments

chenaaron3 (approved):
Lgtm

PR #247, opened 2022-12-16
Make project compatible with Node 18

ryanmohta:
After some research, I realized the issue was an [SSL loophole](https://stackoverflow.com/questions/69692842/error-message-error0308010cdigital-envelope-routinesunsupported) that was closed in Node 17+, resulting in Node 17+ throwing an error when Node 16 was fine. Running `npm audit fix` (as one of the answers suggested) indicated the problem was with the `react-scripts` dependency; upgrading it to the latest version fixed the issue.

I also ensured that no functionality to the site was impacted by the change.


ryanmohta:
Converted to draft since I discovered an issue with the API on Node 18; will reopen once that’s fixed

ryanmohta:
Force-pushed since I rebased from `main` after merging #248

ryanmohta:
Another force push after rebasing to incorporate our GitHub Actions hot-fix

ryanmohta:
Opening this pull request for review now that I’ve (hopefully) fixed all the upgrade issues!

To fix the SSL loophole above, I had to upgrade the `react-scripts` dependency from version 4 to 5; however, this caused an error, “[Invalid options object. Dev Server has been initialized using an options object that does not match the API schema](https://stackoverflow.com/questions/70374005/invalid-options-object-dev-server-has-been-initialized-using-an-options-object).” After examining multiple possible solutions, the one I settled on was adding the `http-proxy-middleware` package, which essentially proxies GET, POST, and other requests to our backend url (`localhost:5000`). This required me to add a `/api` to the front of all paths for our requests; while this may be a minor inconvenience upfront, I think it will improve code quality going forward since it makes clear which URLs are handled by the back-end.

I also updated the GitHub Action to use Node 18 and a simpler installation process; not sure if we’d need to change anything on AWS to use the new Node version.

ryanmohta:
Added the `/api` prefix to all routes accessed by a get/post request, so the routes should work on both dev and prod now

ethanwong16 (commented):
LGTM! Great job Ryan 🥳 !!!

Will wait until staging instances are all set up before physically approving the PR, since there's a lot of configuration changes here that would be good to test out first on a staging environment.

We should also create an issue to note that we may revert back to using proxy : 5000 instead of `http-proxy-middleware` once support is added for Node18

ethanwong16:
Resolves #256

ethanwong16 (approved):
LGTM - approving now

PR #248, opened 2022-12-19
Replace usages of the `websoc-api` library with calls to the PeterPortal API

ryanmohta:
This PR updates the schedule endpoints of the project’s backend to make calls to the PeterPortal API instead of using the `websoc-api` library. Since the API already calls the library internally (and in the future, with caching), this should help improve the performance of the backend and encourage separation of concerns.

This will also unblock #247 (Make project compatible with Node 18) since the version of the `websoc-api` library we were using in the project caused problems with Node 18.


ethanwong16 (approved):
LGTM 🙌!  
Might be an option to keep the old call directly to websoc-api as a backup, but ultimately would be fine either way :)

Issue #244, opened 2022-12-01
Week Number Not Correct


aahei:
On the upper-right corner of the page, it displays the week number of the quarter. 
The week number is not correct. It is Week 10 on December 1st.
![image](https://user-images.githubusercontent.com/18681247/205177678-df5e876d-4c54-48c2-8511-3ebf525a5055.png)

js0mmer:
Special case with fall quarter. Instruction start date is Thursday 9/22/2022, week 0. It is currently Thursday 12/1/2022 so the code calculates it to be week 11 already.

PR #254, opened 2023-02-08
Fix incorrect week number calculation for fall quarter and time zone issues

js0mmer:
## Description
Adjusted quarter begin date for fall quarters to Monday rather than Thursday within the findWeek function in api/src/helpers/week.ts and subtracted 1 to start the counter at week 0 for fall. Added [dayjs](https://day.js.org/en/) and used it to handle time zone issues by creating dates in PST/PDT even if the server is in another time zone. Added fix to adjust any quarter begin/end dates received from the mongo cache in UTC+0 to be in PST/PDT. Fixed processRow function to force quarter begin/end dates to be created and cached in PST/PDT moving forward. Previously the current week would increment 7 or 8 hours earlier than it should have due to dates not being created in the right time zone. Fixes issue #244 

Change your system date/time and check the week number.

## Todos:
- [ ] Write tests
- [ ] Write documentation


api/src/helpers/week.ts
```diff
@@ -49,24 +61,59 @@ export function getWeek(): Promise<WeekData> {
...
+        });
+
+        let isFallQuarter = false;
+        // in fall quarter, instruction start date is not on a monday
+        // it is on a thursday in week 0
+        // let's move it back to monday in week 0 and adjust our week calculations by -1
+        // that way week calculations are correct for fall quarter
+        if (begin.day() !== 1) {
+            isFallQuarter = true;
+            begin = begin.day(1); // moves day back to Monday (monday is index 1)
```
ethanwong16, line 102: Alternative is to just always set `begin` to Monday, but leaving as is works too!

api/src/helpers/week.ts
```diff
@@ -49,24 +61,59 @@ export function getWeek(): Promise<WeekData> {
...
+
+        // adjust instruction end date to last ms of the day
+        end = end.add({
+            hours: 23,
+            minutes: 59,
+            seconds: 59,
+            ms: 999
+        });
+
+        let isFallQuarter = false;
```
ethanwong16, line 95: It might be more explicit to use the `quarter` variable to check if it's currently Fall or not, but using the start date of Monday is fine too.

api/src/helpers/week.ts
```diff
@@ -49,24 +61,59 @@ export function getWeek(): Promise<WeekData> {
...
+        // in fall quarter, instruction start date is not on a monday
+        // it is on a thursday in week 0
+        // let's move it back to monday in week 0 and adjust our week calculations by -1
+        // that way week calculations are correct for fall quarter
+        if (begin.day() !== 1) {
```
ethanwong16, line 100: It would be good to set a single ENUM called Monday == 1
https://www.typescriptlang.org/docs/handbook/enums.html

js0mmer:
Does not work as intended on production. Week number seems to be off by 1 on 2/27/2023. Displays week 7 on prod despite it showing week 8 (correctly) on dev.
Update at 4PM: Now shows correct week. Seems to be 16 hours late. Have to investigate dayjs, seems to not be keeping the time zone in pacific as expected

PR #258, opened 2023-02-16
Sticky search popup sidebar and search results

### LINKED ISSUE #250: Adapt the sidebar display to be "sticky"
Look into making the sidebar display always "stick" to the page so it appears no matter where the user is viewing the page (scrolling).

After I click the course I'd like to see more information about... (there is no visual indication I've clicked that course card or the sidebar opened up):
image

When I scroll back up (the sidebar display is there):
image

js0mmer:
## Description
Added a constrained height to search-popup class and `overflow-y: auto;` for scrolling. Added constrained height and hidden overflow to search-list id for a fixed search bar and scrollable search result container. Fixes #250 and #259. Another issue we might want to open/change we might want to make is make the scroll bar go back to the top after the results update. As you can see in the after gifs, it remains in the same spot and you have to manually scroll back up to see the first result. This is also the same behavior on the course search for Peter's Roadmap since they use the same component.

## Screenshots
Before:
before
After:
after

## Final Checks:
- [ ] Delete branch

(optional)
- [ ] Write tests
- [ ] Write documentation

### DIFF
--- a/site/src/pages/SearchPage/SearchPage.scss
+++ b/site/src/pages/SearchPage/SearchPage.scss
@@ -12,6 +12,10 @@
 
     #search-list {
         width: 50vw;
+        display: flex;
+        flex-direction: column;
+        height: 85vh;
+        overflow-y: hidden;
     }
 
     #search-popup {
@@ -19,6 +23,15 @@
         height: fit-content;
         position: sticky;
         top: 0;
+    }
+
+    .search-hit-container {
+        margin-top: 2vh; // use margin instead of padding so the scroll bar isn't offset/above the first hit item
+        padding-top: 0;
+
+        .hit-item:last-child {
+            margin-bottom: 0; // so scroll bar doesn't extend past the last hit item
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


ethanwong16 (approved):
LGTM - great job @js0mmer  🥳!
Also adding @ryanmohta as a reviewer since he has more expertise with CSS

site/src/pages/SearchPage/SearchPage.scss
```diff
@@ -12,11 +12,26 @@
...
+        flex-direction: column;
+        height: 85vh;
+        overflow-y: hidden;
     }
 
     #search-popup {
         flex-grow: 1;
         height: fit-content;
+        position: sticky;
+        top: 0;
```
ryanmohta, line 25: Now that both the search list and popup have scrollbars within the containers, the popup itself doesn’t move anymore! Because of that, we no longer need to ensure the popup stops at the very top when scrolling down on the page (which is what these two lines accomplish).

I played around with the webpage after disabling these two properties in the web inspector, and the behavior’s exactly the same without them.
```suggestion
```

ryanmohta (commented):
Good work Jacob! I left one optional suggestion that fixes what I think is a remnant from when we previously scrolled on the entire page instead of within the search list and search popups.

Also, I noticed your branch isn’t up to date with the main branch, so make sure you update your branch with the rebase option to ensure you have all the latest changes everyone’s added. If you run into a conflict let me know and I can help fix it!

ryanmohta (approved):

Issue #255, opened 2023-02-10
Deploy staging instances when pull requests are opened/updated


ryanmohta:
Taking inspiration from AntAlmanac; whenever a pull request is opened in the AntAlmanac repo, a GitHub action builds the project and deploys it to `staging-[pull request #].antalmanac.com`. We already have a GitHub action in place to build PeterPortal, so I’d imagine to accomplish this, we’d have to add a staging deployment step to the action and add some configuration to AWS.

PR #261, opened 2023-02-23
Deploy staging instance on PR

ryanmohta:
This PR updates the GitHub Actions workflow to deploy a staging instance when a PR is created or updated. This staging instance will be hosted at `staging-{ pr number }.peterportal.org`.

To accomplish this, I had to dynamically create the `DOMAIN` environment variable to include the staging prefix if the source of the workflow run was a pull request. This environment variable would then be used in the `api/serverless.yaml` file to indicate which domain to deploy the project to. All credit for this process goes to @Voark who provided guidance on how to accomplish this!

I haven’t gotten to test this out yet, as the only way to do so is to actually make the PR. I’ll keep making changes if it doesn’t happen to be correct, and only request reviews once I believe it’s ready to be merged.

### To-do
- [x] Add comment indicating to the user what URL the app was deployed to
- [x] Clean up the stack when the PR is closed
- [x] Add concurrency when multiple commits to a PR are made in quick succession

### DIFF
--- a/.github/workflows/clean-up-pr.yml
+++ b/.github/workflows/clean-up-pr.yml
@@ -6,6 +6,10 @@ on:
     branches:
       - master
 
+concurrency:
+  group: ${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   clean-up-pr:
     runs-on: ubuntu-latest
@@ -19,12 +23,14 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Dump Environment Variables
+        id: dump-env-variables
         run: |
           cd api && printf "STAGING_PREFIX=$STAGING_PREFIX\n \
             STAGE_NAME=$STAGE_NAME" > .env
 
       - name: Remove Serverless stack
         uses: serverless/github-action@v3.1
+        if: ${{ success() || ( cancelled() && steps.dump-env-variables.outcome == 'success' ) }}
         with:
           args: >
             -c "cd api && 
--- a/.github/workflows/clean-up-pr.yml
+++ b/.github/workflows/clean-up-pr.yml
@@ -13,8 +13,8 @@ jobs:
   clean-up-pr:
     runs-on: ubuntu-latest
     env:
-      STAGING_PREFIX: ${{ github.event_name == 'pull_request' && format('staging-{0}.', github.event.pull_request.number) || '' }}
-      STAGE_NAME: ${{ github.event_name == 'pull_request' && format('staging-{0}', github.event.pull_request.number) || 'prod' }}
+      STAGING_PREFIX: staging-${{ github.event.pull_request.number }}.
+      STAGE_NAME: staging-${{ github.event.pull_request.number }}
 
     steps:
       - name: Check Out Repo
--- a/.github/workflows/clean-up-pr.yml
+++ b/.github/workflows/clean-up-pr.yml
@@ -7,14 +7,12 @@ on:
       - master
 
 concurrency:
-  group: ${{ github.head_ref || github.run_id }}
-  cancel-in-progress: true
+  group: deploy-${{ github.head_ref || github.run_id }}
 
 jobs:
   clean-up-pr:
     runs-on: ubuntu-latest
     env:
-      # need to set to false, otherwise react warnings will be treated as errors
       STAGING_PREFIX: ${{ github.event_name == 'pull_request' && format('staging-{0}.', github.event.pull_request.number) || '' }}
       STAGE_NAME: ${{ github.event_name == 'pull_request' && format('staging-{0}', github.event.pull_request.number) || 'prod' }}
 
@@ -23,14 +21,12 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Dump Environment Variables
-        id: dump-env-variables
         run: |
           cd api && printf "STAGING_PREFIX=$STAGING_PREFIX\n \
             STAGE_NAME=$STAGE_NAME" > .env
 
       - name: Remove Serverless stack
         uses: serverless/github-action@v3.1
-        if: ${{ success() || ( cancelled() && steps.dump-env-variables.outcome == 'success' ) }}
         with:
           args: >
             -c "cd api && 
--- a/.github/workflows/clean-up-pr.yml
+++ b/.github/workflows/clean-up-pr.yml
@@ -1,0 +1,37 @@
+name: Clean up PR
+
+on:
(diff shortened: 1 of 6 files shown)


ryanmohta:
Closing this PR to test the “remove the stack when PR is closed/merged” functionality

ryanmohta:
Ready for prime-time (I hope)! Marking this as ready for review and sending out reviews

api/serverless.yaml
```diff
@@ -21,7 +20,8 @@ provider:
   name: aws
   runtime: nodejs14.x
   region: us-east-1
-  stage: prod
+  stage: ${env:STACK_NAME}
+  stackName: peterportal-${env:STACK_NAME}
```
Voark, line 24: Shouldn't this be `peterportal-core-${env:STACK_NAME}`? Not entirely sure how serverless works but this is based off the current stack name I see for your prod stack.

Voark: Otherwise when you deploy to prod you'll make a new stack instead of updating the current one.

Voark (commented):
Besides my comment, LGTM!

Also for some inexplicable reason the first few times I opened the staging site on my current browser (Arc) I got an error in the console, but it was working fine on other browsers (FF, Chrome). I just went back to try and screenshot the error but now it works fine on Arc too. Not a big deal just wanted to mention it...

api/serverless.yaml
ryanmohta: Oh that’s right, I’ll go ahead and update that!

ryanmohta:
Ah you bring up a good point about concurrency, I’ll look more into that and see if there’s a way to have the second action cancel the first one if both of them run simultaneously

Voark:
> Ah you bring up a good point about concurrency, I’ll look more into that and see if there’s a way to have the second action cancel the first one if both of them run simultaneously.
> 
> Edit: I looked at the [GitHub docs](https://docs.github.com/en/actions/using-jobs/using-concurrency) and there’s a couple ways to handle concurrency: waiting for the first job to finish before starting the second one, or canceling the first job in-progress and starting the second one immediately. Which do you think works better for our deployments?

For AntAlmanac we cancel the current build and start a new run. However this introduces a new issue where if the previous run already started deploying, it can't be stopped on a whim so there is a [step](https://github.com/icssc/AntAlmanac/blob/3b236ff59d506795d264b9f18152486ff2889a19/.github/workflows/stage_pr.yaml#L60) that waits for the stack resources to finish updating (if they are). I added this because if you did back to back to back commits you wouldn't build up an action queue, but it does add some complexity.

IMO, I would do cancel in progress. If deploying with Serverless automatically waits for the stack to stabilize it should work well, if not feel free to copy paste our [stabilize action](https://github.com/icssc/AntAlmanac/blob/main/.github/actions/stabilize_stack/action.yaml), should be plug and play.

api/serverless.yaml
ryanmohta: Just fixed this so I’ll resolve this conversation

ryanmohta:
Just added concurrency; I didn’t see any way to stop the serverless deployment while it was in progress, so I added a part to the GitHub Action to keep running the deployment step if the job was canceled. This way, if commit B is pushed before commit A gets to the deployment step, commit A gets canceled, but if commit A is currently deploying, it finishes the deployment while the GitHub Action for commit B starts. Even though this will lead to slightly overlapping GitHub Action runs, based on how long our build process takes it won’t lead to overlapping _deployments_.

Voark:
> Just added concurrency; I didn’t see any way to stop the serverless deployment while it was in progress, so I added a part to the GitHub Action to keep running the deployment step if the job was canceled. This way, if commit B is pushed before commit A gets to the deployment step, commit A gets canceled, but if commit A is currently deploying, it finishes the deployment while the GitHub Action for commit B starts. Even though this will lead to slightly overlapping GitHub Action runs, based on how long our build process takes it won’t lead to overlapping _deployments_.

Yeah there is no way to stop the an aws stack from being deployed, which is why in our action I wait for the deployment to finish before trying a new one. Also forgot to mention you probably also want to add concurrency to the destroy action otherwise you will have to manually delete stacks if it runs while your stack is currently in deployment.

ryanmohta:
Closing PR once again for testing

ryanmohta:
Made some updates to how concurrency is handled: to make it easier to isolate the deployment step from the rest of the workflow (and in good coding practice) I’ve split the workflow into a _build_ job and a _deploy_ job, each with their own concurrency group. If two builds jobs overlap, the second job will cancel the first, while if two deploy jobs overlap, the second job will wait for the first one to finish. This is so that we don’t accidentally cancel in the middle of a deployment, which could corrupt the site.

I made use of the GitHub Actions cache to pass the built files from the build job to the deploy job, and ensured uniqueness between different caches by assigning keys of `${{ github.workflow }} #${{ github.run_number }} attempt ${{ github.run_attempt }}` (for example, `Build and deploy #9 attempt 1`). This way the deployment can use the same cache as the build for the same workflow run, but deployments or builds from two different workflow runs (or even two different attempts of the same run) will get their own cache.

The "Clean up PR” workflow uses the same concurrency group as the deploy job, since they both make changes to AWS. Deployments and clean-ups will run one after the other instead of one canceling the other, just to be safe.

ryanmohta:
Re-requesting a review from @Voark since I fixed all the changes you suggested!

Voark (approved):
Just a small comment on the action, otherwise LGTM! 

I do think the infrastructure should still be looked into to make deploying and building easier/more efficient (obviously out of scope of this PR), I'll open an issue for it sometime.

.github/workflows/clean-up-pr.yml
```diff
@@ -0,0 +1,39 @@
...
+      - master
+
+concurrency:
+  group: deploy-${{ github.head_ref || github.run_id }}
+
+jobs:
+  clean-up-pr:
+    runs-on: ubuntu-latest
+    env:
+      STAGING_PREFIX: ${{ github.event_name == 'pull_request' && format('staging-{0}.', github.event.pull_request.number) || '' }}
```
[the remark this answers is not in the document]
ryanmohta: good catch, must be me copy-pasting from the build and deploy action and forgetting to change this here. I’ll go fix this now!

ethanwong16:
Closes #255

ethanwong16 (approved):
LGTM 🚀 🚀 🚀 🚀 🚀 !
Awesome job @ryanmohta  and thanks @Voark for the help!

PR #263, opened 2023-02-24
Scrollable transfer tab

MFYLM:
## Description

Create a scrollable component to provide a more intuitive way for users to type in missing courses on transfer tab

## Screenshots

Display missing courses for current plan with horizontal scroll component
image

After typing correct course name, course will be removed from the list
image

## Final Checks:
- [ ] Delete branch

(optional)
- [ ] Write tests
- [ ] Write documentation


MFYLM:
For autocomplete, MaterialUI has a great component to complete that functionality. Maybe worth to take a look

Issue #194, opened 2022-03-02
Add a year as default for a new roadmap


chenaaron3:
If there is no saved roadmap, add a year as default so users know they can drag and drop courses. Roadmap looks intimidating with nothing initially there.

PR #264, opened 2023-02-24
Add default year to the roadmap

SenghoungLim:
## Description
Create a new function
Compare if the length of the data array is 0
If so, add the new Default
        (Thanks Ryan for Helping me out)
        
## Screenshots
image

## Final Checks:
- [ ] Delete branch

(optional)
- [ ] Write tests
- [ ] Write documentation

### DIFF
--- a/site/src/pages/RoadmapPage/Planner.tsx
+++ b/site/src/pages/RoadmapPage/Planner.tsx
@@ -6,7 +6,7 @@ import Header from "./Header";
 import AddYearPopup from "./AddYearPopup";
 import Year from "./Year";
 import { useAppSelector, useAppDispatch } from '../../store/hooks';
-import { selectYearPlans, setYearPlans, setInvalidCourses, setTransfers } from '../../store/slices/roadmapSlice';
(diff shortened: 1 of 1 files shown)


site/src/pages/RoadmapPage/Planner.tsx
```diff
@@ -235,25 +235,39 @@ const Planner: FC = () => {
     }
   }
 
+  const initializePlanner = () => {
+    if (data.length == 0) {
```
ethanwong16, line 239: Would be good to add a comment here that explains there is a future TODO once we have support for multiple planners: to ensure that a default new year is added only when a new planner is created, not when an existing planner is empty.

ethanwong16 (approved):
Great job @SenghoungLim ! LGTM 🥳 

I left a suggestion to add a comment explaining how a new year would always be created whenever the planner is empty and how we might want to refactor this later, but other than that feel free to merge and squash the branch when you're ready!

ethanwong16:
Closes #194

site/src/pages/RoadmapPage/Planner.tsx
SenghoungLim: Will do in the code, thank you!

Issue #243, opened 2022-12-01
Investigate login error


ethanwong16:
Login feature with Google Auth fails currently on both Safari and Chrome browsers.
Investigate issue and ensure usage of Google Auth API is up-to-date

PR #267, opened 2023-02-25
Investigate login error - Redirects to / by default

advayanand:
## Description

- Investigation:
  - Relevant file: api/src/controllers/users.ts 
  - Currently, redirection after logging in with Google is implemented by using the Referer header in the login request and redirecting to this URL. The Referer header stores the URL from which a login request was made: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer
  - The Referer header being passed is undefined, causing a redirect to /undefined, which renders a 404 page. 
  - Another problem is that the req.session.returnTo field does not persist after a call to req.login(). This means that the field is undefined by the time successLogin is called, which causes a redirect to /undefined.
- Temporary Fix:
  - This PR redirects the user to / by default if the returnTo address is undefined, which is the case as mentioned above. 

- Log in
- Verify that client is redirected back to / after login

## Final Checks:
- [ ] Delete branch

(optional)
- [ ] Write tests
- [ ] Write documentation


api/src/controllers/users.ts
```diff
@@ -44,6 +46,7 @@ router.get('/auth/google',
  * Callback for Google authentication
  */
 router.get('/auth/google/callback', function (req, res) {
+  console.log(req.session);
```
ethanwong16, line 49: If we're not using it could you comment this line out to minimize log sizes?

ethanwong16 (approved):
When you merge the PR, please keep the branch instead of deleting / squashing it so we can reference it in future advancements to the login.

Other than that LGTM!  Thanks @advayanand !

ethanwong16:
Closes #243

api/src/controllers/users.ts
advayanand: Done

ethanwong16:
5.5.2023 Discussion - @advayanand to merge hot fix since Ryan's Google auth changes are dependent on it + work with Jacob on improving login redirecting back to original page it was clicked from

PR #268, opened 2023-02-28
Fix wrong week calculations due to dayjs bug

js0mmer:
## Description
Replaces usage of dayjs objectSupport with ISO strings due to the dayjs timezone library not properly converting the timezones with objects.
See [this comment](https://github.com/icssc/peterportal-client/pull/254#issuecomment-1447324203) for more details.

Run locally. Switch system time zone to UTC. Switch system date & time to check edge cases (start & end of week, start & end of finals week, fall quarter).

## Final Checks:
- [ ] Delete branch

(optional)
- [ ] Write tests
- [ ] Write documentation

### DIFF
--- a/api/src/helpers/week.ts
+++ b/api/src/helpers/week.ts
@@ -103,7 +103,7 @@ function findWeek(date: dayjs.Dayjs, quarterMapping: QuarterMapping): WeekData {
             }
         }
         // check if date is after instruction end date and by no more than 1 week - finals week
-        else if (date >= end && date <= end.add(1, 'week')) {
+        else if (date >= end && date < end.add(1, 'week')) {
             let display = `Finals Week • ${quarter}. Good Luck!🤞`
             result = {
                 week: -1,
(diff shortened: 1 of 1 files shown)


ryanmohta (changes requested):
Looks mostly good except for one small thing, which I addressed in the comment I left. Once you fix that feel free to re-request a review!

Also out of curiosity, are there any specific reasons you’re using the day.js library and not the [Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) class in the JavaScript standard library? Usually it’s best to stick with what’s included with the language to avoid managing an extra dependency, unless there’s anything specific you need that the standard library doesn’t include.

js0mmer:
I used day.js rather than the JavaScript Date class since JavaScript's Date class lacks proper time zone support. There is no good way to construct Dates in a certain time zone, convert between time zones, or get/adjust the day of the week in another time zone. I discovered existing JavaScript classes/methods such as Intl but none of them seemed to do exactly what I needed. I thought it would be as easy to implement my own solution as adding/subtracting 8 hours to go between PST and UTC but then I discovered issues such as daylight savings time which starts/ends on a different date every year so it seemed like the better solution to use an existing library rather than implement my own solution or messy workaround.

ryanmohta (approved):

ethanwong16 (approved):
LGTM! Great work Jacob 🥳

PR #280, opened 2023-03-19
add year popup goes away when clicking outside

AyushSat:
This PR will allow the add year popup to be closed when anywhere else on the page is clicked.

Added event listeners to document to check where the clicks occur, if they are in the button toggle, else turn off the popup

Go to roadmap, press the add year to see popup, click outside the popup/button and see if it goes away.

### DIFF
--- a/site/src/pages/RoadmapPage/AddYearPopup.tsx
+++ b/site/src/pages/RoadmapPage/AddYearPopup.tsx
@@ -13,20 +13,37 @@ const AddYearPopup: FC<AddYearPopupProps> = ({ placeholderYear }) => {
   const dispatch = useAppDispatch();
   const [year, setYear] = useState(placeholderYear);
   const [show, setShow] = useState(false);
-  const target = useRef(null);
+  const target = useRef<HTMLButtonElement>(null);
+  const popupTarget = useRef<HTMLDivElement>(null);
 
   useEffect(() => { setYear(placeholderYear) }, [placeholderYear]);
 
-  const handleClick = (event: React.MouseEvent) => {
-    setShow(!show);
-  };
+  useEffect(() => {
+    const handleClick = (event: MouseEvent) => {
+      if (target.current && target.current.contains(event.target as Node)) {
+        setShow(!show);
+      } else {
+        setShow(false);
+      }
+    };
+    const rootElement = document.getElementById("root");
+    if(rootElement){
+      rootElement.addEventListener('click', handleClick);
+    }
+    return () => {
+      if(rootElement){
+        rootElement.removeEventListener('click', handleClick);
+      }
+    };
+  }, [popupTarget, target, show]);
 
   return (
     <div>
-      <Button variant="light" ref={target} className="add-year-btn" onClick={handleClick}>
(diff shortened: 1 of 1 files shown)


site/src/pages/RoadmapPage/AddYearPopup.tsx
```diff
@@ -13,20 +13,37 @@ const AddYearPopup: FC<AddYearPopupProps> = ({ placeholderYear }) => {
...
+    const rootElement = document.getElementById("root");
+    if(rootElement){
+      rootElement.addEventListener('click', handleClick);
+    }
+    return () => {
+      if(rootElement){
+        rootElement.removeEventListener('click', handleClick);
+      }
+    };
+  }, [popupTarget, target, show]);
```
js0mmer, line 38:
```suggestion
  useEffect(() => {
    const handleClick = (event: MouseEvent) => {
      if (target.current && target.current.contains(event.target as Node)) {
        setShow(show => !show);
      } else {
        setShow(false);
      }
    };
    const rootElement = document.getElementById("root");
    if(rootElement){
      rootElement.addEventListener('click', handleClick);
    }
    return () => {
      if(rootElement){
        rootElement.removeEventListener('click', handleClick);
      }
    };
  }, [popupTarget, target]);
```

js0mmer (changes requested):
Looks good. One minor gripe I have is that the effect seems to go off and add/remove the event listener every time the popup is toggled since show is in the dependency array. I know it will show a warning when it's not in the dependency but there's another workaround using a function as it mentions in the error message. This will fix the adding/removing the listener on every toggle.

image

I left a suggestion with this change.

Also what does the popupTarget ref do? I see that it references the div you added but I don't see it used anywhere except the dependency array. Is this intentional?

site/src/pages/RoadmapPage/AddYearPopup.tsx
```diff
@@ -13,20 +13,37 @@ const AddYearPopup: FC<AddYearPopupProps> = ({ placeholderYear }) => {
...
-  const target = useRef(null);
+  const target = useRef<HTMLButtonElement>(null);
+  const popupTarget = useRef<HTMLDivElement>(null);
 
   useEffect(() => { setYear(placeholderYear) }, [placeholderYear]);
 
-  const handleClick = (event: React.MouseEvent) => {
-    setShow(!show);
-  };
+  useEffect(() => {
```
ethanwong16, line 21: I'd suggest adding a quick comment here explaining the gist of the change and what it's purpose is, so it's easier to maintain or apply to other areas in the future

ethanwong16 (approved):
LGTM!  Take a quick look at my suggestion + Jacob's feedback 🔥

js0mmer:
Proper way of doing this would actually be to [use an OverlayTrigger](https://www.pluralsight.com/guides/how-to-dismiss-react-bootstrap-popover-on-click-outside)

js0mmer:
Will be redone using an [OverlayTrigger](https://react-bootstrap-v4.netlify.app/components/overlays/#overlay-trigger).

PR #284, opened 2023-04-22
Professor page/grade distribution crash

### LINKED ISSUE #276: Professor page crash
If you attempt to view information on [Professor Alberto Krone-Martins](https://peterportal.org/professor/algol), the page crashes and turns white.
professor page crash
This error is printed out to the console.
grade data error

Update:
After further investigation, the error is due to PeterPortal API returning a 400 error since it can not properly find grades for professors with hyphenated last names. Will open an issue with them. For now, we may want to add a check for this just so the entire page doesn't crash for users.

js0mmer:
## Description
Fixes #276. Changed the express route for `/api/grades/:name` to return the same status that PPAPI responds with instead of a status 200 every time (i.e. status 400 in this case). Caught the error in the request the GradeDist component makes to the backend and changed the code to continue to display nothing while data is being fetched (null) but now display an error message when no grade data is retrieved from PPAPI (empty array []). Logged response to console in the error catch. Also removed an unnecessary useEffect that resulted in grade data being requested twice every time the component rendered/updated.

## Screenshots
Before:
Page crashes
professor page crash
After:
Page loads with error message
professor page loads with error message under grade distribution
Response logged to console
response logged in console

## Final Checks:
- [ ] Delete branch

(optional)
- [ ] Write tests
- [x] Write documentation

### DIFF
--- a/api/src/controllers/professors.ts
+++ b/api/src/controllers/professors.ts
@@ -53,7 +53,6 @@ router.get('/api/grades/:name', function (req, res, next) {
 
   r.then((response) => {
     status = response.status;
-    console.log(status);
     return response.json();
   }).then((data) => res.status(status).send(data))
 });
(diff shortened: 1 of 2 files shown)


ethanwong16 (approved):
LGTM! Great job Jacob 🔥 🥳

nathantoannguyen (approved):
lgtm!

PR #286, opened 2023-04-26
Add `/api` to Google auth URL to reflect Node 18 upgrade

ryanmohta:
Fixes issue where the sidebar links to /users/auth/google instead of /**api**/users/auth/google. The `/api` prefix is used to mark which URLs should be proxied to `localhost:5000`, which is required to be compatible with Node 18 (see #247). Thanks to @Voark for pointing this out and @js0mmer for explaining the issue!


ethanwong16 (approved):
LGTM!

js0mmer (approved):

PR #287, opened 2023-04-27
Multiple plans

MFYLM:
## Description

- added multiplanner component 
- updated roadMapSlice to support multiple plans

## Screenshots
image

## Final Checks:
- [ ] Delete branch

(optional)
- [ ] Write tests
- [ ] Write documentation


ethanwong16:
Blocker for issue #194 , @SenghoungLim to update once closed

SenghoungLim (changes requested):
Great job on this, it looks good. I have a few suggestions. 
1. I think once a user clicks on "Delete Plan", there should not be a scroll bar in the modal window. It might be confusing since we do not have more info to scroll down besides the cancel and delete buttons. 
Delete Plan

2. Ties from the first suggestion, once a user adds plan name, there should not be a scroll, I think it's inconvenient to scroll down to submit the plan. 
image
3. It might be a nitpicking for me, but I think the sentence "Are you sure about deleting the current plan?" should change to "Are you sure you want to delete _"plan name"_? (You cannot undo this action)". I think this will allow users to know that it is permanent and ensure that they don't delete the wrong plan. It also shows consistency with other projects like AntAlmanac.

ethanwong16 (changes requested):
LGTM!  

I'm just requesting more changes so if you could remove the `package-lock.json` and `package.json` issues and test it out on the staging instance before merging that would be great!

js0mmer:
Updated from main and handled merge conflicts. Should now have staging instance, be node 18 compatible, etc.

site/src/store/slices/roadmapSlice.ts
```diff
@@ -167,61 +205,88 @@ export const roadmapSlice = createSlice({
...
 
+
 export const { moveCourse, deleteCourse, addQuarter, deleteQuarter, clearQuarter, clearYear, addYear, editYear, deleteYear, clearPlanner,
     setActiveCourse, setYearPlans, setInvalidCourses, setShowTransfer, addTransfer, setTransfer,
-    setTransfers, deleteTransfer, setShowSearch, setShowAddCourse } = roadmapSlice.actions
+    setTransfers, deleteTransfer, setShowSearch, setShowAddCourse,
+    setRoadmapPlan, addRoadmapPlan, deleteRoadmapPlan, setPlanIndex, setPlanName } = roadmapSlice.actions
+
+
+// export const { setRoadmapPlan, addRoadmapPlan, deleteRoadmapPlan } = roadmapMultiplePlanSlice.actions;
```
js0mmer, line 286:
```suggestion
```

Remove if not needed.

site/src/pages/RoadmapPage/RoadmapMultiplan.tsx
```diff
@@ -0,0 +1,163 @@
...
+                    <InputLabel style={{ width: "min-content" }} id="demo-simple-select-standard-label">{name}</InputLabel>
+                    <Select
+                        labelId="demo-simple-select-standard-label"
+                        id="demo-simple-select-standard"
+                        sx={{ width: "auto" }}
+                        value={""}
+                        label={name}
+                        onChange={(e) => {
+                            let newIndex = +(e.target.value);
+                            // console.log("selected index", newIndex);
```
js0mmer, line 80:
```suggestion
```

js0mmer (changes requested):
Great start. A few things to note:
* Additional schedules are not saved locally or to your account if logged in when clicking save
  * upon reloading the page, additional schedules are lost
  * only the first schedule is saved, also the name of it is reset back to "Schedule 1"
* I would prefer that you used Bootstrap for the schedule selector, delete button, and modals
  * I think it makes more sense to stay consistent with the rest of the site and use Bootstrap rather than add MUI and have extra dependencies/different styling just for this feature
  * Also, I'm not too sure what the emotion package is but I don't think we need to add that either
* I would look into changing the UI for managing different schedules to be more like AntAlmanac's new schedule selector, with the edit button, delete, and add all in the same place
  * I think their new UI is more intuitive to use and we should copy that
image
* Lastly, consider adding a "Cancel" button to the edit name/add schedule modals

Awesome-E:
> I would look into changing the UI for managing different schedules to be more like AntAlmanac's new schedule selector, with the edit button, delete, and add all in the same place
I think their new UI is more intuitive to use and we should copy that

I agree with this; maybe we could put it where it currently says "Peter's Roadmap"?
I'm thinking something like this:
image

Awesome-E:
Progressed picked up in #472

PR #289, opened 2023-04-29
Rework infrastructure

Voark:
# Warning
Do not merge to main, because the stacks are completely different from main. The old stacks will have to be deleted and then this PR should be merged. Also after this PR is merged all the old staging stacks will have to be deleted manually and redeployed. When this PR is approved I'll merge it and monitor it just in case I break prod (probably gonna happen).

## Description
The main goal of this rework was to move the frontend files from a Lambda to an S3 bucket/Cloudfront distribution. However it became too hard to do this with Serverless so I also switched the infra framework to SST (framework built on top of AWS CDK). 

In summary:
- Infra changes
  - Frontend files are now stored on an S3 bucket and served from a Cloudfront distribution
  - An API Gateway routes all requests with `/api/*` to the backend lambda while everything else is routed to the Cloudfront distribution
  - Backend is now built and minified before being uploaded to the Lambda
- Technology changes
  - [SST](https://sst.dev/)
    - `sst deploy --stage {stage}` automatically builds both the front and backend then deploys them. `--sage staging-{pr_num}` deploys to `staging-{pr_num}.peterportal.org` and `--stage prod` deploys to `peterportal.org`.
- Some misc changes and explanation
  - Removed the `.DS_Store` files. It was bothering me too much
  - Updated README. Removed references that seem to be related to PeterPortal API.
  - Commit `package-lock.json`. I'm pretty sur this is standard practice, and all except one of them was committed anyways...
  - Moved AWS region to `us-west-1`

## Potential future improvements
- Move to [vite](https://vitejs.dev/) to build the frontend
- Add [tRPC](https://trpc.io/) to the backend

- [x] Verify changes work as expected on staging instance (it deploys)

## Final Checks:
- [ ] Delete branch


ryanmohta:
These changes look great, I’m excited for when we merge them in! The build and deploy action is so much smaller now LOL.

I noticed you didn’t modify the PR cleanup action at all, was that intentional? How do these frontend and backend stacks get removed once a PR is merged?

Voark:
> These changes look great, I’m excited for when we merge them in! The build and deploy action is so much smaller and cleaner now LOL.
> 
> I noticed you didn’t modify the PR cleanup action at all, was that intentional? How do these frontend and backend stacks get removed once a PR is merged?

I just straight up forgot, will work on it! Thanks for reminding me.

.github/workflows/clean-up-pr.yml
```diff
@@ -20,20 +20,27 @@ jobs:
...
-            serverless plugin install --name serverless-domain-manager && 
-            serverless remove"
-          entrypoint: /bin/sh
+          node-version: '18'
+          cache: 'npm'
+
+      - name: Install Dependencies
+        run: npm install
+
+      - name: Build and deploy
```
ryanmohta, line 32: Might want to change the name of this step to indicate it’s for destroying the stack

ryanmohta (commented):
I left one suggestion for something you might want to change but other then that, looks good! Do you think we need to make any changes to concurrency to accommodate this new system, or is it fine as is?

.github/workflows/clean-up-pr.yml
Voark: Oops I copy and pasted too hard

ethanwong16 (approved):
thank you so much @Voark ! and for the extra cleanup too 😎 

I think we already had .DS_Store ignored in the .gitignore but looks like this was only added after the file was already there and it never was deleted haha

PR #290, opened 2023-05-04
Chart tooltip transparency fix

### LINKED ISSUE #285: Background colors of grade bar graph popup text don't match
The highlight/background color of the text in the hover popup for the grade bar graph does not match the color of the padding/is missing transparency.

bghighlight

js0mmer:
## Description
Resolved an issue with styling being applied twice to both an outer div and an inner div in the tooltip, resulting in background colors stacking and becoming more opaque. Added transparency to the pie chart tooltip for consistency. Closes #285. Also I noticed some redundancy in the code for Chart and Pie. Both of them use the same colors for the grades but they are defined separately for each component. A future improvement to the codebase would be to have a single file defining/exporting the color constants to be used for each component. Also, I'm not sure why the tooltips are different dimensions/sizes for the Pie and the Chart but I'll just leave it.

## Screenshots
Before:

before_l
before_r

After:

after_l
after_r

## Final Checks:
- [ ] Delete branch

(optional)
- [ ] Write tests
- [ ] Write documentation

### DIFF
--- a/site/src/component/GradeDist/Chart.tsx
+++ b/site/src/component/GradeDist/Chart.tsx
@@ -118,7 +118,7 @@ export default class Chart extends React.Component<ChartProps> {
    */
   styleTooltip = (event: BarTooltipDatum) => {
     return (
-      <div style={this.theme.tooltip.container}>
+      <div>
         <strong>
           {event.data.label}: {eval('event.data.' + event.data.label)}
         </strong>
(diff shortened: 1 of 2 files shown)


SenghoungLim:
Nice catch on this. I think for the different size between the pie and the chart is visually fine since the text in there one is longer. Good job on this.

ethanwong16 (approved):
LGTM! great job 🥳

PR #291, opened 2023-05-05
Staging login callback redirect

js0mmer:
## Description
This PR allows logging in on staging instances. Since for each redirect_uri/callback url must be registered with Google Cloud for security reasons, the callbacks for staging instances do not work. However, by default, our code sends the callback back to prod instead of the staging instance. This PR makes use of this by passing the correct host to redirect the callback to in the state parameter in the Google authentication URL and on callback to prod, prod will redirect callback requests that are meant for staging instances back to the staging instance.

The redirection couldn't be tested as is with prod without actually merging to master and testing on prod... so I tested it by changing the callback URL from the prod domain environment variable to http://localhost:5000/api/users/callback. Then I ran PeterPortal locally and tried logging in on the staging instance and it correctly sent the callback to localhost and my local server redirected it back to the staging instance, successfully signing me in on the staging instance. I think I'll leave that right now for the sake of testing during review and will change it back before I merge.

- [x] Checkout this branch and run the API locally

## Final Checks:
- [ ] Delete branch

(optional)
- [ ] Write tests
- [x] Write documentation


api/src/config/passport.ts
```diff
@@ -23,7 +23,7 @@ passport.use(
         {
             clientID: process.env.GOOGLE_CLIENT,
             clientSecret: process.env.GOOGLE_SECRET,
-            callbackURL: process.env.PRODUCTION_DOMAIN + '/api/users/auth/google/callback',
+            callbackURL: 'http://localhost:5000/api/users/auth/google/callback',
```
js0mmer, line 26: TODO: restore prod env variable url before merging

ethanwong16 (approved):
LGTM! Great work @js0mmer :))
Feel free to change back the `callbackURL` prod env variable and merge whenever you'd like

PR #298, opened 2023-05-12
Updated Build & Deploy Message

nicoleiftekhar:
Updated the build and deploy message to display that the request may take a few minutes to process. The message added includes... 

"Please note that it usually takes a couple minutes for the instance to be deployed and the link above may temporarily display an error message."

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -109,4 +109,6 @@ jobs:
         with:
           header: staging url
           recreate: true
-          message: Deployed staging instance to https://staging-${{ github.event.pull_request.number }}.peterportal.org
+          message: |+ 
+          Deployed staging instance to https://staging-${{ github.event.pull_request.number }}.peterportal.org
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-deploy.yml
```diff
@@ -109,4 +109,4 @@ jobs:
         with:
           header: staging url
           recreate: true
-          message: Deployed staging instance to https://staging-${{ github.event.pull_request.number }}.peterportal.org
+          message: Deployed staging instance to https://staging-${{ github.event.pull_request.number }}.peterportal.org Please note that it usually takes a couple minutes for the instance to be deployed and the link above may temporarily display an error message.
```
ethanwong16, line 112: Great work! A minor suggestion, were you able to look into potentially adding a newline between the previous message and the new one you added?

ethanwong16 (approved):
LGTM! Left a minor suggestion

js0mmer:
Closing for now since it no longer seems to be necessary with our current infrastructure (deploying with SST). Instance seems to be up as soon as the link is commented now.

PR #316, opened 2023-05-27
Add review sorting

### LINKED ISSUE #275: Order reviews by most recent
Currently reviews are listed out of order. I think it would make more sense to show the most recent reviews first. Another idea that could be considered though is to have a dropdown of filters like on Amazon with, most recent, highest rated, etc. and an option to look at  reviews by number, 1/5 only, 2/5 only, etc.
out of order reviews

js0mmer:
## Description
Added sorting by most recent, top reviews (highest score), and controversial (lowest score). Also added a checkbox to filter to verified reviews only. Also fixed some react key issues after running into one with the array of subreviews. The key being the index caused data to be mismatched among the reviews when sorting them and changing their order. The new key for each  subreview in the array is now the review id. Closes #275.

## Screenshots
Recording 2023-05-26 at 22 45 18

## Final Checks:
- [ ] Delete branch

(optional)
- [ ] Write tests
- [ ] Write documentation


SenghoungLim (changes requested):
Everything seems to work well, good job! Just a nitpicking, on a smaller screen like iPhone X, the verification checkbox is a bit clustered from the sorting menu. I suggest we can have a newline separation and align the check box to the sorting box above. Otherwise, everything looks good. 
image

js0mmer:
> Everything seems to work well, good job! Just a nitpicking, on a smaller screen like iPhone X, the verification checkbox is a bit clustered from the sorting menu. I suggest we can have a newline separation and align the check box to the sorting box above. Otherwise, everything looks good. image

Good catch. Let me know what you think of it now.

SenghoungLim (approved):
This looks perfect, well done!

PR #323, opened 2023-06-02
Add dark mode & refactor styling

### LINKED ISSUE #9: Dark mode

js0mmer:
## Description

* Created darkMode state in App.tsx
* Created ThemeContext in style/theme-context.ts
* Implemented ThemeContext in App.tsx to provide theme and toggleTheme functions to child components
* Change data-theme attribute on body tag from 'dark' to 'light' on theme toggle
  * Used for styling with `[data-theme='dark']` css selector
* Refactored css to use variables for most colors
* Created style/theme.scss with variables defining colors for light theme and dark theme
* Added additional overrides for bootstrap and semantic-ui styles when using dark theme
* Added a theme toggle to the expanded user sidebar
* Save and load theme preferences from local storage when not signed in
* Added preferences collection to MongoDB
* Added user preferences endpoint to backend (`/api/users/preferences`)
* Save and load theme preferences from MongoDB when user is signed in and toggles their theme
  * Fall back to local preferences if user has no theme preference saved to their account
* Use user's system theme preferences if the user has no locally saved preferences

### Other changes
* Also cleaned up some styling in AppHeader.scss. There was some unused stuff and there and also some stuff that shouldn't be in there like styles that unintentionally selected every icon and shifted them 2px right. If you compare the arrows for the gradedist dropdown on here to prod, you'll noticed those were one of the affected icons by this.
* Made a few other miscellaneous styling fixes. e.g. The blue box shadow when the search box is focused was cut off on the home page. Fixed that by unhiding overflow for the container (compare with prod to see difference).
* Attempted to remove the white background from searching.png. It has a jagged look to it though. Also moved it to the asset folder with the rest of the images.

### Future changes to implement
* ~~Make searching.png (search image on home page) look less jagged~~ (#376)
* Make dark theme compatible gif for 404 page
* Also maybe make a dark guest icon

Additional information on the implementation can be found on the [styling wiki page](https://github.com/icssc/peterportal-client/wiki/Styling).

Closes #9.

## Screenshots
image
image
image
image
image
image
image
image
image
image
image

- [ ] Make sure styling looks right for both dark mode and light mode. I think I got all the modals, dropdowns, etc. without breaking anything for light mode but please make sure I'm not missing anything

## Final Checks:
- [ ] Delete branch

(optional)
- [ ] Write tests
- [ ] Write documentation


Awesome-E:
Fun little thing that you don't have to do – if you want to make the captcha respect dark mode, you can apply the following style to the Recaptcha's iframe:
```css
#some-identifier {
  filter: invert(1) hue-rotate(180deg);
}
```
This only changes the checkbox and won't affect the actual image selector

site/src/style/globalStyle.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  .prereq .ui.button {
+    color: ${({ theme }) => theme.prereqNodeTextColor};
+    background-color: ${({ theme }) => theme.prereqNodeBackgroundColor};
+  }
+
+  `;
+
+  export default GlobalStyle;
```
Awesome-E: new line at end of file?

site/src/style/theme.ts
```diff
@@ -0,0 +1,43 @@
...
+    backgroundColor: string,
+    overlay1Color: string,
+    overlay2Color: string,
+    quarterTitleColor: string,
+    navbarBoxShadow: string,
+    prereqNodeTextColor: string,
+    prereqNodeBackgroundColor: string
+}
+
+export default Theme;
```
Awesome-E: new line at end of file?

site/src/component/AppHeader/AppHeader.scss
```diff
@@ -325,3 +187,11 @@ i.icon {
...
   }
 }
+
+[data-theme='dark'] .beta-popup .button {
+  background-color: var(--overlay3);
+}
+
+[data-theme='dark'] .beta-popup .button:hover, [data-theme='dark'] .beta-popup .button:focus {
+  background-color: #CACBCD;
+}
```
Awesome-E: newline

site/src/component/PrereqTree/PrereqTree.scss
```diff
@@ -187,3 +187,12 @@
...
 }
+
+.prereq-text-box {
+    background-color: var(--overlay2);
+}
+
+.prereq ul {
+    margin: 0;
+    padding: 0;
+}
```
Awesome-E: newline

site/src/component/SideBar/SideBar.tsx
```diff
@@ -74,14 +70,19 @@ const SideBar = ({ darkMode, toggleTheme }: SideBarProps) => {
...
-            <Icon name={darkMode ? 'sun outline' : 'moon outline'} size='large' />
-          </div>
-          <span>
-            {darkMode ? 'Toggle Light Theme' : 'Toggle Dark Theme'}
-          </span>
-        </a>
+        <ThemeContext.Consumer>
+          {({ darkMode, toggleTheme }) => 
+            // eslint-disable-next-line jsx-a11y/anchor-is-valid
+            <a className='theme-toggle' onClick={toggleTheme}>
```
Awesome-E: why are we using an `a` and onclick as opposed to a different element? If it's just for styling we should probably use a different element

site/src/component/SideBar/Sidebar.scss
```diff
@@ -142,3 +148,8 @@
     }
   }
 }
+
+.theme-toggle:hover {
+  cursor: pointer;
+  text-decoration: underline;
+}
```
Awesome-E: new line

api/src/controllers/users.ts
```diff
@@ -15,6 +16,54 @@ router.get('/', function (req, res, next) {
...
+  const userID = req.session.passport.user.id;
+
+  getDocuments(COLLECTION_NAMES.PREFERENCES, { _id: userID }).then((preferences) => {
+    if (preferences.length > 0) {
+      res.json(preferences[0]);
+    } else {
+      res.json({ error: 'No preferences found' });
+    }
+  });
+});
```
Awesome-E, line 34: Why not use an async function for the request handler?

```suggestion
router.get('/preferences', async (req, res) => {
  if (!req.session.passport) {
    res.json({ error: 'Must be logged in to get preferences.' });
    return;
  }

  const userID = req.session.passport.user.id;

  const preferences = getDocuments(COLLECTION_NAMES.PREFERENCES, { _id: userID });
  if (preferences.length > 0) return res.json(preferences[0]);
  res.json({ error: 'No preferences found' });
});
```

api/src/controllers/users.ts
```diff
@@ -15,6 +16,54 @@ router.get('/', function (req, res, next) {
...
+  if (!req.session.passport) {
+    res.json({ error: 'Must be logged in to get preferences.' });
+    return;
+  }
+
+  const userID = req.session.passport.user.id;
+
+  // make user's preference doc if it doesn't exist
+  const preferencesCollection = await getCollection(COLLECTION_NAMES.PREFERENCES);
+  if ((await preferencesCollection.find({ _id: userID }).count()) == 0) {
```
Awesome-E, line 50: Any reason we can't do the same logic as above?
```ts
const preferences = getDocuments(COLLECTION_NAMES.PREFERENCES, { _id: userID });
if (preferences.length === 0) // add document
```

Otherwise let's still use `===` for good practice

site/src/App.tsx
```diff
@@ -15,28 +17,81 @@ import AdminPage from './pages/AdminPage';
...
+import { useCookies } from 'react-cookie';
+
 export default function App() {
+  // default darkMode to local or system preferences
+  const [darkMode, setDarkMode] = useState(
+    localStorage.getItem('theme') === 'dark'
+      ? true
+      : localStorage.getItem('theme') === 'light'
+        ? false
+        : window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches,
```
Awesome-E, line 31: not sure what I think of having nested ternaries. I think we could do something like this, but not sure how readable this is either 😅 

```ts
const localDark = ({ dark: true, light: false})[localStorage.theme];
const mediaDark = window.matchMedia?.('(prefers-color-scheme: dark)')?.matches
// ...
useState(localDark ?? mediaDark)
```

site/src/App.tsx
```diff
@@ -15,28 +17,81 @@ import AdminPage from './pages/AdminPage';
...
+        } else if (theme === 'light') {
+          setDarkMode(false);
+        } else {
+          // not defined or set to use local/system prefs
+          setDarkMode(
+            localStorage.getItem('theme') === 'dark'
+              ? true
+              : localStorage.getItem('theme') === 'light'
+                ? false
+                : window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches,
```
Awesome-E, line 51: duplicate code; consider refactoring this into a variable or function

Awesome-E (changes requested):
This text/button looks to be a little bit low contrast. I know not all of the buttons follow the "proper contrast ratios" (I don't really care too much for that tbh 🙃), but since this one uses that color for just text it makes it a little bit harder to see than I'd like.
image

The borders here should have a dark mode equivalent; having light gray instead of a darker variant makes it a little jarring

Current:
image image
Using Dark reader chrome extension:
image

Also noticed the extra whitespace with Google Recaptcha not following theme, which is annoying. Not sure if you'd want to implement [a monkeypatch for that](https://stackoverflow.com/questions/32547219/recaptcha-dark-theme-displays-with-extra-white-space-on-edges) though

---

Course planner looks fine except for the same thing with borders as mentioned above.

---
Reviews and verify reviews look good

site/src/component/SideBar/SideBar.tsx
js0mmer: I think I was just too lazy too change the styling for a different element at the time. Should probably change it to a button but might just leave it for now. This is a temporarily a place for the toggle. In the future, I would like the redesign the header and nav menu and have a profile menu in the top right.

image

Idea is that the theme button would expand to have system, light, and dark theme options, kind of like YouTube.

image
image

And yes, jsx-a11y was removed so the eslint-disable was not needed. We might consider adding it back in the future, but it was too much trouble at the time to incorporate in the vite PR due to the abundance of accessibility errors we have.

Awesome-E: I see. I think if there's not any warnings then we can leave it for now until we make changes to the sidebar and/or user preferences

api/src/controllers/users.ts
js0mmer: Good question

api/src/controllers/users.ts
js0mmer: We actually have a `containsID` helper that seems to do this so I'll just use that instead here.

site/src/App.tsx
js0mmer: I'll refactor to use a function for both occurrences of this ternary.

api/src/controllers/users.ts
Awesome-E: Ok nice I looked for that cause I thought it'd be a good thing to have but didn't notice. That's definitely better!

Awesome-E (approved):

PR #329, opened 2023-07-19
Courses won't be deleted if they have no destination

### LINKED ISSUE #279: Investigate : Peter's Roadmap drag to remove classes
Feedback from Google Forms

> It would be nice if removing a class via dragging prompted an are you sure message. I keep on accidentally removing my classes when I miss dragging it to another quarter and then I forget which class I just removed.

KevinWu098:
## Description

- Separated results for No Destination and for Search
- If `!result.destination` the function simply returns and no action is taken.

## Screenshots

Before:
Courses get Thanos-ed

After:
Courses don't get Thanos-ed

## Final Checks:
- [ ] Delete branch

Closes #279 

<sub>_First PR over here in PP-C land, so LMK if there's some nonsense that I've written or some convention that I've missed when contributing!_</sub>

### DIFF
--- a/site/src/pages/RoadmapPage/index.tsx
+++ b/site/src/pages/RoadmapPage/index.tsx
@@ -14,8 +14,11 @@ const RoadmapPage: FC = () => {
 
   const onDragEnd = useCallback((result: DropResult) => {
     if (result.reason === 'DROP') {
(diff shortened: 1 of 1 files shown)


js0mmer (approved):
LGTM!

PR #331, opened 2023-07-20
Bar Charts with larger scales have wider margins

### LINKED ISSUE #296: Scale on grade distribution chart cut off for 4 digits
Numbers with more than 3 digits are cut off on the y-axis scale of the bar chart.
gradedistscalecutoff
([ART 8](https://peterportal.org/course/ART8) for reference)

KevinWu098:
## Description

- `this.classData()` is set to a new const `data` so that `largeGraphScale` can access the array without recalling `this.classData()` 
- Added new const `greatestCount` to find the largest grade value within data
- Added new const `marginX` which calculates margins depending on the magnitude of greatestCount, with every magnitude above 100 increasing the base margin (30) by 5. For example, 5173 would set marginX to 35, while 123 would set it to 30. 

More detailed documentation is in code comments!

## Screenshots

Before:
Before Chart

After: 
chrome-capture-2023-6-28

## Final Checks:
- [ ] Delete branch

Closes #296

### DIFF
--- a/site/src/component/GradeDist/Chart.tsx
+++ b/site/src/component/GradeDist/Chart.tsx
@@ -46,7 +46,7 @@ export default class Chart extends React.Component<ChartProps> {
    * Create an array of objects to feed into the chart.
    * @return an array of JSON objects detailing the grades for each class
    */
-  getClassData = () => {
+  getClassData = (): Bar[] => {
     let gradeACount = 0, gradeBCount = 0, gradeCCount = 0, gradeDCount = 0,
       gradeFCount = 0, gradePCount = 0, gradeNPCount = 0;
 
@@ -132,7 +132,7 @@ export default class Chart extends React.Component<ChartProps> {
    */
   render() {
     const data = this.getClassData()
-    const largeGraphScale = data.some((grade) => grade[grade.id] > 999));
+    const largeGraphScale = data.some((grade) => grade[grade.id] as number > 999);
 
     return <>
       <ResponsiveBar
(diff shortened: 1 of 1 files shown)


site/src/component/GradeDist/Chart.tsx
```diff
@@ -131,16 +131,19 @@ export default class Chart extends React.Component<ChartProps> {
    * @return a JSX block rendering the chart
    */
   render() {
+    const data = this.getClassData()
+    const largeGraphScale = data.some((grade) => Object.values(grade).some((value) => value > 999));
```
js0mmer, line 135: I assume that it's always the case that 1000 will never appear on the scale unless one of the bars has a height of 1000 or more, is that correct?

Also small suggestion but this condition can be rewritten:

```suggestion
    const largeGraphScale = data.some((grade) => grade[grade.id] > 999));
```

KevinWu098: I'm currently stuck on mobile, but I'd be happy to try and hunt down a class with ~1000 later to check and make sure this is absolutely true.

From my understanding, unless the number of students is greater than the next tick, whatever increment it may be, the label wouldn't display. 

TLDR: Yes, I think you're right about 1,000 not being displayed unless we've got a value >= 1000, but I'm happy to absolutely confirm this later.

KevinWu098: Screenshot 2023-07-27 at 3 53 14 PM

Screenshot 2023-07-27 at 3 52 56 PM

It's a bit scuffed, but just looking at these two examples and just how other graphs handle new ticks, I'm pretty confident that the current implementation with `> 999` should work! @js0mmer

KevinWu098:
After reviewing the code a bit, I realized that only checking for four digit number would kick the can down the road for five digits etc.

TLDR Wrote some new code to find the largest grade count, then increase the margin size based on the magnitude of that largest count.

js0mmer (approved):
LGTM

PR #332, opened 2023-07-26
Removing &nbsp; prevents title overflow

### LINKED ISSUE #321: Overflow for courses with long titles on mobile
Courses with long titles overflow on mobile and cause extra space to the right and a horizontal scrollbar to be added.

image

KevinWu098:
## Description

1. Removed `&nsbp;` between props.number and props.title which was preventing long titles from wrapping appropriately and added a comment documenting why `props.number` and `props.title` now share a span.

## Screenshots

Before:

After (Mobile S):
Screenshot 2023-07-28 at 5 13 40 PM

- [ ] Make sure text-wrapping works across a range of different devices for different course title lengths.

## Final Checks:
- [ ] Delete branch

(optional)
- [ ] Write tests
- [ ] Write documentation

Closes #321

### DIFF
--- a/site/src/pages/SearchPage/CourseHitItem.tsx
+++ b/site/src/pages/SearchPage/CourseHitItem.tsx
@@ -39,9 +39,12 @@ const CourseHitItem: FC<CourseHitItemProps> = (props) => {
           <h3>
             <span>{props.department}</span>
             &nbsp;
-            <span>{props.number}</span>
-            {isMobile && !isTablet ? <br /> : <>&nbsp;</>}
-            <span>{props.title}</span>
+            <span>{props.number} {props.title}</span>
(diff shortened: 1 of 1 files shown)


KevinWu098:
Now that I think about it, there is still an extreme edge case such that `props.number` could overflow if the screen size was somehow that small in width.

Not very sure what the remaining `&nbsp;` is intended to do other than just being a space, so I'm happy to also remove it if that makes sense to whoever's reviewing!

js0mmer (approved):
We don't need to worry about that edge case. Realistically, any site will break with a small enough screen size. Feel free to remove the other `&nbsp;` and span. They don't seem to serve any purpose.

js0mmer (approved):

PR #366, opened 2023-10-17
- force render when search index changes

### LINKED ISSUE #305: Not fetching professor data when user clicks professor tab
If you go to https://peterportal.org/search/courses and click professor tab on the top left, you will go to https://peterportal.org/search/professors and see an empty page about professors like this, which is not consistent with course page.
image
But if you refresh the page, it will retrieve necessary data from backend, so it may have something to do with changing state that does not trigger component to be re-rendered.

MFYLM:
## Description
Before when users click between professors and courses, it won't trigger necessary API call to retrieve data from backend and there will be a blank page with no information. Now I force the page to render and users could see both pages.

## Final Checks:
- [ ] Delete branch

(optional)
- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #305

### DIFF
--- a/site/src/component/SearchModule/SearchModule.tsx
+++ b/site/src/component/SearchModule/SearchModule.tsx
@@ -1,16 +1,15 @@
-import React, { useState, useEffect, Component, FC } from 'react';
-import './SearchModule.scss';
-import wfs from 'websoc-fuzzy-search';
-import axios from 'axios';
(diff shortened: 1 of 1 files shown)


js0mmer (approved):
LGTM!

aahei (approved):
LGTM

PR #368, opened 2023-10-19
Added renaming year on roadmap

### LINKED ISSUE #300: Allow Naming Years in Peter's Roadmap
currently years are just titled Year 1, Year 2, etc. and are ordered chronologically.

Because some students may have unconventional timelines (e.g. transfer students technically start at "Year 3"), perhaps it would be nice to allow for renaming of years.

kylerpan:
Issue #300
## Description

* Created an extra form group on top of 'Start Year' to change the name (within 'Edit Year')
  * Both for edit year and add year
* Added 'name' attribute to PlannerData to refer to the name of the year
* Added new reducer to edit PlannerData 'name' attribute
* Trimmed the name users enter for duplicate checking purposes

### Potential concerns
* No check if the name is empty
* The 'Clear' button doesn't reset the name 

## Screenshots

https://github.com/icssc/peterportal-client/assets/66697740/e77726d9-2294-4949-8c82-731569d2dec1

## Final Checks:
- [ ] Delete branch

(optional)
- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #300


site/src/pages/RoadmapPage/Year.tsx
```diff
@@ -90,7 +92,7 @@ const Year: FC<YearProps> = ({ yearIndex, data }) => {
               ) : (
                 <CaretRightFill className="caret-icon" />
               )}
-              <span id="year-number">Year {yearIndex + 1} </span>
+              <span id="year-number">{data.name} </span>
```
js0mmer, line 95: The name attribute will not exist for user's with existing roadmaps. Given that, I think we should fall back to "Year {yearIndex + 1} in the event that the year does not have a name set. It displays nothing, otherwise.

js0mmer (changes requested):
Nice work. I do agree that being able to set no name is a concern. I think there should be some input constraints: non-empty and also a character limit.
I'm not sure what you mean about the clear button, it seems work fine for me. Also requesting an additional review from @MFYLM since he's worked with the planner page.

site/src/pages/RoadmapPage/Year.tsx
kylerpan: Changed to a conditional operator for data.name, that way if the user has a name for the year it wouldn't be "Year 1 Name".

PR #372, opened 2023-10-21
Set up different development/prod mongo databases

### LINKED ISSUE #361: Separate mongo db for dev and prod
So we aren't testing on prod db.

Database with name peterPortalDevDB is already set up in mongo, just need a condition in the backend to use that when running the server in a dev envrionment.

js0mmer:
## Description
Uses a different Mongo database for local development/staging instances and production.

- [x] Verify changes work as expected on staging instance (make sure it's using dev db)

## Final Checks:
- [ ] Verify successful deployment (make sure prod is actually using prod db)
- [ ] Delete branch

(optional)
- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #361


advayanand:
LGTM. Testing in staging uses the development Mongo database as expected.

advayanand (approved):

PR #379, opened 2023-11-04
Setup different dev/prod mongo databases attempt #2

### LINKED ISSUE #361: Separate mongo db for dev and prod

js0mmer:
## Description
Redo of #372 
Forgot to add the env variable to the SST backend file.

## Issues

Closes #361


SenghoungLim (approved):
There is only one way to find out buddy :)

PR #389, opened 2023-12-11
Add prettier

### LINKED ISSUE #370: Add prettier
Add prettier and set up a config for consistent code formatting

js0mmer:
## Description
* Adds prettier and a husky precommit hook that calls lint-staged
* Next step would be to run formatting on the repo, otherwise there will be large diffs when reviewing PRs.
  * I think this would be better to implement in a separate PR, that way changes from this PR can be added to existing branches first which can help mitigate merge conflicts

## Screenshots
Example commit
Screenshot 2023-12-07 203350

- [x] Check out locally, `npm install` in root, then make a change and a test commit

## Final Checks:
- [ ] Delete branch

(optional)
- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #370


lint-staged.config.cjs
```diff
@@ -0,0 +1,6 @@
+/** @type {import('lint-staged').Config} */
+const config = {
+  '*.{scss,js,ts,tsx}': ['prettier --write'],
+};
+
+module.exports = config;
```
Awesome-E, line 6: new line at end of file

prettier.config.cjs
```diff
@@ -0,0 +1,8 @@
+/** @type {import('prettier').Config} */
+const config = {
+  printWidth: 120,
+  singleQuote: true,
+  trailingComma: 'none',
+};
+
+module.exports = config;
```
Awesome-E, line 7: new line at end of file

Awesome-E (changes requested):
image
At least on MacOS, the hook doesn't run because it's not given executable permissions

Awesome-E (approved):
Formatting looks like it's working now!

I think one more thing just to be cautious of – since all commits will autoformat the entire file now, we should really make an effort to resolve the other PRs because any change to the same files will lead to lots of merge conflicts

PR #391, opened 2023-12-15
Checking for already existing review from same professor,courseID, and userID

### LINKED ISSUE #378: Multiple reviews can be added to same course/professor
I don't think we should allow this since there's no reason for the user to need to review the same course taught by the same professor a second time.

In the backend, for the endpoint to add a review, query for any review with the user id, course id, and professor id. If one exists, do not allow a second review to be created. Send back an appropriate error message and display it on the frontend.

timobraz:
## Description

Queried for the review using the userID from the passport, and courseID/professorID from the body and checked if there is already a result for this user
## Screenshots

https://github.com/icssc/peterportal-client/assets/35554964/c62aa888-22ef-48d6-a4dc-a88fb9cb10fd

- [ ] Login with an account into the portal
- [ ] Try creating 2 reviews for the same course/professor

## Final Checks:
- [ ] Delete branch

(optional)
- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #378

### DIFF
--- a/api/src/controllers/reviews.ts
+++ b/api/src/controllers/reviews.ts
@@ -5,17 +5,9 @@
 import express from "express";
 import { ObjectID } from "mongodb";
 import { VoteData } from "../types/types";
-import {
-  COLLECTION_NAMES,
-  getCollection,
-  addDocument,
-  getDocuments,
-  updateDocument,
-  deleteDocument,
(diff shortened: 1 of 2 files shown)


js0mmer (approved):
LGTM!

PR #405, opened 2024-01-23
Use passport's userDisplay as source of truth

### LINKED ISSUE #395: Determine review userDisplay on server side
Right now, the display name for a user on their review is determined by the client and sent to the server. We should change it so instead the server determines their display name and the client just specifies whether or not they want to be anonymous. This will prevent users from being able to send whatever they want as their display name in a POST request and force them to use their google account name or be "Anonymous Peter".

timobraz:
## Description

Instead of trusting the client's useDisplay value we instead extract it from the authenticated passport. If userDisplay is Anonymous Peter we keep it as such. 
## Screenshots
image

## Final Checks:

(optional)
- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #395

### DIFF
--- a/api/src/controllers/reviews.ts
+++ b/api/src/controllers/reviews.ts
@@ -134,7 +134,6 @@ router.get('/', async function (req, res, next) {
 router.post('/', async function (req, res, next) {
   if (req.session.passport) {
     //^ this should be a middleware check smh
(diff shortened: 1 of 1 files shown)


js0mmer (approved):
LGTM

Issue #167, opened 2022-02-20
Close AddYear Popup when click away


chenaaron3:
Add year button, edit year button, mobile menu button will stay open even if the user clicks away. Make it so when user clicks away at anything else the menus will close.
![screen-capture-_8_](https://user-images.githubusercontent.com/44650208/155229837-beee9fbc-def1-44a0-b87c-f4b5b0ffc786.gif)

js0mmer:
Use an [OverlayTrigger](https://react-bootstrap-v4.netlify.app/components/overlays/#overlay-trigger).

js0mmer:
Fixed by #406 and #419

PR #406, opened 2024-01-24
Clicking outside closes Quarter and Add Year Popovers

Awesome-E:
## Description

Uses an `OverlayTrigger` around the buttons that create a popover, which allows React Bootstrap to handle the toggle events for these menus.

**Note:** This PR does **NOT** cover the Edit Year popover, as I believe that should be reworked to not include nested menus. Currently, using an OverlayTrigger will close the Year overlay when you click on the input in the Edit Year form because it is outside of the Year overlay.

We could use a custom handler of some sort to determine when to close the overlay trigger, but that's not any better than the what we had before with the Overlays and their handlers, so I'll hold off on changing this.

image

(Clicking the "name" or "start year" fields causes the Overlay Trigger to close)

## Screenshots

image image
Clicking outside of these will close them

## Final Checks:

## Issues

Partially fixes issue #167


site/src/pages/RoadmapPage/AddYearPopup.tsx
```diff
@@ -16,99 +16,95 @@ const AddYearPopup: FC<AddYearPopupProps> = ({ placeholderName, placeholderYear
...
-                  setName(placeholderName);
-                }}
-              >
-                Add Year
-              </Button>
-            </Form>
-          </Popover.Content>
-        </Popover>
-      </Overlay>
+      <OverlayTrigger placement="top" overlay={overlay} trigger={['click']} rootClose>
```
js0mmer, line 102: Can just be a string like the other one for trigger.

```suggestion
      <OverlayTrigger placement="top" overlay={overlay} trigger="click" rootClose>
```

js0mmer (changes requested):
The `showQuarterMenu` and `show` states can be removed since the overlaytrigger now handles that functionality and they're not doing anything.

The popovers do not close when you click a button within them. Previously clicking add year would close the popover. Clicking clear wouldn't close the popover for a quarter but I think it should. A quick hack for this is adding `document.documentElement.click()` to the onClick handlers for these buttons.

js0mmer:
Looks like there are a couple merge conflicts after merging my dark mode PR. The only conflicting changes I made were adding a [class name to the add year form](https://github.com/icssc/peterportal-client/pull/323/files#diff-58755201935ad6399262219ab819635ae118c52796cd35f56dc2983164f27e8f) and using the current theme to determine the [button variant for the quarter popover menus](https://github.com/icssc/peterportal-client/pull/323/files#diff-78e2097498bcf0c22b5f5c92afbc1dd12e37e977d5b144aaab791583b97a250c). Make sure to implement those on top of your current changes when merging changes from master and you should be fine.

Awesome-E:
> The popovers do not close when you click a button within them. Previously clicking add year would close the popover. Clicking clear wouldn't close the popover for a quarter but I think it should.

Good catch. I did call the event to close it when you click "delete", but forgot to do it for "clear"

Awesome-E:
> The showQuarterMenu and show states can be removed since the overlaytrigger now handles that functionality and they're not doing anything.

This isn't completely true. I am using the `onTrigger` to call `showQuarterMenu` and update `show` because this is what allows clicking the buttons in the menu to actually close it. `OverlayTrigger` is handling the clicking outside and putting it into a `onTrigger` event for us and can keep the state on its own, but we want to also be able to have the state separate so we can control it manually for the case I just described.

site/src/pages/RoadmapPage/AddYearPopup.tsx
```diff
@@ -16,99 +16,97 @@ const AddYearPopup: FC<AddYearPopupProps> = ({ placeholderName, placeholderYear
...
-                  setName(placeholderName);
-                }}
-              >
-                Add Year
-              </Button>
-            </Form>
-          </Popover.Content>
-        </Popover>
-      </Overlay>
+      <OverlayTrigger placement="top" overlay={overlay} trigger="click" rootClose>
```
js0mmer, line 104: > This isn't completely true. I am using the onTrigger to call showQuarterMenu and update show because this is what allows clicking the buttons in the menu to actually close it. OverlayTrigger is handling the clicking outside and putting it into a onTrigger event for us and can keep the state on its own, but we want to also be able to have the state separate so we can control it manually for the case I just described.

I see. You are missing the onToggle and show attributes for the add year overlay trigger though.

js0mmer (changes requested):

site/src/pages/RoadmapPage/AddYearPopup.tsx
Awesome-E: I'm assuming my commit just now is what you mean?

js0mmer: Yes.

js0mmer (approved):
🚀

PR #408, opened 2024-01-26
Alert user when they leave without saving

### LINKED ISSUE #303: Alert user when they leave without saving
Frequently, I will make changes to my schedule then forget to save and switch to the catalogue or close the site, losing all my changes. I think an alert should appear asking the user "Are you sure you want to leave? You will lose unsaved changes." when they attempt leave the roadmap page with unsaved changes.

ptruong0:
## Description

Keep track in state of when there are unsaved changes (mark true when planner is validated, and mark false when save button is clicked). Use the onbeforeunload event listener when unsaved changes is true, which shows an alert when the user leaves the page, closes the window, or reloads. Was unable to use useBlocker because of the following issue: 
image
Only the new routers e.g. createBrowserRouter are compatible with useBlocker, see https://reactrouter.com/en/6.21.3/routers/picking-a-router. I didn't want to mess with the existing router, so I used vanilla JS listeners which work fine in this case.

## Screenshots

https://www.loom.com/share/5822237a4a5c428cace0d954b225eb8a?sid=0e1e2bc3-38b4-4e8e-921a-348587ae86a9

## Final Checks:

## Issues

Closes #303


js0mmer (changes requested):
Nice, works for the most part. I see what you mean about react router now.

I ran into a couple issues. When first visiting the roadmap page, if I try to leave without interacting with the planner at all it shows the unsaved changes alert. I think the problem here is that the upon loading the roadmap from mongo or local storage, this useEffect is called a second time.
https://github.com/icssc/peterportal-client/blob/3c7c8da08e68f72c1c076d1a91eaa2e65165dce5/site/src/pages/RoadmapPage/Planner.tsx#L42-L54

Another thing we may want to take into consideration is that if you make a change then undo it, say add a class then remove it, there's still considered to be unsaved changes. Maybe an approach that can resolve these two would be to compare the current state of the roadmap with its previous save.

Switching to the catalogue page with unsaved changes and then attempting to close the tab seems to show the alert appropriately. There appears to be an issue though where if you switch to the catalogue, or another page, then switch back, unsaved changes are lost. This is due to the roadmap loading data again and overwriting any edits to the current state. We would only want it to load data if there's none in the redux slice, not every time the component is mounted. I think I'll put up a separate issue for that.

ptruong0:
On roadmap edit, check whether the current roadmap is equal to the last-saved roadmap in local storage. Only mark as unsaved if they are different

js0mmer (commented):
Nice. For my last issue, I think replacing the isFirstRenderer in the planner's useEffect with a check if the current roadmap is empty would work.

That hook could also be deleted entirely in that case since it is used nowhere else.

ptruong0:
Removing `isFirstRender` causes the `useEffect` to be called repeatedly, but checking for both `isFirstRender` and empty roadmap solves the issue. 

Also, now that I think about it more, having a toast pop up every time the tab is switched would be pretty annoying. The user doesn't really need to be notified constantly that their changes are unsaved. The alert will still show up when the user exits the site from the catalogue tab.

site/src/pages/RoadmapPage/Planner.tsx
```diff
@@ -39,13 +40,27 @@ const Planner: FC = () => {
...
-    if (isFirstRenderer) {
+    // stringify current roadmap
+    const roadmapStr = JSON.stringify({
+      planner: collapsePlanner(data),
+      transfers: transfers,
+    });
+
+    // stringified value of an empty roadmap
+    const emptyRoadmap =
+      '{"planner":[{"startYear":2024,"name":"Year 1","quarters":[{"name":"fall","courses":[]},{"name":"winter","courses":[]},{"name":"spring","courses":[]}]}],"transfers":[]}';
```
js0mmer, line 51: I would rather construct this as a SavedRoadmap object first then stringify. That way we get the type safety and can avoid any bugs down the road when the schema changes (e.g. when we add multiple planner support).

```suggestion
    const emptyRoadmap = JSON.stringify({
      planner: [
        {
          startYear: 2024,
          name: 'Year 1',
          quarters: [
            { name: 'fall', courses: [] },
            { name: 'winter', courses: [] },
            { name: 'spring', courses: [] },
          ],
        },
      ],
      transfers: [],
    } as SavedRoadmap);
```

js0mmer (approved):
Ok, looks good! Just left one suggestion.

PR #413, opened 2024-02-01
Added authorization requirements for posting and deleting reports

### LINKED ISSUE #397: Add authentication/admin checks to report api routes
There is no authentication/admin checking on the reports API routes. The get and delete routes should only be accessible to an authenticated user who is an admin.

timobraz:
## Description

if passport doesn't exist or no admin value on passport then you fail to access GET or DELETE routes. Still works if you're authenticated and admin status.

## Screenshots
Screenshot 2024-01-31 at 6 00 51 PM

## Final Checks:

(optional)
- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #397

### DIFF
--- a/api/src/controllers/reports.ts
+++ b/api/src/controllers/reports.ts
@@ -1,3 +1,4 @@
+import { use } from 'sst/constructs';
 /**
  @module ReportsRoute
 */
@@ -13,6 +14,8 @@ const router = express.Router();
  * Get all reports
  */
 router.get('/', async (req, res) => {
+  if (!req.session.passport) return res.status(401).send('Unathenticated');
+  if (!req.session.passport.admin) return res.status(403).send('Unauthorized');
   const reports = await getDocuments(COLLECTION_NAMES.REPORTS, {}); // get all reports in collection
 
   res.json(reports);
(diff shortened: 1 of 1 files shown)


api/src/controllers/reports.ts
```diff
@@ -13,6 +13,8 @@ const router = express.Router();
  * Get all reports
  */
 router.get('/', async (req, res) => {
+  if (!req.session.passport) return res.status(401).send('Unathenticated');
```
js0mmer, line 16:
```suggestion
  if (!req.session.passport) return res.status(401).send('Unauthenticated');
```

api/src/controllers/reports.ts
```diff
@@ -34,6 +36,8 @@ router.post('/', async (req, res) => {
  */
 router.delete('/', async (req, res) => {
   let status;
+  if (!req.session.passport) return res.status(401).send('Unathenticated');
```
js0mmer, line 39:
```suggestion
  if (!req.session.passport) return res.status(401).send('Unauthenticated');
```

js0mmer (approved):
LGTM! Just left comments on a couple typos that don't really matter

api/src/controllers/reports.ts
timobraz: lmfaooo that's my bad

PR #416, opened 2024-02-01
Re-add jsx-a11y and accessibility fixes

### LINKED ISSUE #415: Re-add jsx-a11y
Re-add jsx-a11y. Previously was included with react-scripts, was omitted when setting up vite + eslint.

js0mmer:
## Description
This PR adds back jsx-a11y (previously included with react-scripts) and fixes relevant accessibility errors
Accessiblity fixes:
* Made search module hit items accessible
  * they use divs so I assigned a "button" role and a tab index so they can be focused on when tabbing through the page
  * added a listener for enter key pressing down in addition to the mouse down listener
* Replaced improper usage of label element in reports page
  * label is for labeling inputs only
* Added alt tags where needed
* Replaced theme toggle anchor tag in sidebar with button
* Refactored how dragging a course from search into the planner works on the roadmap page
  * desktop version previously relied on mouse down listener to set the course that was actively being dragged
    * this isn't accessible - doesn't work if a user isn't using a mouse
  * made use of the drag start listener and the search slice to set the active course that is being dragged
  * mobile version has enter key listener now
* made profile image a button
* made overlaytriggers on roadmap accessible by using a button and using functional form (which always gets rid of strict mode warnings)

## Screenshots

https://github.com/icssc/peterportal-client/assets/8922227/843a7d8d-8a13-4bb8-b2f3-873c818f027c

  - [ ] test tab accessibility - use spacebar to start/end drag for drag/drop components and arrow keys to move
  - [ ] test dragging from search on roadmap

## Final Checks:

(optional)
- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #415


SenghoungLim (commented):
I think overall it is great, especially the catalog page. A few things about the roadmap page, (1) I think when a user tab to the drop down menu on the roadmap year, we should have a rectangle to indicate that the tab is there as well for consistent and visual. 
Gif showing the year drop down tab that has no rectangle indicate selection on it

(2) I think the same with the select verified and add review: 
Gif showing the select verified review and add review do not have rectangle select visual

(3) Lastly, minor thing probably from previous code, we should remove console.log(review):
console log image

js0mmer:
Will merge this as is! Issues with accordion outlines and popover accessibility can be dealt with in a later PR as discussed in today's meeting since they don't fall under the scope of jsx-a11y lint errors.

Issue #252, opened 2023-01-28
Tab Accessibility on RoadMap Page Does Not Work Once We Reach Add Year


SenghoungLim:
Using tab to navigate the website. Once we reach add year, after [enter], we cannot tab to add year or change the year from the drop down menu; it just goes straight to the search bar on the right side.
<img src='https://recordit.co/IFmS3rlhmO.gif' title='Video Walkthrough' width='' alt='Video Walkthrough' />

PR #419, opened 2024-02-08
Make add and edit year a modal

### LINKED ISSUE #398: User Input Menus should be modals, not context menus
Example (Add Year button in planner):
image

Most modern UIs have things that require input (i.e. forms) as either a full screen (its own page) or in a modal where other page interaction is disabled. For the most part, floating context menus should only have buttons that the user can click.

(design to be created)

Awesome-E:
## Description

When you click the "add year" or "edit year" buttons, there will now be a modal window to add/edit details instead of a floating popover. it also combines "add quarter" and "edit year" functionalities into the same menu and allows you to add or remove several years at once a bit faster.

Because this makes changes to the edit year menu, I have also made that use an OverlayTrigger like the add year was previously.

### **BEFORE:**
image
image image

### **AFTER:**
image
image
image

## Final Checks:

## Issues

Closes #398


js0mmer:
Fixes #252

site/src/pages/RoadmapPage/CourseYear.tsx
```diff
@@ -0,0 +1,152 @@
...
+        id={'quarter-checkbox-' + q.id}
+        label={q.name}
+        value={q.id}
+        checked={q.checked ?? false}
+        onChange={() => handleClick(i)}
+      />
+    );
+  });
+
+  const title = type === 'add' ? 'Add Year' : `Editing "${placeholderName}"`;
```
js0mmer, line 63: This runs into the same issue from implementing #368 where for schedules before custom year names were added, their name is undefined. We had it fall back to using the `Year {index + 1}` format in the implementation but given that we are running into this problem again and possibly will in the future, I've drafted up a script that should add default names to any planner with years that are missing names. On first test, it looks good but I might want to find a way to back up the database before proceeding to modify prod. I'll keep you posted.

```ts
import { MongoClient, ObjectId } from 'mongodb';
import dotenv from 'dotenv';

dotenv.config();

const client = new MongoClient(process.env.MONGO_URL);
const DB_NAME = 'peterPortalDevDB';
const db = client.db(DB_NAME);
const roadmaps = db.collection('roadmaps');
const documents = roadmaps.find();

interface Year {
  startYear: number;
  name?: string;
  quarters: object[];
}

interface Roadmap {
  planner: Year[];
  transfers: object[];
}

interface RoadmapDoc {
  _id: ObjectId;
  roadmap: Roadmap;
}

let document = await documents.next() as RoadmapDoc;

while (document != null) {
  let planner = document.roadmap.planner;
  for (let i = 0; i < planner.length; i++) {
    if (planner[i].name == null) {
      planner[i].name = 'Year ' + (i + 1);
    }
  }
  roadmaps.updateOne({ _id: document._id }, {
    $set: {
      roadmap: document.roadmap
    }
  })
  document = await documents.next() as RoadmapDoc;
}

console.log('done')
```

js0mmer (changes requested):
Noted a few things. will take a closer look this weekend

site/src/pages/RoadmapPage/CourseYear.tsx
Awesome-E: I just copied what was used for the popup before, but if I understand correctly, you're saying that `placeholderName` may be undefined for people who created roadmaps a long time ago and that we should have a fallback as "Year {index + 1}" right?

site/src/pages/RoadmapPage/AddYearPopup.scss
```diff
@@ -45,3 +104,33 @@
 h5 {
   font-weight: bold;
```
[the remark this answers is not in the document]
Awesome-E: sure

site/src/pages/RoadmapPage/CourseYear.tsx
```diff
@@ -0,0 +1,152 @@
...
+      if (match) match.checked = true;
+    });
+    return base;
+  };
+
+  const [quarters, setQuarters] = useState<YearPopupQuarter[]>(quarterValues(currentQuarters));
+  const quarterCheckboxes = quarters.map((q, i) => {
+    const handleClick = (i: number) => {
+      quarters[i].checked = !quarters[i].checked;
+      setQuarters(quarters.slice());
```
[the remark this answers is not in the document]
Awesome-E: I'm assuming it's a general react practice to treat state arrays as "frozen"? I can definitely change this though.

site/src/pages/RoadmapPage/CourseYear.tsx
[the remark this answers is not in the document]
Awesome-E: good point — didn't get a chance to update the file name after making the modals

site/src/pages/RoadmapPage/CourseYear.tsx
Awesome-E: I don't have access to `yearIndex` from within the modal so I am putting it in the call to the Modal in `Year.tsx`

site/src/pages/RoadmapPage/AddYearPopup.scss
```diff
@@ -35,6 +36,64 @@
...
 }
 
+.planner-year-modal {
```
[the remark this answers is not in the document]
Awesome-E: yeah that makes sense

site/src/pages/RoadmapPage/CourseYear.tsx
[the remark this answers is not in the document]
js0mmer: Yes, more on that [here](https://react.dev/learn/updating-arrays-in-state).

site/src/pages/RoadmapPage/CourseYear.tsx
js0mmer: > I just copied what was used for the popup before, but if I understand correctly, you're saying that `placeholderName` may be undefined for people who created roadmaps a long time ago and that we should have a fallback as "Year {index + 1}" right?

Yes, any roadmaps made before #368 may not have year names resulting in `placeholderName` being undefined.

site/src/pages/RoadmapPage/YearModal.tsx
```diff
@@ -0,0 +1,154 @@
...
+      { id: 'summer 1', name: 'Summer 1' },
+      { id: 'summer 2', name: 'Summer 2' },
+      { id: 'summer 10 Week', name: 'Summer 10 Week' },
+    ];
+    data.forEach((name) => {
+      const match = base.find((q) => q.id === name);
+      if (match) match.checked = true;
+    });
+    return base;
+  };
```
js0mmer, line 43: This can be moved outside the YearModal functional component. Slight optimization since it will avoid redefining it every render. Also, why not use a record object here mapping id to { name, checked } like:
```ts
{
  'fall': { name: 'Fall' },
  'winter': { name: 'Winter' },
  'spring': { name: 'Spring' },
  'summer 1': { name: 'Summer 1' },
  'summer 2': { name: 'Summer 2' },
  'summer 10 Week': { name: 'Summer 10 Week' },
}
```
Then you can just update it by indexing for each quarter id instead of using `find`.

site/src/pages/RoadmapPage/YearModal.tsx
```diff
@@ -0,0 +1,154 @@
...
+  placeholderName: string;
+  placeholderYear: number;
+  show: boolean;
+  setShow: React.Dispatch<React.SetStateAction<boolean>>;
+  type: 'add' | 'edit';
+  saveHandler: (x: PlannerYearData) => void;
+  currentQuarters: string[];
+}
+
+const CourseYearModal: FC<CourseYearModalProps> = (props) => {
```
js0mmer, line 22: Rename this to YearModal as well and its props.

js0mmer (changes requested):

site/src/pages/RoadmapPage/YearModal.tsx
```diff
@@ -0,0 +1,154 @@
...
+      setQuarters(newQuarters);
+    };
+    return (
+      <Form.Check
+        key={q.id}
+        type="checkbox"
+        id={'quarter-checkbox-' + q.id}
+        label={q.name}
+        value={q.id}
+        checked={q.checked ?? false}
```
js0mmer, line 59: Nullish coalescing with false shouldn't be necessary.
```suggestion
        checked={q.checked}
```

Awesome-E: This is necessary. Without this, React will give a warning that you're now assigning a new prop instead of just changing its value.

site/src/pages/RoadmapPage/YearModal.tsx
Awesome-E: We need access to the keys, and the order matters. I should be able to extract it outside of the component though

site/src/pages/RoadmapPage/AddYearPopup.scss
```diff
@@ -42,6 +43,32 @@
...
+    box-shadow: 0 0 0 0.25rem #0d6efd40;
+    outline: 0;
+  }
+  &:checked {
+    background-color: var(--peterportal-primary-color-1);
+    border-color: var(--peterportal-primary-color-1);
+    &[type='checkbox'] {
+      --bs-form-check-bg-image: url("data:image/svg+xml;charset=utf-8,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 20 20'%3E%3Cpath fill='none' stroke='%23fff' stroke-linecap='round' stroke-linejoin='round' stroke-width='3' d='m6 10 3 3 6-6'/%3E%3C/svg%3E");
+    }
+  }
```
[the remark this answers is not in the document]
Awesome-E: It is (with a few minor changes), but that's the issue I was telling you about – this was the style that wasn't loading when it should have with the React Bootstrap form component.

site/src/pages/RoadmapPage/YearModal.tsx
Awesome-E: This is because undefined is equivalent to not defining the prop, in which case I think adding a comment about this would be good so it's not changed unintentionally.
something like `// must be assigned a value that is not undefiend`

site/src/pages/RoadmapPage/YearModal.tsx
js0mmer: Ok. `Object.keys` and `Object.entries` could be used but I suppose that's no better than the current implementation.

site/src/pages/RoadmapPage/AddYearPopup.scss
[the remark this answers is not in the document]
js0mmer: We are using react bootstrap v4, that's why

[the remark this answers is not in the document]
js0mmer: Everything breaks when migrating to v5 lol

Awesome-E:
Alright, addressed the remaining comments. Let me know if there's anything else you'd like me to change.

site/src/pages/RoadmapPage/YearModal.tsx
```diff
@@ -0,0 +1,155 @@
...
+    { id: 'summer 1', name: 'Summer 1' },
+    { id: 'summer 2', name: 'Summer 2' },
+    { id: 'summer 10 Week', name: 'Summer 10 Week' },
+  ];
+  data.forEach((name) => {
+    const match = base.find((q) => q.id === name);
+    if (match) match.checked = true;
+  });
+  return base;
+};
```
js0mmer, line 36: Being a little pedantic here but this function doesn't read as clearly as I like. I would change some variable names, id and name seem to be used interchangeably. Also, we could probably use null assertion here. Reads a little cleaner and there should never be a case where we select a nonexistent quarter ID, and if there is, we made a mistake and would probably benefit from the error that arises.

```suggestion
const quarterValues: (selectedQuarters: string[]) => YearPopupQuarter[] = (quarterIds: string[]) => {
  const base: YearPopupQuarter[] = [
    { id: 'fall', name: 'Fall' },
    { id: 'winter', name: 'Winter' },
    { id: 'spring', name: 'Spring' },
    { id: 'summer 1', name: 'Summer 1' },
    { id: 'summer 2', name: 'Summer 2' },
    { id: 'summer 10 Week', name: 'Summer 10 Week' },
  ];
  quarterIds.forEach((id) => {
    const quarter = base.find((q) => q.id === id)!;
    quarter.checked = true;
  });
  return base;
};
```

js0mmer (approved):
LGTM! I left a small suggestion on refactoring/renaming variables in the quarterValues function. I will go ahead and give you the approval and leave that up to you if you want to implement (I think the suggested variable names make it a little clearer).

site/src/pages/RoadmapPage/YearModal.tsx
Awesome-E: the `id` is supposed to be consistent with other places in PPC, which have it written as `summer 1` and `summer 2` last I checked. The name can be changed though

Awesome-E: wait never mind you just showed otherwise in the DB.

Awesome-E:
Testing locally, this still seems to work with editing quarters. @js0mmer I think it'd be good if you verify on staging real quick then if it does go ahead and merge it for me.

js0mmer:
LGTM!

PR #431, opened 2024-02-14
Set data-theme attribute at render-time

### LINKED ISSUE #430: When in dark mode, PPC first loads the whole page in light mode
The preference seems to load after the entire page is loaded. It should instead load this preference right when the body tag starts loading so that the user doesn't get a white screen flash of the whole UI while the page is still loading

js0mmer:
## Description
* Set data-theme attribute on body in render-time rather than in useEffect
  * useEffect gets called after first render, so first render of page will be in light theme which is default
  * solution is to set the data-theme on the body in render-time
  * compares with previous dark mode state

## Screenshots

tests reloading the page:

before

https://github.com/icssc/peterportal-client/assets/8922227/a9cb334f-603b-4a32-b057-7cd8a7437e14

after

https://github.com/icssc/peterportal-client/assets/8922227/a99b1d4c-1fc2-4af3-80dd-0f8c1e8bcc8e

## Final Checks:

(optional)
- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #430

### DIFF
--- a/site/src/App.tsx
+++ b/site/src/App.tsx
@@ -43,7 +43,7 @@ export default function App() {
    */
   if (darkMode != prevDarkMode) {
     // Theme styling is controlled by data-theme attribute on body being set to light or dark
-    document.querySelector('body')!.setAttribute('data-theme', darkMode ? 'dark' : 'light');
+    document.body.setAttribute('data-theme', darkMode ? 'dark' : 'light');
     setPrevDarkMode(darkMode);
   }
 
--- a/site/src/App.tsx
+++ b/site/src/App.tsx
@@ -34,6 +34,18 @@ export default function App() {
     usingSystemTheme ? isSystemDark() : localStorage.getItem('theme') === 'dark',
   );
   const [cookies] = useCookies(['user']);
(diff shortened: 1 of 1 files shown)


site/src/App.tsx
```diff
@@ -34,6 +34,18 @@ export default function App() {
...
+  const [prevDarkMode, setPrevDarkMode] = useState(false); // light theme is default on page load
+
+  /**
+   * we run this check at render-time and compare with previous state because a useEffect
+   * would cause a flicker for dark mode users on page load since the first render would be without
+   * the data-theme property set (light would be used by default)
+   */
+  if (darkMode != prevDarkMode) {
+    // Theme styling is controlled by data-theme attribute on body being set to light or dark
+    document.querySelector('body')!.setAttribute('data-theme', darkMode ? 'dark' : 'light');
```
Awesome-E, line 46: Is there a reason we don't just say `document.body`?
```suggestion
    document.body.setAttribute('data-theme', darkMode ? 'dark' : 'light');
```

Awesome-E (approved):
Seems to work as intended! Just one small question about the code, but otherwise that looks good too

PR #432, opened 2024-02-14
Optimize upvote and downvote colors

### LINKED ISSUE #315: Optimize upvote/downvote colors
Currently for the Review component, it does one request on first render to fetch the upvote colors for each review (whether or not they upvoted or downvoted) and then updates the colors for each review.

Additionally, when a user clicks an upvote or downvote, it sends the upvote/downvote request to the backend then it does the fetch again on the colors for each review. This is inefficient and results in a slight delay between the user clicking the upvote/downvote arrow and the color updating. We should fetch the colors for all reviews only that one time on first render. Any upvotes/downvotes the user makes should cause the colors to be updated by changing the color state for that individual review rather than updating them all.

advayanand:
## Description

- An API request is made for only a single review's vote every time it is upvoted or downvoted, instead of for all reviews.
- Changes to vote color and score instantly reflect when a vote is cast, instead of after the API response, which used to cause a delay. If the vote fails to go through, the color and score are reverted to their original state.
- Added aggregateDocuments function to use Mongo's aggregation pipelines to retrieve documents. 

## Screenshots

https://github.com/icssc/peterportal-client/assets/22901073/3cd27b47-b261-4e2d-b2a4-0c44f6fa53cb

## Final Checks:

(optional)
- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #315


api/src/controllers/reviews.ts
js0mmer, line 1: The `/getVoteColor` and `/getVoteColors` routes can be deleted.

site/src/component/Review/Review.tsx
```diff
@@ -160,8 +117,7 @@ const Review: FC<ReviewProps> = (props) => {
               key={review._id}
               course={props.course}
               professor={props.professor}
-              colors={getU(review._id) as VoteColor}
-              colorUpdater={updateVoteColors}
+              userVote={review.userVote!}
```
js0mmer, line 120: I don't think we should need to assert non-null here. All reviews received from backend now have a vote associated with them, right? If so, we can update the type in `types.ts` so that userVote is required.

site/src/types/types.ts
```diff
@@ -152,7 +153,8 @@ export interface MongoRoadmap {
 }
 
 export interface VoteColor {
-  colors: boolean[];
+  colors?: boolean[];
```
js0mmer, line 156: Does this interface actually get used anywhere? If not, we can delete it.

site/src/component/Review/Review.tsx
js0mmer: Additionally, why specify it as a prop? We already pas the review prop so we can just do `review.userVote` within the SubReview component.

js0mmer (changes requested):
Looks good visually. Left some comments on a little bit of refactoring that could be done. Also, that aggregation pipeline looks insane. I have no idea what I'm looking at but I'll trust you know what you're doing lol 🔥

site/src/types/types.ts
```diff
@@ -21,6 +21,7 @@ export interface ReviewData {
   tags: string[];
   verified?: boolean;
   captchaToken?: string;
+  userVote?: number;
```
js0mmer, line 24: Going off my previous comment. I see now that all reviews retrieved from the reviews route will have this attribute but reviews retrieved from the featured will not.

I think we could make this distinction clear by having another type for `FeaturedReviewData` on the frontend.

e.g. so userVote can be required here on ReviewData and featured review will be the following

```ts
export type FeaturedReviewData = Omit<ReviewData, 'userVote'>;
```

which is `ReviewData` but with the `userVote` attribute omitted.

```suggestion
  userVote: number;
```

site/src/component/Review/Review.tsx
advayanand: Done

site/src/types/types.ts
advayanand: Done. I created the `FeaturedReviewData` type and used it in ProfessorPopup.tsx, which was the only file that was using the `/featured` endpoint.

site/src/component/Review/Review.tsx
```diff
@@ -160,8 +131,7 @@ const Review: FC<ReviewProps> = (props) => {
               key={review._id}
               course={props.course}
               professor={props.professor}
-              colors={getU(review._id) as VoteColor}
-              colorUpdater={updateVoteColors}
+              updateScore={(newUserVote) => updateScore(review._id!, newUserVote)}
```
js0mmer, line 134: Do we need to pass the `updateScore` function? Or could we just define this function within the SubReview since it just uses the redux slice? Also this could maybe be a new reducer on the slice instead.

js0mmer (approved):
LGTM

PR #426, opened 2024-02-13
Senghoung/edit review feature

SenghoungLim:
## Description
- Added edit review feature in user's review menu/tab.
- Added delete confirmation dialog for user's review.
- Added user-friendly design of the edit pen icon accommodating dark mode.

## Screenshots
- Showcase of delete review & pen UI for dark/light mode
Showcase of delete review & pen UI for dark/light mode
- Showcase delete review confirmation & dark/light mode UI
Showcase delete review confirmation & dark/light mode UI

## Steps to verify/test this change:
- [ ] Verify changes work as expected on staging instance

## Final Checks:
- [ ] Verify successful deployment

(optional)
- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #273

PR #440, opened 2024-02-25
Changelog Modal

ptruong0:
## Description

Created a modal element that contains a text description and image. Should only appear once since I set a variable in local storage that prevents the modal from being shown the next time the website is loaded.

## Screenshots
image

## Final Checks:

(optional)

- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #


ptruong0:
The content can be changed later, does the UI look okay?

Awesome-E:
Same thing as I said here for the most part: https://github.com/icssc/peterportal-client/pull/426#issuecomment-1943080903

I think it should be consistent with #419's modal with the background color, no borders, and large X button. I believe I also used a filled background with white text for the button, though that was more of a call to action than this one.

Layout other than that looks fine to me.

ptruong0:
Do you think the close button is still necessary? Right now, there are 3 ways to close the modal: (1) X button, (2) Close button at the bottom, (3) Clicking outside the modal.

js0mmer:
I'd say it's unnecessary.

site/src/component/ChangelogModal/ChangelogModal.tsx
```diff
@@ -0,0 +1,58 @@
...
+import Row from 'react-bootstrap/Row';
+
+const DESCRIPTION = 'You can now view recently added features to the PeterPortal website, listed in this modal.';
+const IMAGE_URL =
+  'https://media.tenor.com/ufm_0t3ACEAAAAAM/ginger-cat-ginger-cat-eating-then-staring-at-the-camera.gif';
+
+const ChangelogModal = () => {
+  const [showModal, setShowModal] = useState(false);
+
+  const currentMonth = new Date().toLocaleString('default', { month: 'long', year: 'numeric' });
```
js0mmer, line 14: We should replace this with an additional constant above for the date (rather than just the month) of the last changelog update. This will then be manually updated anytime the description and image are updated. That way, this can support multiple updates to the changelog within a month.

js0mmer (changes requested):

site/src/component/ChangelogModal/ChangelogModal.tsx
```diff
@@ -0,0 +1,49 @@
...
+    <div onClick={closeModal}>
+      <Modal
+        className="changelog-modal"
+        show={showModal}
+        centered
+        animation={false}
+        onClick={(e: MouseEvent) => e.stopPropagation()}
+      >
+        <Modal.Header closeButton>
+          <h2>What's New - {new Date().toLocaleString('default', { month: 'long', year: 'numeric' })}</h2>
```
js0mmer, line 39: I think we should show the month/year of the last update rather than the current month/year.

```suggestion
          <h2>What's New - {new Date(LAST_UPDATED).toLocaleString('default', { month: 'long', year: 'numeric' })}</h2>
```

site/src/component/ChangelogModal/ChangelogModal.tsx
```diff
@@ -0,0 +1,49 @@
...
+      localStorage.setItem('changelogSeen', LAST_UPDATED);
+    }
+  }, []);
+
+  const closeModal = () => {
+    setShowModal(false);
+  };
+
+  return (
+    <div onClick={closeModal}>
```
js0mmer, line 30: Don't think this extra div is necessary.

site/src/component/ChangelogModal/ChangelogModal.tsx
```diff
@@ -0,0 +1,49 @@
...
+      <Modal
+        className="changelog-modal"
+        show={showModal}
+        centered
+        animation={false}
```
js0mmer, line 35: Close button doesn't work. I think this should make it work.
```suggestion
        animation={false}
        onHide={closeModal}
```

site/src/component/ChangelogModal/ChangelogModal.scss
```diff
@@ -0,0 +1,37 @@
...
+    font-size: 1.2rem;
+    margin: 0;
+  }
+
+  .modal-img {
+    margin: 1rem 0;
+    max-width: 70vw;
+    max-height: 50vh;
+    object-fit: contain;
+  }
```
js0mmer, line 29: Close button font size can be adjusted to make it bigger/match the year modal.
```suggestion
  }
  
  button.close {
    font-size: 32px;
  }
```

js0mmer (changes requested):

site/src/component/ChangelogModal/ChangelogModal.tsx
Awesome-E: Agree with this

site/src/component/ChangelogModal/ChangelogModal.tsx
ptruong0: The extra div is to allow clicks outside the modal to close the modal.

js0mmer (approved):
LGTM! I think for now we can just comment out the ChangelogModal within the App component. Once we add something new, someone can uncomment it and write up a changelog.

Awesome-E:
The padding of this modal looks different from the padding of the year modal, but I think this is not as big of a deal right now (I should make a PR later to use those styles for all modal elements instead of asking each PR author to copy code to make theirs consistent)

However, I do think you should make the modal enter/exit smoothly like the add year one instead of just disappearing instantly when the close button is clicked

PR #441, opened 2024-02-28
Replace `react-device-detect` with media query hooks

### LINKED ISSUE #422: Small screen layouts should not be exclusive to touch-first (mobile) devices
Use case: opening a split view window on desktop, I would much rather have the mobile layout than this. I think we should be using CSS media queries rather than just checking what device type the client is on when the page first loads.

image

(steps to reproduce is same as the use case)

Awesome-E:
## Description
- Create media query hooks for screen width detection
- Replace existing `isMobile` and `isDesktop` checks
- Remove the `react-device-detect` package.

## Screenshots
_This now changes layout due to screen width, not the user agent_
image

## Final Checks:

## Issues

Closes #422


js0mmer (approved):
lgtm!

PR #444, opened 2024-02-29
Add mongoose

timobraz:

## Final Checks:

(optional)

- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #


js0mmer:
Looks like changes from master will need to be merged first before it can deploy.

api/src/app.ts
```diff
@@ -105,14 +106,43 @@ app.use(function (req, res) {
...
+      dbName: DB_NAME,
+      serverSelectionTimeoutMS: 5000,
+    });
+  }
+  return conn;
+};
+app.use(async function (req, res, next) {
+  await connect();
+  next();
+});
```
js0mmer, line 123: I don't think we need this middleware since we connect on lambda setup or before starting the dev server.

```suggestion
```

api/src/app.ts
```diff
@@ -41,7 +42,7 @@ if (process.env.MONGO_URL) {
     collection: COLLECTION_NAMES.SESSIONS,
   });
   // Catch errors
-  store.on('error', function (error) {
+  mongoose.connection.on('error', function (error) {
```
js0mmer, line 45: I think we still want to listen to connect-mongodb-session store errors here.

api/package.json
```diff
@@ -18,6 +18,7 @@
     "express-session": "^1.17.1",
     "mongodb": "^6.3.0",
     "mongodb-client-encryption": "^6.0.0",
```
js0mmer, line 20: We should be able to remove `mongodb` and `mongodb-client-encryption` (not entirely sure about the second one).

api/src/controllers/reports.ts
js0mmer, line 1: When i make a new report, it doesn't show up in the reports.

image

edit: may have to do with `reviewId` instead of `reviewID` on the model.

js0mmer: errors in network tab

image

api/src/models/vote.ts
```diff
@@ -0,0 +1,21 @@
...
+  },
+  reviewID: {
+    type: String,
+    ref: 'Review',
+  },
+  timestamp: {
+    type: Date,
+    required: true,
+    default: Date.now,
+  },
```
js0mmer, line 16: should also have a score attribute with value being 1 or -1 (1 being upvote, -1 being downvote)

api/src/models/report.ts
```diff
@@ -0,0 +1,21 @@
+import mongoose from 'mongoose';
+
+const reportSchema = new mongoose.Schema({
+  reviewId: {
```
js0mmer, line 4: it seems like it is capitalized everywhere else and case matters
```suggestion
  reviewID: {
```

js0mmer (changes requested):

api/src/models/vote.ts
js0mmer: upvoting and downvoting is also broken, i think this is the cause

api/out.txt
js0mmer, line 1: remove this

js0mmer (commented):
LGTM. need to find a time to take prod down and update db then merge

js0mmer (approved):

PR #454, opened 2024-03-03
Update Sidebar Styles Across Mobile and Desktop

Awesome-E:
## Description
- Consolidate all **navigation** links into the sidebar
- Remove any **user-related** actions (settings, links) from the old sidebar
- Show the user popover in the top right on mobile → this makes navigation between mobile and desktop consistent
- Makes the styling consistent between the mobile and desktop sidebars

### Changes since this post
- [x] Close button is kind of just hanging there, there is space for something to the left of it. (logo or text?)
- [x] User popover is not updated yet

## Screenshots

Desktop Collapsed Sidebar Before & After:
 

Mobile Collapsed Sidebar Before & After:
 

Mobile Expanded Sidebar Before & After:
 

Desktop Sidebar (previously removed):

## Final Checks:

## Issues

Closes #

### DIFF
--- a/site/src/component/AppHeader/AppHeader.scss
+++ b/site/src/component/AppHeader/AppHeader.scss
@@ -30,7 +30,7 @@
   height: 8vh;
   background-color: var(--overlay1);
   padding-right: 3vw;
-  padding-left: 2vw;
+  padding-left: 16px;
   display: flex;
   top: 0;
 
@@ -58,6 +58,7 @@
       align-items: center;
       border: none;
       background: none;
+      padding: 4px;
     }
 
     .desktop-toggle {
--- a/site/src/component/AppHeader/AppHeader.scss
+++ b/site/src/component/AppHeader/AppHeader.scss
@@ -111,7 +111,7 @@
       right: 0;
       margin-left: auto;
       margin-right: auto;
-      width: 7vw;
+      width: fit-content;
     }
 
     @media only screen and (max-width: 800px) {
@@ -160,7 +160,6 @@
         right: 0;
         margin-left: auto;
         margin-right: auto;
-        width: 18vw;
       }
 
       .school-term {
--- a/site/src/component/AppHeader/AppHeader.scss
+++ b/site/src/component/AppHeader/AppHeader.scss
@@ -1,6 +1,9 @@
 #peterportal-logo {
   width: auto;
   height: 32px;
+  @media (min-width: 800px) {
+    height: 36px;
+  }
 }
 
 [data-theme='dark'] {
--- a/site/src/component/AppHeader/AppHeader.scss
+++ b/site/src/component/AppHeader/AppHeader.scss
@@ -64,47 +64,6 @@
       padding: 4px;
     }
 
-    .desktop-toggle {
-      display: contents;
-    }
-
-    .mobile-toggle {
-      display: none;
-    }
-
-    .navbar-toggle {
-      background-color: var(--overlay2);
-      margin-left: 5vw;
-      display: flex;
-      flex-direction: row;
-      align-items: center;
-      height: 60%;
-      border-radius: 1rem;
-      width: 12vw;
-
-      .navbar-toggle-item {
-        width: 50%;
-        height: 100%;
-        text-align: center;
-        display: flex;
-        justify-content: center;
-        align-items: center;
-        border-radius: 1rem;
-
-        a {
-          color: var(--text);
-        }
-      }
-
-      .active {
-        background-color: var(--peterportal-secondary-orange);
-
-        a {
-          color: white;
-        }
-      }
-    }
-
     .navbar-logo {
       display: flex;
       flex-direction: row;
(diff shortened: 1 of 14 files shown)


Awesome-E:
Updated the user popover and added the logo on the top left of the sidenav to fill the empty space. I think this looks better:

Awesome-E:
I think there may be a better way to implement the sidenav (would it be better as an overlay or something similar?), but what I have changes the existing the code the least (and I wasn't sure if things like overlays would be better for a full-height element) so I stuck with that.

js0mmer:
I think the implementation is fine as is, don't think an overlay would be any better.

js0mmer (commented):
Looks nice! Just some initial observations:
- styling is all messed up for courses in the roadmap, the "footer" class that's being used there should be renamed to something unique and have its styles restored

image

- is there a reason you use px for font sizes instead of a relative unit like em/rem?
- peterportal logo in the header has shrunk quite a bit and just looks odd to me lol. was that intentional?
- an alternative for the logo in the sidebar would be the favicon. nothing wrong with the full logo though
- i think the link to the feedback form (and all the other links in the footer) should open in a new tab
- there is an edge case where having just the week # in the header might be unclear
  - summer session 2 starts during summer 10 week, so week 1 of session 2 is like week 5 of the 10 week session
  - the API handles this appropriately in the display text, but by splicing that it, we're just showing the week # for one of the quarters ([api docs](https://docs.api-next.peterportal.org/docs/developers-guide/rest-api/reference/week))

Awesome-E:
> peterportal logo in the header has shrunk quite a bit and just looks odd to me lol. was that intentional?

I changed it from `vw` to a fixed size, so on most screens it is actually larger. This is intentional, but I can bump it up slightly or clamp it between two widths instead of locking it to one.

> i think the link to the feedback form (and all the other links in the footer) should open in a new tab

Yes, thank you for reminding me! I did intend to do this.

> there is an edge case where having just the week # in the header might be unclear

hmm that's interesting. I think we should come up with a better solution than putting it in the topnav though, so it's more of a temporary measure (and would hopefully be resolved before summer :wink:)

Awesome-E:
Updated/Fixed the header image size and the roadmap page.

Before:
image
After:
image

js0mmer:
> It is easier for me to think of things in px instead of rem. Additionally, I don't really find the argument for rem too convincing since the root element is almost always constant anyways (in PPC it's 14px), so it's just more conversion.

Well the main argument for em/rem is accessibility. By default 1rem is 16px (or I guess 14 here, idk why, that does make the math a little annoying), but someone who has trouble seeing might bump that up in a font size setting on their browser/device.

> I changed it from vw to a fixed size, so on most screens it is actually larger. This is intentional, but I can bump it up slightly or clamp it between two widths instead of locking it to one.

I see, yeah that is much better on mobile. Thanks for fixing the usage of vw/vh, also seemed weird that everything would shrink in height/width based on screen. But yeah bumping it up so the desktop one is closer in size to what we had before would be nice.

edit: looks like you beat me to it

Awesome-E:
> Well the main argument for em/rem is accessibility. By default 1rem is 16px (or I guess 14 here, idk why, that does make the math a little annoying), but someone who has trouble seeing might bump that up in a font size setting on their browser/device.

I see, that makes sense. I assumed that Zoom would fix that but I see how it can be more than just that. Is it safe to assume this is primarily for font sizes and not for other measurements then?

js0mmer:
Yeah, I think that's a safe assumption.

Awesome-E:
Re-requesting review since I'm not sure if you got a chance to review the entire PR yet.

I don't think it's worth the extra effort in changing any additional `font-size` properties to `rem` since the root needs to change size later _and_ there are 20 ish other instances of `font-size` being in `px` as well (it's not an odd one out).

Seems like it would be better to just leave it as is so it doesn't break when changing the root font size. Will just do all of the font size fixes in a future PR

site/src/component/AppHeader/Profile.scss
js0mmer, line 1: `divider-before` class is unused now. it's selector can be removed.

site/src/component/SideBar/Sidebar.scss
js0mmer, line 1: selectors for `theme-button` class can be removed

site/src/component/AppHeader/AppHeader.scss
js0mmer, line 1: selectors related to the course/professor toggle can be removed

`navbar-toggle`, `mobile-toggle`, `desktop-toggle` classes

js0mmer (changes requested):
When viewing on my laptop, there's a scrollbar now on the root.

image

site/src/component/AppHeader/Profile.tsx
```diff
@@ -27,15 +27,15 @@ const Profile = () => {
   const DefaultTab = (
     <>
       <div className="profile-popover__header">
-        <img src={picture} alt={name} />
-        <span>
-          <h1 title={name}>{name}</h1>
```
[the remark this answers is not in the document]
Awesome-E: Do you have an example of a name/email that would do this? The font size was changed so more text should fit in the box, and I also tested it with a few longer names. I will put this back though since that reason makes a lot of sense to me.

Awesome-E:
> When viewing on my laptop, there's a scrollbar now on the root.

Oh, did not realize this was not there before. I will take a look.

site/src/component/AppHeader/Profile.tsx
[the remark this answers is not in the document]
js0mmer: Any full name from a culture that uses 4 names would probably overflow here easily.

site/src/component/SideBar/Sidebar.scss
Awesome-E: this doesn't seem to be entirely true, as it's still present in the user popover and the styles to underline it are still the only ones causing it to underline on hover. I am removing it from the sidebar styles as suggested though

Issue #266, opened 2023-02-25
Multiple planner on roadmap page


MFYLM:
On roadmap page, it would be nice to allow users to have multiple four year plans to try out.

ethanwong16:
<img width="1439" alt="Screen Shot 2023-02-25 at 18 00 34" src="https://user-images.githubusercontent.com/57075492/221388186-0a00e933-bb9d-448a-8cbc-b6ddb126fd47.png">

Currently users only have one planner where they add their years to.  This PR should explore allowing users to have multiple planners, each capable of containing multiple years.

The first step should be to update the `roadmapSlice.ts` redux file to add support for multiple planners, similar to how there is currently support for multiple years and quarters.

The second step is to create front end changes to add toggling between planners (similar to how AntAlmanac allows toggling between schedules).

PR #455, opened 2024-03-04
Allow the user to import transcripts into the Roadmap Planner

### LINKED ISSUE #443: Import unofficial transcript
Allow users to import unofficial transcript into their roadmap

Awesome-E:
## Development Notes
Any comments would be very helpful!
- [ ] This would be good to release together with #266 since it currently overrides the user's schedule. (This is still open to discussion though)
- [x] Lint checks will currently be failing due to updating older files that have ESLint warnings in them.
- [x] Styles are still using the year modal styles, so this needs to be refactored.
- [x] Button placement and other layout/styles are not finalized yet.

## Description
- Add package to parse the HTML without injecting content into the page
- Extract the transferred credits as well as quarters and courses taken at UCI from the HTML Transcript
- Group these quarters into academic years (in a planner-compatible format)
- Set the current roadmap's planner based on these year entries

## Screenshots

Import Dialog with Information (styles and WIP)
image

## Final Checks:

## Issues

Closes #443


Awesome-E:
@js0mmer do you think you could take a look at the files failing the lint check and update the dependency arrays for them?

js0mmer:
Yeah

Awesome-E:
Interesting, yours allowed you to commit even with the dependency array warnings, but mine didn't.

js0mmer:
Yeah, formatting just seemed to be messed up somehow.

js0mmer:
I will leave the pre-existing dependency warnings to be resolved later, don't want to move too much stuff around in this PR.

Awesome-E:
> I will leave the pre-existing dependency warnings to be resolved later, don't want to move too much stuff around in this PR that's out of scope.

yeah it's not giving an error on commit anymore so not going to worry about it

Awesome-E:
Going to request review on this once we decide what the behavior of the import should be.

js0mmer:
Now that #440 is merged, I think this would be a great first feature to add to the changelog modal!

Awesome-E:
Included the changes we discussed yesterday regarding quarters and CC course credits @js0mmer 
image

site/src/pages/RoadmapPage/ImportTranscriptPopup.tsx
```diff
@@ -0,0 +1,243 @@
...
+async function htmlFromFile(file: Blob): Promise<HTMLElement> {
+  const reader = new FileReader();
+  const fileText = (await new Promise((resolve) => {
+    reader.addEventListener('load', () => resolve(reader.result as string));
+    reader.readAsText(file);
+  })) as string;
+  return parseHTML(fileText);
+}
+
+type CourseLookup = { [k: string]: CourseGQLData };
```
js0mmer, line 72: this is a duplicate of the BatchCourseGQLData type

site/src/pages/RoadmapPage/ImportTranscriptPopup.tsx
```diff
@@ -0,0 +1,243 @@
...
+  }).then((x) => x.json());
+
+  // Query Response is { data: { [key: string]: [CourseData] } }
+  // Transform this response into a lookup { courseID: {CourseData} }
+  return Object.fromEntries(
+    Object.values<CourseGQLData[]>(response.data)
+      .flat()
+      .map((c) => [c.department + c.courseNumber, c]),
+  );
+}
```
js0mmer, line 108: use `searchAPIResults` helper function instead

site/src/pages/RoadmapPage/Planner.tsx
```diff
@@ -309,6 +307,9 @@ const Planner: FC = () => {
         placeholderName={'Year ' + (data.length + 1)}
         placeholderYear={data.length === 0 ? new Date().getFullYear() : data[data.length - 1].startYear + 1}
       />
+      <ImportTranscriptPopup />
+      <br />
+      <br />
```
js0mmer, line 312: What are the 2 `<br />`s for?

Awesome-E: was messing with spacing but it doesn't do anything if there's not enough content on the left anyways so gonna remove this

site/src/pages/RoadmapPage/ImportTranscriptPopup.tsx
Awesome-E: It's not a lookup by course ID, it's a lookup by department and code. It seems I could get the ID by simply removing all the spaces, but that seems like a fragile solution, even if that is consistent for every entry in the API.

I do think it would be good to not have this duplicate code though – do you have any suggestions for the best way to fix this?

js0mmer: Replacing spaces with empty strings seems fine to me. That's how the API's scrapers generate the course id.

Awesome-E: > that seems like a fragile solution, even if that is consistent for every entry in the API

To be honest I still stand by this but I think this is the best we can do for now.

Awesome-E:
Still not a huge fan of just removing spaces but I do agree that it's better than essentially reusing 40 lines of code.

js0mmer (approved):
lgtm

PR #458, opened 2024-03-06
Prerequisite unmet  popup in search results

### LINKED ISSUE #451: Add prerequisite met indicator within search module
Add an indicator within the search module that shows whether prerequisites have been met for this course. Can help when finding classes to add to planner.

Can be have a tooltip similar to what currently exists in the planner that tells you what prerequisites you're missing.

timobraz:
## Description
When searching for a course, an indicator may show up in the search bar if you haven't met prerequisites from your current roadmap.

## Screenshots

https://github.com/icssc/peterportal-client/assets/35554964/fc16ea05-c53d-4c2f-bad3-ea08e39f2a13

## Final Checks:

(optional)

- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #451


Awesome-E:
I think the icon itself is not very easy to tell what it means. Why not just make it consistent with what happens if you actually put it on the roadmap, that is, use the warning sign icon (at least for now)?

js0mmer (commented):
I think the warning icon should be moved to the bottom left so it is consistent with the roadmap. Additionally, top right would conflict with #462.
Why not reuse the existing warning popover/logic?

timobraz:
> I think the warning icon should be moved to the bottom left so it is consistent with the roadmap. Additionally, top right would conflict with #462. Why not reuse the existing warning popover/logic?

The current popover logic relies on finding invalid courses that are **already** in the roadmap by going through each course and creating an array of objects that have `location` and `required` fields.  

This will probably have to be refactored in a PR of a different scope because the current way the roadmap detects unmatched prerequisites isn't reusable to `SearchHitItem`.

js0mmer (commented):
Looks good. My one concern though is that the logic for this is implemented in SearchHitContainer component which is also reused for the course/professor catalog pages. It doesn't seem to be cause any errors though.

site/src/component/SearchHitContainer/SearchHitContainer.tsx
```diff
@@ -22,15 +22,33 @@ const SearchResults = ({
   CourseHitItem,
   ProfessorHitItem,
 }: Required<SearchHitContainerProps> & { results: CourseGQLData[] | ProfessorGQLData[] }) => {
+  const roadmap = useAppSelector((state) => state.roadmap);
+  const allExistingCourses = roadmap?.yearPlans.flatMap((yearPlan) =>
+    yearPlan.quarters.flatMap((quarter) => quarter.courses),
+  );
```
js0mmer, line 28: Needs a fix for the multiplanner.

```suggestion
  const allExistingCourses = roadmap?.plans[roadmap.currentPlanIndex].content.yearPlans.flatMap((yearPlan) =>
    yearPlan.quarters.flatMap((quarter) => quarter.courses),
  );
```

js0mmer (changes requested):

js0mmer (approved):

PR #462, opened 2024-03-07
Make the quarters offered tooltip more useful

### LINKED ISSUE #295: Make the quarters offered tooltip more useful
Right now the tooltip shows what quarters a class was offered in 2020.
(edit: this is literally hardcoded to show only 2020 terms, see [CourseQuarterIndicator.tsx](https://github.com/icssc/peterportal-client/blob/master/site/src/pages/SearchPage/CourseQuarterIndicator.tsx))

quarteroffered

I feel like this stat isn't super useful in 2023, especially when a lot of classes weren't even offered in 2020 and no information is shown for them. A few ideas I have to make this more useful are:
## Calculate this stat for the previous academic year rather than 2020
  * pro: more recent/relevant
  * con: the issue of no quarter offering info for a class would persist

or

## Show the most recent year the class was offered in each of the quarters
  * pro: all classes would now have quarter offering info
  * con: could be a bit verbose depending on how the information is presented
  * usage of icons could help with quickly identifying quarters (i.e. 🍂 for fall, ❄️ for winter, etc.)
  * maybe would want to order the list of quarters by recency of the last year offered

or

## Show how many times the class has been offered in each of the quarters for the past x years
  * an optimal value of x would have to be chosen to balance having recent/relevant quarter offerings while also making sure information is available for most courses. I'm thinking something small, around 3.
  * could present as just text or maybe a chart.
  * if displaying as text: display quarters in decreasing order by the number of times the class was offered that quarter
  * same idea as above with using icons could be applicable

I think the quarter offerings should also be available on the course page (the red dot I placed in the corner is a suggestion) so the user doesn't have to go back to the search page to view them. Perhaps the course page could go a bit more in depth on this information too, but I think this kind of ties in with #201.

quarterofferedinfo

ptruong0:
## Description

Changed the colored dots to emojis that represent the quarter, and used the previous academic year (e.g. fall 2022 & winter/spring/summer 2023) instead of 2020. Inside the tooltip on hover, added a chart that shows the frequencies of each term in the past. Also added the tooltips to the course page and the course cards on the roadmap page.

## Screenshots
image
### Tooltip on hover
image
### Not offered in the previous year
image
### No recent term data
image
### Course page
image
### Roadmap page
image

## Final Checks:

(optional)

- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #295


site/src/component/QuarterTooltip/Chart.tsx
```diff
@@ -0,0 +1,170 @@
...
+    let fallCount = 0,
+      winterCount = 0,
+      springCount = 0;
+
+    // for summer, count unique years rather than total terms (e.g. count SS1 2023 and SS2 2023 as one)
+    const summerYears = new Set<string>();
+
+    this.props.terms.forEach((data) => {
+      const year = data.split(' ')[0];
+      const term = data.split(' ')[1];
```
js0mmer, line 45: i think you could do this with one line:

```suggestion
      const [year, term] = data.split(' ');
```

js0mmer (commented):
LGTM! Excited for this feature. I want to wait for the api team to fix the terms list before we merge though.

js0mmer (approved):
Looks like we're good to go.

ptruong0:
> LGTM! Excited for this feature. I want to wait for the api team to fix the terms list before we merge though.
> 
> Note: I see that the chart is mostly copy/pasted from the gradedist. This is fine for now. In the future, we probably want to refactor to eliminate redundancy (and switch from class component to functional).

Yeah, I was wondering why that component in particular was a class component. I just copied over the component instead of reusing because there were a lot of small chart settings that I changed

PR #470, opened 2024-05-01
pnpm

js0mmer:
## Description
- switched to pnpm
  - updated scripts
  - updated actions workflows
  - deleted package-locks
  - added pnpm-workspace file
  - updated readme
- updated a few packages
- fixed lint/tsc errors

- [ ] test locally
  1. delete node_modules folders
  2. run `pnpm install`
  3. test dev `pnpm dev`
  4. test making a commit

## Final Checks:

(optional)

- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #


api/src/controllers/reviews.ts
```diff
@@ -95,7 +94,7 @@ router.get('/', async function (req, res) {
     courseID,
     professorID,
     userID,
-    _id: reviewID === undefined ? undefined : new ObjectId(reviewID),
+    _id: reviewID === undefined ? undefined : reviewID,
```
Awesome-E, line 97: not sure why this is a ternary condition as it functionally does nothing
```suggestion
    _id: reviewID,
```

api/package.json
Awesome-E, line 1: in the `dev` script, it looks like there's still a non-pnpm command, though it looks trivial to me.

site/src/component/SearchModule/SearchModule.tsx
```diff
@@ -22,7 +22,7 @@ interface SearchModuleProps {
 const SearchModule: FC<SearchModuleProps> = ({ index }) => {
   const dispatch = useAppDispatch();
   const search = useAppSelector((state) => state.search[index]);
-  const [pendingRequest, setPendingRequest] = useState<NodeJS.Timeout | null>(null);
+  const [pendingRequest, setPendingRequest] = useState<number | null>(null);
```
Awesome-E, line 25: NodeJS.Timeout seems more clear from a readability standpoint so that I'd know what to do with this, but I'm fine changing to number for consistency's sake

api/src/controllers/reviews.ts
```diff
@@ -95,7 +94,7 @@ router.get('/', async function (req, res) {
...
     verified: verified === undefined ? undefined : verified === 'true' ? true : false,
```
Awesome-E, line 98: I feel like this can be written better but it's not something that originally came up in the PR so I don't think it has to be addressed right now.

Awesome-E (changes requested):
Just a few minor things.

Also for some reason, running `pnpm install` gives me a massive diff on the `pnpm-lock.yaml` file, not sure if there's a cause for that

site/src/component/SearchModule/SearchModule.tsx
js0mmer: Timeout type/NodeJS namespace no longer exists when using pnpm

js0mmer:
pnpm-lock issues look like they're resolved now

Awesome-E (approved):

PR #472, opened 2024-05-07
Add support for multiple planners

### LINKED ISSUE #266: Multiple planner on roadmap page

Awesome-E:
> *This UI has a starting point of having the multi-planner being able to switch schedules within the browser*

## Description
- [x] Consolidated the UI placement of the course picker in the topbar.
- [x] Edit and delete buttons now work for individual plans rather than just the current plan
- [x] Potentially make it so that opening an edit or delete dialog does not close the dropdown?
- [x] Make it save to `localStorage`
- [x] Make it save to the server
- [x] Make it so that it auto-upgrades old planners

## Screenshots

Starting point (before):
image

WIP 1:

After:
multiplanner

- [x] Adding a new roadmap works
- [x] Adding a third roadmap should not default to the same name as the second one if it's called "Roadmap 3"
- [x] Removing the second of 3 roadmaps while the last one is active should work and keep the current one active
- [x] Removing the second of 2 roadmaps while the first is active should work and keep the current one active
- [x] (Add one more roadmap so you have two again) Removing the second while it's active should show the first one after.
- [x] Removing the only roadmap should default you back to a new blank "Peter's Roadmap"
- [x] Saving when signed out works
- [x] Saving to the cloud works

## Final Checks:

## Issues

Closes #266


Awesome-E:
Draft PR is set up – I have task progress as well as some basic acceptance criteria for this feature listed in the PR description.

Awesome-E:
Going to save the dialog interaction and checking whether something is saved for another PR unless this one doesn't get reviewed before the meeting

site/src/pages/RoadmapPage/Planner.tsx
```diff
@@ -35,21 +37,23 @@ const Planner: FC = () => {
...
-  const data = useAppSelector(selectYearPlans);
+  const currentPlanData = useAppSelector(selectYearPlans);
+  const allPlanData = useAppSelector(selectAllPlans);
   const transfers = useAppSelector((state) => state.roadmap.transfers);
 
   const [missingPrerequisites, setMissingPrerequisites] = useState(new Set<string>());
 
   useEffect(() => {
     // stringify current roadmap
+    // console.log(allPlanData, currentPlanData)
```
js0mmer, line 48:
```suggestion
```

site/src/pages/RoadmapPage/Planner.tsx
```diff
@@ -35,21 +37,23 @@ const Planner: FC = () => {
...
     const roadmapStr = JSON.stringify({
-      planner: collapsePlanner(data),
+      planners: collapseAllPlanners(allPlanData),
       transfers: transfers,
     });
 
     // stringified value of an empty roadmap
     const emptyRoadmap = JSON.stringify({
-      planner: [defaultYear()],
+      planners: [{ name: "Peter's Roadmap", content: [defaultYear()] }],
```
js0mmer, line 56: I think it would be good to have a constant for "Peter's Roadmap" since I assume it is used elsewhere to initialize the planner.

site/src/pages/RoadmapPage/Planner.tsx
```diff
@@ -135,15 +153,21 @@ const Planner: FC = () => {
       return;
     }
 
+    // Support conversions from the old format
+    const loadedData =
+      'planners' in roadmap
+        ? roadmap.planners
+        : [{ name: "Peter's Roadmap", content: (roadmap as { planner: SavedPlannerYearData[] }).planner }];
```
js0mmer, line 160: same thing before about a constant for the name

site/src/store/slices/roadmapSlice.ts
```diff
@@ -307,10 +377,20 @@ export const {
...
   setShowAddCourse,
+  setRoadmapPlan,
+  addRoadmapPlan,
+  deleteRoadmapPlan,
+  setPlanIndex,
+  setPlanName,
   setUnsavedChanges,
 } = roadmapSlice.actions;
 
+// export const { setRoadmapPlan, addRoadmapPlan, deleteRoadmapPlan } = roadmapMultiplePlanSlice.actions;
```
js0mmer, line 388:
```suggestion
```

site/src/pages/RoadmapPage/Planner.tsx
```diff
@@ -61,15 +65,16 @@ const Planner: FC = () => {
     else {
       validatePlanner();
 
+      // TODO: idk if this works properly for multiplanner
```
js0mmer, line 68: I've found an edge case. Sometimes the unsaved prompt doesn't appear if you change the name of the planner. To reproduce: load up the page for the first time, changing the name of your current planner, attempt to leave or reload. No prompt will block you.

js0mmer (changes requested):
Looks mostly good. There was an edge case with the unsaved prompt that I found.

> Adding a third roadmap should not default to the same name as the second one if it's called "Roadmap 3"

This does not seem to be the case. If I have 2 roadmaps and the second is called "Roadmap 3" then attempt to add another, the new one is also called "Roadmap 3". Additionally, right now the user is allowed to give 2 planners the same name but I'm not sure if we want that. Having 2 planners with the same name throws a key error at the moment.

image

This can be fixed by appending the index of the planner into the key name but I'm not sure if we want to do that. For now, it's fine since the order is always the same but I think a good feature to add in the future would be the ability to rearrange the order of the planners and relying on the index for the key would not work in that scenario.

Awesome-E:
Can you check off the things in the PR description that _did_ work while testing? Would be helpful to see.

js0mmer:
> Can you check off the things in the PR description that _did_ work while testing? Would be helpful to see.

Yeah.

One more thing I forgot to mention, clicking on the multiplanner dropdown again should close it. Right now the only way to close it is to click outside of it somewhere.

site/src/pages/RoadmapPage/Planner.tsx
Awesome-E: Cannot reproduce

https://github.com/icssc/peterportal-client/assets/54484616/ea04ed99-6457-4164-bda4-52a3fbfbde0e

js0mmer: https://github.com/icssc/peterportal-client/assets/8922227/2bafe14f-bead-4200-a06a-63beea115055

Awesome-E:
> If I have 2 roadmaps and the second is called "Roadmap 3" then attempt to add another, the new one is also called "Roadmap 3".

Forgot that in the original PR, but it's there now

> One more thing I forgot to mention, clicking on the multiplanner dropdown again should close it. Right now the only way to close it is to click outside of it somewhere.

Addressed this

site/src/pages/RoadmapPage/RoadmapMultiplan.tsx
```diff
@@ -0,0 +1,257 @@
...
+              index={index}
+              clickHandler={() => {
+                dispatch(setPlanIndex(index));
+                setCurrentPlanIndex(index);
+              }}
+              editHandler={() => setEditIdx(index)}
+              deleteHandler={() => setDelIdx(index)}
+            />
+          ))}
+          {/* <Dropdown.Item onClick={() => }>Edit Plan Name</Dropdown.Item> */}
```
js0mmer, line 122:
```suggestion
```

js0mmer (changes requested):
image

One more thing with roadmap name input I noticed, you can provide an empty name which I don't think should be allowed.

It's weird that you can't reproduce the unsaved dialog issue.

Everything else looks good.

Awesome-E:
> It's weird that you can't reproduce the unsaved dialog issue.

do you have a video of this? Maybe it has something to do with local DB data being shared with staging?

js0mmer:
> > It's weird that you can't reproduce the unsaved dialog issue.
> 
> do you have a video of this? Maybe it has something to do with local DB data being shared with staging?

I replied with one in the original comment thread

Awesome-E:
I see; I was able to repro after clearing my own localStorage

Awesome-E:
It looks like the issue is with `allPlanData` not being in the useEffect deps array. If I add that though, I still get the warning that the other functions aren't in it and it causes a bunch of other issues with infinite looping

The code to load seems to need a larger refactoring

Awesome-E:
My internet is a bit weird where I am rn – let me know if this change fixes the issue @js0mmer

Awesome-E:
Not sure why it's trying to lint unchanged files again

site/src/pages/RoadmapPage/Planner.tsx
```diff
@@ -191,102 +85,37 @@ const Planner: FC = () => {
...
-        return new Set();
-      }
+      validatePlanner(transfers, currentPlanData, (missing, invalid) => {
+        // set missing courses
+        setMissingPrerequisites(missing);
+        // set the invalid courses
+        dispatch(setInvalidCourses(invalid));
+      });
+
+      // TODO: idk if this works properly for multiplanner
```
js0mmer, line 113:
```suggestion
```

js0mmer (approved):
LGTM

PR #473, opened 2024-05-14
Sync local roadmap with cloud

### LINKED ISSUE #313: Synchronize local roadmap storage with account storage
If the user starts modifying and saves their roadmap when they are not logged in and then log in, the roadmap saved to their account is loaded and they have no way of retrieving their locally saved roadmap. Consider adding a timestamp to saved roadmap and prompt the user asking if they want to load the local one if the local one is newer than the account one. Also consider adding a load button for manual loading the local schedule.

js0mmer:
## Description
* prompt the user to load their local roadmap if it is newer than the one saved to their account in mongo
* added a timestamp to the roadmap object for the sake of determining which is newer
* as of right now, after loading the local roadmap, you have to click save to save it to your account in the db
  * refreshing the page will bring up the prompt again
  * maybe could add a toast or some other message letting the user know to save if they want to keep these changes
* also, it takes a second to load after clicking yes, may need to add a loading indicator or something

## Screenshots

image

  1. Log out
  2. make changes to schedule and save
  3. log back in, prompt should now appear
  4. if you save, roadmap will be saved both to db and localstorage, prompt should no longer appear if you refresh now

## Final Checks:

(optional)

- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #313


site/src/helpers/planner.ts
```diff
@@ -137,12 +137,15 @@ export const loadRoadmap = async (
...
+
+  let isLocalNewer = false;
+
   if (!roadmap && localRoadmap) {
-    roadmap = JSON.parse(localRoadmap);
-  }
-  // no saved planner
-  if (!roadmap) {
+    roadmap = localRoadmap;
+  } else if (roadmap && localRoadmap && (localRoadmap.timestamp ?? 0) > (roadmap.timestamp ?? 0)) {
```
Awesome-E, line 145: I suppose this logic would work on prod instances, but it was confusing that it didn't work locally because a different branch didn't set the timestamp of the roadmap. I think this PR in itself is a pretty rare occurrence though, and for those cases this code would work

site/src/pages/RoadmapPage/Planner.tsx
```diff
@@ -112,14 +131,42 @@ const Planner: FC = () => {
...
+      <Modal
+        show={showSyncModal}
+        onHide={() => {
+          setShowSyncModal(false);
+        }}
+        className="ppc-modal"
+        centered
+      >
+        <Modal.Header closeButton>
+          <h2>Cloud roadmap out-of-date</h2>
```
Awesome-E, line 153: Capitalization is inconsistent with other modals

site/src/pages/RoadmapPage/Planner.tsx
```diff
@@ -112,14 +131,42 @@ const Planner: FC = () => {
...
+        </Modal.Header>
+        <Modal.Body>
+          <p>
+            Your roadmap saved locally is newer than the one saved to the cloud. Would you like to use the local one
```
Awesome-E, line 157: This is not very user-friendly text – I would recommend something more along the lines of:
> This device's saved roadmap is newer than the one synced to your account. Where do you want to load the roadmap from?

site/src/pages/RoadmapPage/Planner.tsx
```diff
@@ -112,14 +131,42 @@ const Planner: FC = () => {
...
+            instead?
+          </p>
+        </Modal.Body>
+        <Modal.Footer>
+          <Button variant="danger" onClick={handleLoadLocal}>
+            Yes, Use Local
+          </Button>
+          <Button variant="secondary" onClick={() => setShowSyncModal(false)}>
+            No, Use Cloud
```
Awesome-E, line 166: Consider which one you want to be the primary action. Both are potentially destructive, but in the case that this popup occurs, do you think it's more likely that the user "just got signed out and wants to use their local one"? (That's what I imagine, but not sure if you agree)

If you agree, I would make "Yes" -> "This Device" and change the color to the primary, and change "No, Use Cloud" -> "My Account"

Awesome-E (changes requested):
See changes regarding text in the modal; I think we should word it in a way that's more clear to people who don't understand that there's such thing as a "local" and "cloud" copy.

Awesome-E:
Other code changes seemed pretty minimal though so good on that front

site/src/pages/RoadmapPage/Planner.tsx
js0mmer: Yes, the scenario I wrote this issue for was when making changes to my roadmap locally without realizing I'm signed out and having no way to load them onto my account. The only other scenario where I imagine this modal appearing is if multiple people are logging in/out of the same device which I don't see happening often.

site/src/helpers/planner.ts
js0mmer: I'm not sure what you mean. It seemed to work on my initial tests locally and I nullish coalesced the timestamps with 0 to try to avoid issues with missing timestamps on old schedules.

Awesome-E: I think it was just an issue for some branches without the code, so you can disregard this thread.

Awesome-E (approved):

Awesome-E:
Looks like you need to re-trigger the GitGuardian Security Check

js0mmer:
GitGuardian appears to be broken so I removed it

PR #474, opened 2024-05-15
Clean up Roadmap Year/Quarter Styles

Awesome-E:
## Screenshots

Before:
image
After:
image

## Final Checks:

### DIFF
--- a/site/src/pages/RoadmapPage/Course.scss
+++ b/site/src/pages/RoadmapPage/Course.scss
@@ -4,23 +4,26 @@
   border-radius: 15px;
   height: auto;
   font-size: 16px;
-  padding: 10px 20px;
+  padding: 12px;
   position: relative;
+
+  .course-and-info {
+    color: var(--zot-blue);
+    a {
+      color: inherit;
+    }
+  }
 
   .name {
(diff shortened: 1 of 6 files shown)


js0mmer (commented):
Haven't looked at any code yet but visually it looks pretty good!

A couple things that stood out to me were:
1. The context menu for quarters gets chopped off when all the quarters in that row are empty
image
3. When a user first opens the roadmap, I don't think they'll know how to interact with the roadmap when they just see 3 rectangles with quarter names. This ties back to your point about the necessity of an onboarding tutorial but I also think that something as small as adding a message into the empty quarter saying "No courses added yet! Drag a course from the search sidebar on the right to start building your planner." could make it much easier for new users to understand what they're looking at and how to interact with it. Thoughts?
image

Awesome-E:
image
This commit should address both issues

js0mmer (changes requested):
One last thing I just noticed. The add course buttons are missing on mobile

image

Awesome-E:
This should be fixed now

js0mmer (approved):
LGTM

PR #478, opened 2024-05-20
Can't use backend without mongo url env secret

### LINKED ISSUE #446: Can't use backend without mongo url env secret
Backend will crash without the mongo url. Something must've changed when updating packages. We need to fix this so open source contributors can develop without access to our mongo (they should be able to use anything that doesn't require the DB, right now the entire express server crashes).

Blocked by #414

ptruong0:
## Description

Check for mongo env variable in app.ts. Also added some error handling on the backend to handle failed requests to roadmap, reviews, etc routes

Search, course, and professor pages should work as usual. Roadmaps can be saved locally but not to the user account.

## Final Checks:

(optional)

- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #446


api/src/app.ts
```diff
@@ -35,35 +35,36 @@ const app = express();
...
+    secret: process.env.SESSION_SECRET!,
+    resave: false,
+    saveUninitialized: false,
+    cookie: { maxAge: SESSION_LENGTH },
+    store: store,
+  }),
+);
+app.use(passport.initialize());
+app.use(passport.session());
+require('./config/passport');
```
js0mmer, line 67:
```suggestion
// Setup Passport and Sessions
if (!process.env.SESSION_SECRET) {
  console.log('SESSION_SECRET env var is not defined!');
}
app.use(
  session({
    secret: process.env.SESSION_SECRET ?? 'secret',
    resave: false,
    saveUninitialized: false,
    cookie: { maxAge: SESSION_LENGTH },
    store: store,
  }),
);

if (process.env.GOOGLE_CLIENT && process.env.GOOGLE_SECRET) {
  app.use(passport.initialize());
  app.use(passport.session());
  require('./config/passport');
} else {
  console.log('GOOGLE_CLIENT and/or GOOGLE_SECRET env var(s) not defined! Google login will not be available.');
}
```

js0mmer (commented):
I'm getting this error when running locally with the example env file. Doesn't seem related to Mongo but it still prevents users from running without secrets.

image

I've suggested some conditions that should fix this

js0mmer (approved):

PR #481, opened 2024-05-27
Conditonally hide transfer missing prereqs

timobraz:
## Description
Simple change to hide the text saying missing prereqs when there aren't any prereqs in the set.

## Screenshots
No missing:
image
With missing:
image

## Steps to verify/test this change:

- [ ] Verify changes work as expected on staging instance

## Final Checks:

- [ ] Verify successful deployment

(optional)

- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #448

PR #482, opened 2024-05-29
Featured review criteria changes

### LINKED ISSUE #355: Update featured review criteria
Currently, it seems to grab the latest review for the professor. It some cases, this review may have no content.

Example:
image

We should update criteria so that it will always feature a review with content. In the case of no review that meets that criteria, we can display an N/A message. We might also want to show reviews with the highest score too instead of the latest or have some other sort of algorithm/criteria for determining what should be featured.

timobraz:
## Description
Changed the sorting algorithm for retrieving reviews to follow this priority order:

1. Review Content Length
2. Score
3. Verified Status

On the frontend, when no review exists with content, it will display the rating of the best matching review

## Screenshots
When some review has content, with same score and unverified
image
When no review with content, with same score and unverified
image

## Final Checks:

(optional)

- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #355

### DIFF
--- a/api/src/controllers/reviews.ts
+++ b/api/src/controllers/reviews.ts
@@ -64,7 +64,7 @@ router.get('/featured', async function (req: Request<never, unknown, never, Feat
 
   // find first review with the highest score
   const reviewsCollection = await Review.find({ [field]: req.query.id })
-    .sort({ score: -1 })
+    .sort({ reviewContent: -1, score: -1, verified: -1 })
     .limit(1);
   if (reviewsCollection) {
     res.json(reviewsCollection);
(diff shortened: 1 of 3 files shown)


site/src/pages/RoadmapPage/Transfer.tsx
```diff
@@ -94,8 +93,12 @@ const Transfer: FC<MissingCoursesProps> = ({ missingPrereqNames }) => {
...
           must be entered as "AP COMP SCI A")
         </p>
-        <p>Missing Prerequisites</p>
-        <DisplayMissingCourses />
+        {missingPrereqNames?.size > 0 && (
+          <>
+            <p>Missing Prerequisites</p>
+            <DisplayMissingCourses />
+          </>
+        )}
```
js0mmer, line 101: seems to be from #481 
if you merge that one first, this should be fine though

api/src/controllers/reviews.ts
```diff
@@ -64,7 +64,7 @@ router.get('/featured', async function (req: Request<never, unknown, never, Feat
 
   // find first review with the highest score
```
js0mmer, line 65: comment can be updated to state new criteria

js0mmer (approved):

PR #484, opened 2024-08-29
Remove empty quarter because we use grids now

Awesome-E:
## Description
Remove the empty-quarter elements because they just showed up as empty boxes of different size in our grid.

## Screenshots

**Before:**
image

**After:**
image

## Final Checks:


js0mmer (approved):
LGTM

PR #495, opened 2024-10-30
Adjust appearance of course hit

Awesome-E:
## Description

Tweaks the appearance of course hits so the titles look nicer and the tags follow our site's theming

This also fixes the issue with the last PR where light mode would cause the text inside professor name icons to be black

I'm not super satisfied with the end result, but I do think it is a step up. I think the problem is there's still a mental load when trying to differentiate between the course title and the department. We could experiment with changing the color of the subtitle at some point.

Another consideration could be to remove the subtitle entirely and just show that in the full course page, since which school a class is in is usually the same as the course title. (I know cross-listings are where this might be useful, but to me that sounds like something you'd want to know only if you're interested in it, and therefore click into the full page)

## Screenshots

Before:
image

After:
image

## Final Checks:


js0mmer (approved):
🚀

PR #496, opened 2024-10-30
Use ToastifyJs to replace alerts

timobraz:
## Description
All alerts are replaced with a toast popup in bottom left with the signature peterportal color.

## Screenshots
image

## Final Checks:

(optional)

- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #


js0mmer (commented):
Could you update the of the hook function/calls to be consistent? For a hook, `useToastify` would be a preferred name over `toaster`.

Also, I think having the error toasts in red would be better, blue for only the info toasts. The hook could take an argument where we specify the type of toast as 'info' or 'error' and the color is adjusted accordingly. Thoughts @Awesome-E?

timobraz:
I did that originally and the linter complains because it's used in cotext files.

timobraz:
I think I'll instead just move it out of the hooks folder and into lib.

js0mmer:
> I think I'll instead just move it out of the hooks folder and into lib.

Ok, that sounds fine. I guess it's not really a hook? You can move it out and just call it toaster.

Awesome-E:
> Also, I think having the error toasts in red would be better, blue for only the info toasts. The hook could take an argument where we specify the type of toast as 'info' or 'error' and the color is adjusted accordingly. Thoughts @Awesome-E?

I think that makes sense yeah.


> I guess it's not really a hook?

Yeah it should be treated more like a helper function, not really a hook

js0mmer (approved):

PR #497, opened 2024-11-01
Migrate to Postgres

### LINKED ISSUE #490: Migrate to Postgres
postgres + drizzle

js0mmer:
## Description

* Migrated to postgres
* added drizzle
* refactored routes
  * made sure we are determining userid from session, not letting user dictate it
* refactored login checks on the frontend with a hook
* added course bag routes
* added separate course bag slice

## todo
- [ ] need to clean up mongoose stuff before merging
- [x] update docs

- [ ] make sure the site still works

## Final Checks:

(optional)

- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #490


api/src/controllers/reviews.ts
```diff
@@ -3,185 +3,162 @@
...
+    .from(review)
+    .where(
+      and(
+        ...[
+          ...(courseId ? [eq(review.courseId, courseId)] : []),
+          ...(professorId ? [eq(review.professorId, professorId)] : []),
+          ...(userId ? [eq(review.userId, userId)] : []),
+          ...(reviewId ? [eq(review.id, reviewId)] : []),
+          ...(verified !== undefined ? [eq(review.verified, verified)] : []),
+        ],
```
Awesome-E, line 68: This seems a bit clunky. At the very least we don't have to construct an array just to spread it. If anything, I'd prefer to declare `const filters = []` and simply add each `eq()` to the filters array, then just do `.and(...filters)` at the end.

```suggestion
        ...(courseId ? [eq(review.courseId, courseId)] : []),
        ...(professorId ? [eq(review.professorId, professorId)] : []),
        ...(userId ? [eq(review.userId, userId)] : []),
        ...(reviewId ? [eq(review.id, reviewId)] : []),
        ...(verified !== undefined ? [eq(review.verified, verified)] : []),
```

api/src/controllers/reviews.ts
```diff
@@ -3,185 +3,162 @@
...
-    const query: ReviewFilter = {
-      courseID: input.courseID,
-      professorID: input.professorID,
-      userID: ctx.session.passport?.user.id,
-    };
-
-    const reviews = await Review.find(query);
-    if (reviews?.length > 0)
+    //check if review already exists for same professor, course, and user (do this before verifying captcha)
+    const existingReview = await db
```
Awesome-E, line 130: Ideally this check happens before they even open the form – let's create an issue for that

api/src/controllers/reviews.ts
```diff
@@ -191,77 +168,62 @@ const reviewsRouter = router({
...
+      .as('vote_query');
+
+    const field = input.type === 'course' ? review.courseId : review?.professorId;
+    const featuredReview = (
+      await db
+        .select()
+        .from(review)
+        .where(eq(field, input.id))
+        .leftJoin(voteSubQuery, eq(voteSubQuery.reviewId, review.id))
+        .orderBy(desc(review.content), desc(voteSubQuery.score), desc(review.verified)) // featured review criteria
```
Awesome-E, line 217: Put as a variable instead of as a comment

`const featuredReviewCriteria = [desc(review.content), desc(voteSubQuery.score), desc(review.verified)`

`.orderBy(...featuredReviewCriteria)`

api/src/controllers/reviews.ts
```diff
@@ -191,77 +168,62 @@ const reviewsRouter = router({
...
+        .limit(1)
+    )[0];
+
+    return {
+      ...featuredReview.review,
+      createdAt: featuredReview.review.createdAt.toISOString(),
+      updatedAt: featuredReview.review.updatedAt?.toISOString(),
```
Awesome-E, line 224: It looks like converting dates is repeated behavior. We should have some sort of helper function to automatically convert all date key/value pairs into ISO strings

api/src/controllers/roadmap.ts
```diff
@@ -1,27 +1,76 @@
...
-      // add roadmap to mongo
-      return await new Roadmap({ roadmap, userID, coursebag }).save();
-    }
+  save: userProcedure.input(savedRoadmap).mutation(async ({ input, ctx }) => {
+    const { planners, transfers, timestamp } = input;
+    const userId = ctx.session.userId!;
+
+    const plannerUpdates = planners
+      .filter((planner) => planner.id !== undefined)
+      .map((planner_) =>
```
Awesome-E, line 40: if `planner` is an import that you don't want to override, I'd rather have this variable name be `plannerData`

api/src/controllers/users.ts
```diff
@@ -2,57 +2,33 @@
...
-  getPreferences: userProcedure.query(async ({ ctx }): Promise<UserPreferences> => {
-    const userID = ctx.session.passport?.user.id;
-    const preference = await Preference.findOne({ userID: userID });
-
-    return (preference ? preference : {}) as UserPreferences;
+  get: userProcedure.query(async ({ ctx }) => {
+    const userData = (await db.select().from(user).where(eq(user.id, ctx.session.userId!)))[0];
+    return {
+      ...userData,
+      lastRoadmapEditAt: userData.lastRoadmapEditAt?.toISOString(),
```
Awesome-E, line 20: same as previous occurrences of this pattern

api/src/db/migrate.ts
```diff
@@ -0,0 +1,259 @@
...
+  db.delete(review),
+  db.delete(user),
+]);
+
+// get all possible users from sessions, then reviews, then roadmaps, then votes
+const sessions = await Session.find();
+// need to filter out bad sessions (never signed in with google, just clicked log in)
+const sessionsUserRecords = sessions
+  .filter((session) => session.session?.passport?.user?.id != null)
+  .map((session) => ({
```
Awesome-E, line 36: I know it's nice to be able to directly return the object, but at the same time repeating `sessions.session!.passport!.user!` is a bit clunky

If you _really_ want to, you could also do `.map(({ session: { session: { passport: { user } } }) => ({...})` but that's equally as confusing

Awesome-E: I think it'd be best to just do something like this:

```ts
.map(session => {
  const { id, name, email, picture } = session.session!.passport!.user!
  return { googleId: id, name, email, picture }
})
```

api/src/db/migrate.ts
```diff
@@ -0,0 +1,259 @@
...
+const roadmaps = await Roadmap.find<Omit<MongoRoadmap, 'userId'> & { userID: string }>();
+for (const roadmap of roadmaps) {
+  const newId = await db
+    .insert(user)
+    .values({
+      googleId: roadmap.userID,
+      name: anonymousName,
+      email: '',
+      picture: '',
+      ...(roadmap.roadmap.timestamp ? { lastRoadmapEditAt: new Date(Number(roadmap.roadmap.timestamp)) } : {}),
```
Awesome-E, line 75: similar to dates the other way, we should have a helper to construct this pattern
`dateLookup('lastRoadmapEditAt', roadmap.roadmap.timestamp)`

api/src/db/schema.ts
```diff
@@ -0,0 +1,141 @@
...
+    reviewId: integer()
+      .notNull()
+      .references(() => review.id, { onDelete: 'cascade' }),
+    userId: integer()
+      .notNull()
+      .references(() => user.id),
+    vote: integer().notNull(),
+  },
+  (table) => ({
+    voteCheck: check('votes_vote_check', sql`${table.vote} = 1 OR ${table.vote} = -1`),
```
Awesome-E, line 118: is there a reason for using `sql'...'` here instead of something like `or()`?

api/src/models/review.ts
```diff
@@ -67,6 +67,10 @@ const reviewSchema = new mongoose.Schema({
     type: Boolean,
     required: false,
   },
+  textbook: {
+    type: Boolean,
+    required: false,
+  },
```
Awesome-E, line 73: why is this added in this PR?

api/src/models/session.ts
Awesome-E, line 1: similar to previous file – why are the mongoose schemas still here?

site/src/component/SideInfo/SideInfo.tsx
Awesome-E, line 1: Lint warning on changed file – I'm indifferent about changing it here

site/src/helpers/planner.ts
```diff
@@ -116,40 +92,31 @@ export const expandAllPlanners = async (plans: SavedPlannerData[]): Promise<Road
   return await Promise.all(
     plans.map(async (p) => {
       const content = await expandPlanner(p.content);
-      return { name: p.name, content: { yearPlans: content, invalidCourses: [] } };
+      return { ...(p.id ? { id: p.id } : {}), name: p.name, content: { yearPlans: content, invalidCourses: [] } };
```
Awesome-E, line 95: I don't like this type of pattern, but not sure if there's a better way

site/src/pages/RoadmapPage/ImportTranscriptPopup.tsx
```diff
@@ -7,7 +7,7 @@ import { useAppDispatch } from '../../store/hooks';
 import { parse as parseHTML, HTMLElement } from 'node-html-parser';
 import ThemeContext from '../../style/theme-context';
 import { BatchCourseData, PlannerQuarterData, PlannerYearData } from '../../types/types';
-import { normalizeQuarterName, quarterNames } from '../../helpers/planner';
+import { quarters as quarterNames } from '@peterportal/types';
```
Awesome-E, line 10: should we consider renaming this to `quarterNames` in `@peterportal/types` altogether since it's just names and not JSON?

site/src/pages/RoadmapPage/Transfer.tsx
```diff
@@ -55,7 +55,7 @@ const TransferEntry: FC<TransferEntryProps> = (props) => {
         <Form.Control
           type="number"
           placeholder="Units"
-          value={units}
+          value={units ?? undefined}
```
Awesome-E, line 58: I assume this is because `null` was giving an error?

Awesome-E (changes requested):

site/src/pages/RoadmapPage/Transfer.tsx
js0mmer: yes, database returns a real number or null for this but the value attribute can only be number or undefined

api/src/db/migrate.ts
js0mmer: i don't really care about the code in this file because i'm not merging it, this just for a one-time use of migrating everything over

api/src/models/session.ts
js0mmer: i'll delete them before merging, I needed them for the migration script

site/src/component/SideInfo/SideInfo.tsx
js0mmer: i might make a pr in the future addressing all the remaining warnings

api/src/controllers/reviews.ts
js0mmer: Ah, so you mean like doing the check to prevent them from even having the option on the form to create an identical review?

api/src/controllers/reviews.ts
js0mmer: Yes, I'll remove the outer array. I'm really not sure if adding will look any cleaner though since we would just be adding based on these same conditions so it would look kind of the same. I took the spreading idea from [drizzle docs](https://orm.drizzle.team/docs/select#conditional-select).

api/src/controllers/reviews.ts
js0mmer: On it

site/src/pages/RoadmapPage/ImportTranscriptPopup.tsx
js0mmer: Yeah, the type actually comes from peterportal-api-next-types though which is exported by `@peterportal/types`

api/src/db/schema.ts
js0mmer: I think `or()` works here? I have to use a non-null assertion for some reason but I think it's working.
```ts
    voteCheck: check('votes_vote_check', or(eq(table.vote, 1), eq(table.vote, -1))!),
```

api/src/controllers/reviews.ts
Awesome-E: I think there is a way to make the helper type-safe. I'm not used to writing complicated type-defs like this, but I think that's a good use for ChatGPT.

api/src/db/schema.ts
Awesome-E: Great, I think that's better for consistency

site/src/pages/RoadmapPage/ImportTranscriptPopup.tsx
Awesome-E: I see. In that case, we might want to rename the export

api/src/controllers/reviews.ts
js0mmer: Cooked something up with some help from github copilot

site/src/pages/RoadmapPage/ImportTranscriptPopup.tsx
js0mmer: Added an export for `quarterNames`

api/src/controllers/reviews.ts
js0mmer: Ok I found another reference on the [drizzle docs](https://orm.drizzle.team/docs/guides/conditional-filters-in-query) and I was able to remove the spread operators.

Awesome-E (approved):
Just create the issues for the two unresolved comments, otherwise looks pretyt much good to me! Wouldn't hurt to get a second opinion from @timobraz

.github/workflows/build-and-deploy.yml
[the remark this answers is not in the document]
js0mmer: #499

api/src/controllers/reviews.ts
js0mmer: #500

api/src/db/schema.ts
js0mmer: for future reference, `or()` does not work here. we have to use `sql` for checks

PR #530, opened 2024-12-07
Fix search page crash

### LINKED ISSUE #529: Page crash when switching between course/professor popup results
Search up any query, click on one result, then click on another and the page will crash

js0mmer:
## Description

* Updated conditions in GradeDist to make sure gradeDistData is not undefined
* Removed use of non-null assertions in state initializations for better type safety

## Final Checks:

(optional)

- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #529

### DIFF
--- a/site/src/component/GradeDist/GradeDist.tsx
+++ b/site/src/component/GradeDist/GradeDist.tsx
@@ -104,7 +104,7 @@ const GradeDist: FC<GradeDistProps> = (props) => {
 
   // update list of professors/courses when new course/professor is detected
   useEffect(() => {
-    if (gradeDistData && gradeDistData.length !== 0) {
+    if (gradeDistData?.length) {
       if (props.course) {
         createProfEntries();
       } else if (props.professor) {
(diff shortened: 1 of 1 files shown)


site/src/component/GradeDist/GradeDist.tsx
```diff
@@ -156,7 +156,7 @@ const GradeDist: FC<GradeDistProps> = (props) => {
 
   // update list of quarters when new professor/course is chosen
   useEffect(() => {
-    if ((currentProf || currentCourse) && gradeDistData.length !== 0) {
+    if ((currentProf || currentCourse) && gradeDistData && gradeDistData.length !== 0) {
```
Awesome-E, line 159: If you want it to be more concise, you can do:
```suggestion
    if ((currentProf || currentCourse) && gradeDistData?.length) {
```

That's readable to me, but I may be used to it more than others. Your call.

Awesome-E (approved):

Awesome-E:
LGTM!

PR #532, opened 2024-12-07
added button to duplicate roadmap

### LINKED ISSUE #519: Add the ability to duplicate a roadmap
details to be filled out later, but essentially having this would be nice in the case that people have a plan for 3 years persay and the rest of the roadmap depends whether classes will be available.

timobraz:
New button to duplicate currently selected roadmap.
Will copy the content of the selected roadmap and select the copy upon duplication.
### Before:
image
### After:
image

- [ ] Check if all the courses are transferred over
- [ ] Check if all transfer courses are transferred over
- [ ] Check if copy number is incremented when you have two copies 
- [ ] Check that if you save and reload all your roadmaps are there

## Final Checks:

(optional)

- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #519


Awesome-E (commented):
Changes seem to work correctly, and code looks good! A couple things to note though:
- Transfers are separate from the roadmap so they're not really "copied" but rather it's just reading the same one. It does still work though
- I think it'd make more sense to have the duplicate button between the pencil and the trash can icons, because right now it's not clear (even with text) that "duplicate roadmap" refers to the current one (it may be confused as a button to open a selection menu). I don't know how technically difficult it would be to make it so that every roadmap can be duplicated easily, but I think it's worth exploring

charlieweinberger (commented):
LGTM. However, I agree with Ethan's 2nd point, and think if possible it should be implemented before merging with main. See AntAlmanac as an example.

timobraz:
image
Addressed the changes to look more like Ant Almanac and be more explicit about which roadmap you're duplicating.

site/src/pages/RoadmapPage/RoadmapMultiplan.tsx
```diff
@@ -89,18 +94,12 @@ const RoadmapMultiplan: FC = () => {
...
   };
 
-  const duplicateCurrentPlan = () => {
-    const currentPlan = allPlans.plans[currentPlanIndex];
-    let newName = `${currentPlan.name} (Copy)`;
-    let counter = 1;
-    while (allPlans.plans.find((p) => p.name === newName)) {
-      counter++;
-      newName = `${currentPlan.name} (Copy ${counter})`;
-    }
```
Awesome-E: Copying does not seem to increment the numbers correctly after this change. I assume it's because of this.
image

Awesome-E (changes requested):

Awesome-E:
Is it intended that duplicating "test (copy)" creates "test (copy) (copy)" instead of "test (copy 2)"?

Awesome-E (approved):

PR #533, opened 2024-12-07
Change tick size for charts

ethank2222:
## Description

Always 6 ticks, each tick size is calculated as the max table value divided by 6.

## Screenshots
Begin:
image

End:
Screenshot 2024-12-06 191455

Look at chart for class availabilities

## Final Checks:

Yes, I did

## Issues

None

Closes #


api/.env.example
Awesome-E, line 1: This is an _example_ `.env` file that shows contributors the names of variables that should be used. It should not be deleted from the repository

package.json
Awesome-E, line 1: It looks like you installed `pnpm` locally (in this project) instead of globally. `pnpm` is used as a global package manager, so it should be installed with `npm i -g pnpm` rather than within the project. You should revert the changes to this file in the PR

package-lock.json
Awesome-E, line 1: Same comment with `package.json`

pnpm-lock.yaml
Awesome-E, line 1: Same comment as `package.json`

site/src/component/QuarterTooltip/Chart.tsx
```diff
@@ -153,7 +156,7 @@ export default class Chart extends React.Component<ChartProps> {
                 legendOffset: 36,
               }}
               axisLeft={{
-                tickValues: Array.from({ length: greatestCount }, (_, i) => i + 1), // integers from 1 to max
+                tickValues: Array.from({ length: greatestCount / tickSize }, (_, i) => i * tickSize + tickSize), // integers from 1 to max
```
Awesome-E, line 159: Change or remove the out-of-date comment
```suggestion
                tickValues: Array.from({ length: greatestCount / tickSize }, (_, i) => i * tickSize + tickSize),
```

Awesome-E (changes requested):
Double check some of the repo files – it looks like this PR should just be changing the `QuarterTooltip` component, and not any dependencies

Awesome-E (approved):

PR #535, opened 2024-12-07
Tooltip and chart styling fixes

ellisl10:
Tooltip and chart styling fixes

## Description

Moved color constants and tooltip dimensions/styling to separate files and made dimensions the same.

## Screenshots
Before
Screenshot 2024-12-09 at 8 04 34 PM

After
Screenshot 2025-01-01 at 6 17 56 PM

## Final Checks:

(optional)

- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #


api/.env.example
Awesome-E, line 1: This is an example file for the format of environment variables, not env variables themselves. It should not be deleted

site/src/component/GradeDist/Pie.tsx
```diff
@@ -165,24 +177,11 @@ export default class Pie extends React.Component<PieProps> {
           enableArcLinkLabels={false}
           innerRadius={0.8}
           padAngle={2}
-          colors={['#60A3D1', '#81C284', '#F5D77F', '#ECAD6D', '#E8966D', '#4AB486', '#E36436']}
```
Awesome-E, line 168: It looks like the last two removed colors here are actually different from the other list. I think we should use the green and red for pass + no pass instead of making them gray

site/src/component/GradeDist/Pie.tsx
```diff
@@ -102,7 +104,7 @@ export default class Pie extends React.Component<PieProps> {
         id: 'A',
         label: 'A',
         value: gradeACount,
-        color: '#60A3D1',
+        color: '#54B058',
```
Awesome-E, line 107: consider using something like `colors[0]` for this so we can make these use the same source for colors

site/src/component/GradeDist/Colors.tsx
Awesome-E, line 1: Generally, files without components should start with lowercase and would end in `.ts` instead of `.tsx`, i.e. `chartColors.ts`

Since `Colors` is not a component, I suggest renaming this file to `colors.ts` with an `export const colors = [...]`

site/src/component/GradeDist/TooltipStyle.tsx
```diff
@@ -0,0 +1,17 @@
+import { type Theme } from '@nivo/core';
+
+export const TooltipStyle: Theme = {
```
Awesome-E, line 3: same comment as `Colors`

Awesome-E (changes requested):

site/src/component/GradeDist/Chart.tsx
```diff
@@ -70,61 +70,46 @@ export default class Chart extends React.Component<ChartProps> {
         id: 'A',
         label: 'A',
         A: gradeACount,
-        color: '#2484C6',
```
Awesome-E, line 73: Missing `colors[0]`

Although ideally, we use a typescript enum to name the colors instead of just putting them in a list.

```ts
export enum GradeColors {
  A = '#2484C6',
  B = '#54B058',
  // ...
}
// or Blue = '#2484C6', Green = '#54B058', ...
```

```ts
color: GradeColors.A // or color: GradeColors.Blue
```

site/src/component/GradeDist/colors.ts
Awesome-E, line 1: same comment as previous about considering an enum instead of a hard-coded list

site/src/component/PrereqTree/PrereqTree.tsx
```diff
@@ -72,7 +72,7 @@ const PrereqTreeNode: FC<TreeProps> = (props) => {
           }
           content={
             props.prerequisiteNames[
-              prereq.prereqType === 'course' ? prereq.courseId.replace(/ /g, '') : (prereq.examName ?? '')
+              prereq.prereqType === 'course' ? prereq.courseId.replace(/ /g, '') : prereq.examName ?? ''
```
Awesome-E, line 75: while this is correct, the precedence of `? :` over `??` is not obvious to everyone, so I think this is a change that should be reverted

(unless that was a prettier auto-format)

site/src/component/SideInfo/SideInfo.tsx
Awesome-E, line 1: Same as previous comment

api/.env.example
Awesome-E, line 1: Looks like there's still a diff here since the newline at the end was deleted. We should add it back

Awesome-E:
Just ran `pnpm lint` to find the problematic lint files, then `npx prettier **/*.tsx --write` to fix them, then committed them. Should be good to go!

PR #536, opened 2024-12-07
Restyle transfer credits

### LINKED ISSUE #517: Restyle transfer credits modal to conform to rest of modals
Current modal
image

Make use of `ppc-modal` css class and do some adjustments to style like the other modals
example (edit year modal):
image

CadenLee2:
Restyle the Transfer Credits modal (under Roadmap page)

## Description

- Make the modal conform to `ppc-modal` (ex. the `edit year` modal)
- Make the font size for `p` smaller when inside a `transfer-modal`
- Make the units column smaller
- Remove the unnecessary Close button (the X still exists in the corner)
- Move and restyle the Add Entry button, which does not match the buttons on the other modals because it serves a different purpose (working with the form rather than confirming changes)
- Locally test functionality still works (continues to count units when courses are added with units, allow adding/removing entries, closing the modal, changing the unit counts with up/down arrows) (tested on Firefox, should work elsewhere)
- Locally test appearance on light/dark mode, on phone, on small screen, and with many transfer courses added (inner scroll bar)

## Final Checks:

(optional)

- [ ] Write tests
- [ ] Write documentation

## Issues

Closes #517


Awesome-E:
If you want to style that button specifically a little bit more, you can do this so that the plus is aligned and the button is slightly more pronounced:
```css
  display: flex;
  align-items: center;
  gap: 4px;
  font-weight: bold;
```

Before (left) and after (right):

Awesome-E (commented):
Code changes look good, just a couple things
- See above comment about styling
- You opened the PR but didn't set two reviewers. Make sure to set me as one reviewer and one other individual contributor as a second reviewer.

Awesome-E:
Looks a bit better now, but just one more thing to check – is the input font size at least 16px? When I tap the field on iOS, it zooms the entire page, which usually means that the font size is less than that size

CadenLee2:
> Looks a bit better now, but just one more thing to check – is the input font size at least 16px? When I tap the field on iOS, it zooms the entire page, which usually means that the font size is less than that size

The input font size seems to be 14px. However, it's the same with the original Transfer Credits modal, as well as Edit Roadmap, New Roadmap, Add Year, etc. According to the inspector in Firefox and Chrome, all these input boxes seem to have `font-size: 1rem` due to the `form-control` css class from Bootstrap (`Form.Control`), which sets them to the root font size of 14px.

I tested the inputs of existing modals on iOS 18 (with appetize.io), and even the main course catalog search bar, and they all seem to have this zoom behavior. I think this makes sense and is expected, to let users see the input more clearly? But if it's a problem, maybe we should make a new issue for it, since it affects much more than just the Transfer Credits restyle?

site/src/pages/RoadmapPage/Transfer.scss
```diff
@@ -16,11 +17,45 @@
...
 }
 
 .entry {
   margin-top: 1%;
 }
+
+.add-entry {
+  background-color: transparent;
+  color: var(--peterportal-primary-color-1);
+  font-size: 17px;
```
Awesome-E, line 32: Maybe change to 16px or 18px, whatever is consistent with other modals

Awesome-E: After that, feel free to marge this

Awesome-E (approved):

site/src/pages/RoadmapPage/Transfer.scss
CadenLee2: Alright, done

PR #539, opened 2024-12-09
Remove axios

js0mmer:
## Description

* We use `fetch` in all but one place for making requests on the backend
* Converted that from `axios` to `fetch` to standardize things and remove a dependency

## Test Plan

- [ ] Verify reCAPTCHA for reviews still functions appropriately and succeeds when checked, fails if not checked


Awesome-E:
~Now we just have to do this on the frontend~
Will review this later when the deployment is up

js0mmer:
Oh interesting, it is installed on the frontend despite not being used there.

Awesome-E (approved):

PR #545, opened 2025-01-03
Fix rating review overflow

### LINKED ISSUE #304: In user Review page, there is an overflow the "Quality" & "Difficulty" heading
Gif display:

NOTE: The review tab on the right is not yet to merged, check out #270 to fix it.

charlieweinberger:
Fixed the overflow of the "quality"/"difficulty" headers in ratings for professor/course reviews.

## Description

Wherever reviews can be, the rating headers (and actually the rating values and the colored box itself) were not properly resizing when the browser window changed size. I fixed this by adding a media query for between widths of 1300px and 600px (there were no issues for widths outside of that range), and adjusting the widths and heights of the rating box, rating label, and rating value. I also fixed a bug where the rating boxes weren't the correct sizes at exactly 600px wide, by updating the smaller media query to start at 599px instead of 600px.

I definitely applied some of my personal opinions to this fix, since there isn't one clear way to make this section look the best. Let me know what you guys think about my changes, and if I should make the elements react different when the window changes dimensions.

## Screenshots

Before:

Screenshot 2025-01-02 at 4 44 24 PM

After:

Screenshot 2025-01-02 at 4 44 17 PM

## Test Plan

Create a review, then change the dimensions of the browser window, observing the ratings at various dimensions and ensuring that the rating header both stays inside the box the entire time, and looks good in terms of spacing & sizing for all window sizes. Make sure to check different widths, different heights, and both. Focus on widths between 1300px and 600px. 

## Issues

Closes #304

### DIFF
--- a/site/src/component/Review/Review.scss
+++ b/site/src/component/Review/Review.scss
@@ -26,23 +26,47 @@
   justify-self: flex-end;
   margin-right: 3vw;
 }
+.rating {
+  // TODO:
+  margin: 1vh 0;
+  border-radius: var(--border-radius);
+  width: 5vw;
+  height: 5vw;
+  min-width: 7rem;
+  min-height: 7rem;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  gap: 0.5rem;
+  color: white;
+}
 .rating-label {
   line-height: 1rem;
   font-size: 1rem;
   font-weight: bold;
 }
-.rating {
-  margin: 1vh 0;
-  border-radius: var(--border-radius);
-  width: 5vw;
-  height: 5vw;
-  display: flex;
-  flex-direction: column;
-  align-items: center;
-  justify-content: center;
+.rating-value {
   line-height: 3rem;
   font-size: 3rem;
-  color: white;
+}
+// screen width is between 1300px and 600px, inclusive
+@media only screen and (max-width: 1300px) and (min-width: 600px) {
+  .rating {
+    width: 8vw; // The rating box width & height shrink...
+    height: 8vw;
+    min-width: 5rem; // ...but only by so much.
+    min-height: 5rem;
+    gap: max(0.5vw, 0.4rem); // The same idea goes for gap...
+  }
+  .rating-label {
+    line-height: max(1.2vw, 0.8rem); // ...and rating label height/size.
+    font-size: max(1.2vw, 0.8rem);
+  }
+  .rating-value {
+    line-height: 4vw; // The rating value looks fine w/out a min value.
+    font-size: 4vw;
+  }
 }
 // color scheme for numerical ratings
 // reversed for difficulty so 1 => r5, etc.
@@ -136,7 +160,6 @@
   display: flex;
   flex-direction: column;
 }
-
 .add-report-button {
   background-color: transparent;
   border-color: transparent;
@@ -146,7 +169,6 @@
   transition: 0.2s;
   padding: 3px;
 }
-
 .add-report-button:hover {
   background-color: lightgray;
 }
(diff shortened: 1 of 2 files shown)


site/src/component/Review/Review.scss
```diff
@@ -26,22 +26,46 @@
...
 }
-.rating-label {
-  line-height: 1rem;
-  font-size: 1rem;
-}
 .rating {
   margin: 1vh 0;
   border-radius: var(--border-radius);
   width: 5vw;
   height: 5vw;
```
Awesome-E, line 33: Change this to use a fixed width instead of `vw`

site/src/component/Review/Review.scss
```diff
@@ -26,22 +26,46 @@
...
+  min-width: 7rem;
+  min-height: 7rem;
```
Awesome-E, line 35: If you have a fixed height, you can remove this

site/src/component/Review/Review.scss
Awesome-E: Having a fixed height also makes it so you only have to check a couple different screen sizes (i.e. phone, desktop) instead of checking the whole continuous range

site/src/component/Review/Review.scss
```diff
@@ -26,22 +26,46 @@
...
+}
+// screen width is between 1300px and 600px, inclusive
+@media only screen and (max-width: 1300px) and (min-width: 600px) {
+  .rating {
+    width: 8vw; // The rating box width & height shrink...
+    height: 8vw;
+    min-width: 5rem; // ...but only by so much.
+    min-height: 5rem;
+    gap: max(0.5vw, 0.4rem); // The same idea goes for gap...
+  }
```
Awesome-E, line 60: This can be a lot less confusing if we have a fixed width for the rating tiles. It's common to think that things should scale with screen size (and that was the assumption before this PR), but most of the time that's not really true.

site/src/component/Review/Review.scss
```diff
@@ -217,10 +208,33 @@
...
-  gap: 10px;
+  .subreview-ratings {
+    display: flex;
+    justify-content: space-between;
+  }
+
+  .rating {
+    width: 45%;
+    height: 20vw;
+  }
```
Awesome-E, line 225: Another rule that can be removed if we simplify the rating tile sizing

Awesome-E (changes requested):
most of my comments are about the same concept, but to put it visually, these are two rating tiles that have the same width... it would be better to have everything conform to a fixed width instead of having some parts of it depend on the user's view port width (i.e. why is the rating text bigger on one?)

charlieweinberger:
After doing further work on improving the UI of reviews, I've realized that the code I pushed a few hours ago isn't ready for prod. I'm working on a fully-functional version right now, which will also have a bunch of other improvements to the UI of reviews. Unfortuately, I got a little carried away and have changed a lot more than just fixing this issue, althought I still believe that I've simply solved a number of other bugs related to the UI/UX of reviews. Within the next day I'll update this pull request with all the changes, and you can be the judge.

charlieweinberger:
I've realized that its likely easier to create a new pull request, because of how much I changed and in the interest of preserving this pull request in case my new pull request is too much. The new pull request is #552: Improve review UI. Go check that out and respond over there. 

I'm really sorry if this entire situation is problematic. I don't know a lot about the good practices and typical norms of pull requests / team collaboration, so I apologize if I'm causing a mess here.

Awesome-E:
Closing as the follow-up PR was merged

Awesome-E:
I haven't deleted this branch yet, but if you agree there is nothing else we need from it, go ahead and delete it @charlieweinberger

PR #546, opened 2025-01-07
Move Feedback Form to Asana

### LINKED ISSUE #540: Update Feedback Form to Asana
As title states

Awesome-E:
This PR changes the Google Form link into the Asana feedback form. Additionally, the missing professor Google Form link is removed because professors populate from the API, and as such, issues should go to them instead.

## Reviewers

@timobraz requesting your review as primary review for code
@charlieweinberger review staging instance, and look at code if you want.

## Screenshots

N/A

## Test Plan

Check that the staging instance links to Asana feedback forms instead of Google Forms. There should only be two places, but let me know if there are any I missed:
- In the sidebar (open the hamburger menu), there should be a link that says Feedback on the bottom
- In the beta version tag that shows on desktop only

## Issues

Closes #540 
Asana Task: https://app.asana.com/0/1208487280290067/1208487339635800/f

### DIFF
--- a/site/src/component/AppHeader/AppHeader.tsx
+++ b/site/src/component/AppHeader/AppHeader.tsx
@@ -85,7 +85,7 @@ const AppHeader: FC = () => {
                     </a>
                     <a
                       className="ui button"
-                      href="https://forms.gle/JjwBmELq26daroTh9"
+                      href="https://form.asana.com/?k=4h9ZTRkVUT9ZwfJrmvxDDw&d=1208267282546207"
                       target="_blank"
                       rel="noopener noreferrer"
                     >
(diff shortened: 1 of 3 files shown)


timobraz:
LGTM, but I would suggest having a dedicated env variable for the feedback form link instead of replacing every occurrence of it in case it changes later.

timobraz (approved):

Awesome-E:
@charlieweinberger ready for your review

Awesome-E:
@timobraz we could consider that for the future, but we'd either have to create a new route to redirect to the form, or find a way to define frontend env variables since tmk that does not exist yet.

charlieweinberger:
LGTM, other than one tiny tiny thing. The description of the "How can we improve?" question mentions an option labeled "Issue/Bug" for the previous question, but the previous question lists just "Bug" as an option, not "Issue" or "Issue/Bug". Other than that, looks good.

PR #555, opened 2025-01-24
Zot4plan import button

### LINKED ISSUE #547: There should be a button to import 4-year plans from Zot4Plan
- User should be able to see a button on the roadmap page to import a schedule from Zot4Plan
- PeterPortal should be able to read data from Zot4Plan and display the 4-year plan in the roadmap tab.

We should not have to perform any write operations on the Zot4Plan database (not even deletes).

This most likely will take form of having a backend route in PeterPortal to retrieve the data based on the schedule ID, format it into our roadmap format, then return that to the user.

CadenLee2:
There is now a button on the roadmap page for users to import a schedule from [Zot4Plan](https://zot4plan.com/).
This opens a modal prompting them for their Zot4Plan schedule name and their year (to determine the starting year for their roadmap).
They then can import the schedule from Zot4Plan, which is formatted and loaded locally as a new roadmap in PeterPortal with the same name; the roadmap is not saved to their account/browser (yet).

## Description

- Added the button and modal
  - The modal requires the user to input a schedule name that is 8 characters or longer because no Zot4Plan schedule is less than 8 characters
  - There is a helpful screenshot to clarify to users which schedule name is being asked for
- On the backend, in a new route, PeterPortal now uses Zot4Plan's API to retrieve the schedule by name, then formats it into the correct roadmap format (the same format that PeterPortal roadmaps are saved as)
  - PeterPortal course names are Zot4Plan course names without the spaces. No exceptions have been found yet, but there are safeguards in place in case an invalid course name is encountered (see below)
  - Zot4Plan by default stores all 4 years, each with 4 quarters. Unnecessary (empty) additional years are trimmed off during conversion, and if the summer quarter is blank, it is not included
- On the frontend, the "saved roadmap" format is expanded into the full roadmap format with course details and is loaded into Redux using a dispatch
  - The new roadmap is automatically selected
  - Fixed an issue with the roadmap selection dropdown where the name displayed did not properly update from state
- Error handling/checking on the client side:
  - If the schedule was not found/imported successfully, the user is notified via a toast and it is not loaded
  - If during the expanding of the roadmap, courses with unrecognized names were found, they are removed and the user is notified how many problematic courses there were. This was never actually seen during testing with real Zot4Plan roadmaps, but artificially making the route return a fake course name caused the screen to go blank due to errors, so it is good to handle this just in case (especially if the roadmap is automatically saved in the future)
  - PeterPortal does not allow multiple roadmaps of the same name. If the imported roadmap has the same name as a current one, a "+" is appended to the end until there are no more conflicts (a simple solution because users can easily edit their roadmap names)

## Screenshots

The import modal:
Screenshot 2025-01-23 204800
A Zot4Plan schedule and the imported roadmap side by side:
Screenshot 2025-01-23 204948

## Test Plan

- Test loading various schedules from [Zot4Plan](https://zot4plan.com/). You can make schedules there by clicking "save" then typing in the new schedule name.
  - Existing schedules I used for testing include `test_migration`, `test_migration2`, and `test_migration_complex`
  - Try loading a schedule name that does not exist
  - Try loading a schedule name that is the same as an existing roadmap name in PeterPortal (notice the "+"'s appended to prevent name collision)
  - Try having/not having summer classes, having empty years in various places
- Code: the schedule conversion/roadmap verification code specifically is quite bulky because of all the various steps involved, but I tried to make it readable with comments and naming. This is a somewhat temporary feature so it isn't the most elegant, but if there are any glaring issues they can be fixed. Ideally it can be easily extended for the upcoming APs/auto saving features.

## Issues

Closes #547


api/src/controllers/zot4planimport.ts
```diff
@@ -0,0 +1,137 @@
...
+ */
+const getFromZot4Plan = async (scheduleName: string) => {
+  let res = {};
+  await fetch('https://api.zot4plan.com/api/loadSchedule/' + scheduleName, {
+    method: 'PUT',
+  })
+    .then((response) => {
+      if (!response.ok) {
+        return {};
+      }
```
Awesome-E, line 39: If we have an issue, I think it'd be better to return an error rather than an empty object. That way, we can tell the use instead of having something silently fail

api/src/controllers/zot4planimport.ts
```diff
@@ -0,0 +1,137 @@
...
+            },
+          ],
+          transfers: [],
+        };
+        // Determine the start year based on the current year and the student's year
+        let startYear = new Date().getFullYear();
+        if (input.studentYear === '1') startYear -= 1;
+        if (input.studentYear === '2') startYear -= 2;
+        if (input.studentYear === '3') startYear -= 3;
+        if (input.studentYear === '4') startYear -= 4;
```
Awesome-E, line 79:
```suggestion
        if (input.studentYear) startYear -= parseInt(input.studentYear);
```

api/src/controllers/zot4planimport.ts
```diff
@@ -0,0 +1,137 @@
...
+            // Convert quarter
+            const quarter = year[j];
+            const courses: string[] = [];
+            for (let k = 0; k < quarter.length; k++) {
+              // Convert course
+              // (PeterPortal course IDs are the same as Zot4Plan course IDs except all spaces are removed)
+              const originalCourseName = quarter[k];
+              const transformedCourseName = originalCourseName.replace(/\s/g, '');
+              courses.push(transformedCourseName);
+            }
```
Awesome-E, line 95: Everything inside this should be extracted to its own helper function. A generally good practice that you can take is for everywhere that you need a comment to explain what's going on, instead use a helper function and make that comment the name of the helper.

api/src/controllers/zot4planimport.ts
```diff
@@ -0,0 +1,137 @@
...
+      try {
+        // Get the raw schedule data
+        const originalScheduleRaw = await getFromZot4Plan(input.scheduleName);
+        if (Object.keys(originalScheduleRaw).length === 0) {
+          // Failed
+          return { planners: [], transfers: [] } as SavedRoadmap;
+        }
+
+        // Convert it to the PeterPortal roadmap format
+        const converted: SavedRoadmap = {
```
Awesome-E, line 65: This function is about getting the schedule, so any logic that doesn't fit the function name, i.e. converting a roadmap, should be in its own helper function (or even helper file).

api/src/controllers/zot4planimport.ts
```diff
@@ -0,0 +1,137 @@
...
+          }
+          converted.planners[0].content.push({
+            startYear: startYear + i,
+            name: 'Year ' + (i + 1),
+            quarters: quartersList,
+          });
+        }
+        // Trim trailing empty years other than the first year
+        // (do not trim empty years in the middle because that makes it hard to add years there)
+        while (converted.planners[0].content.length > 1) {
```
Awesome-E, line 113: `trimEmptyYears()` should be another helper function

site/src/pages/RoadmapPage/ImportZot4PlanPopup.scss
```diff
@@ -0,0 +1,26 @@
...
+  .import-schedule-warning-icon {
+    margin-bottom: 4px;
+    margin-right: 4px;
+    font-size: 16px;
+  }
+}
+
+[data-theme='dark'] {
+  .import-schedule-warning {
+    color: red;
```
Awesome-E, line 24: do you have css variables from the red color you used in the merge popover PR?

site/src/pages/RoadmapPage/ImportZot4PlanPopup.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const allPlanData = useAppSelector(selectAllPlans);
+
+  const obtainImportedRoadmap = async (schedName: string, currYear: string) => {
+    // Get the result
+    const result = await trpc.zot4PlanImportRouter.getScheduleFormatted.query({
+      scheduleName: schedName,
+      studentYear: currYear,
+    });
+    // Verify that the result has one planner (if not, the import failed)
+    if (result.planners.length == 0) {
```
Awesome-E, line 29: (related to first comment) if it returns an empty object, isn't `result.planners` undefined?

site/src/pages/RoadmapPage/ImportZot4PlanPopup.tsx
```diff
@@ -0,0 +1,157 @@
...
+      // Notify the user
+      spawnToast('The schedule "' + schedName + '" could not be successfully retrieved', true);
```
Awesome-E, line 31: "could not be retrieved" is more concise

site/src/pages/RoadmapPage/ImportZot4PlanPopup.tsx
```diff
@@ -0,0 +1,157 @@
...
+      return;
+    }
+    // Expand the result
+    const expandedPlanners = await expandAllPlanners(result.planners);
+    // Check for validity: length and invalid course names
+    if (expandedPlanners.length < 1) {
+      spawnToast('The schedule "' + schedName + '" could not be successfully imported', true);
```
Awesome-E, line 38: same as previous comment

site/src/pages/RoadmapPage/ImportZot4PlanPopup.tsx
```diff
@@ -0,0 +1,157 @@
...
+        while (i < quarter.courses.length) {
+          if (quarter.courses[i] === undefined) {
+            // An unknown course name was encountered when expanding the planner;
+            // It will crash PeterPortal if loaded, so remove it
+            quarter.courses.splice(i, 1);
+            problemCount++;
+          } else {
+            i++;
+          }
+        }
```
Awesome-E, line 54: While loop seems kinda finicky... maybe do filtering instead of splicing: `quarter.courses = quarter.courses.filter(...)`

site/src/pages/RoadmapPage/ImportZot4PlanPopup.tsx
```diff
@@ -0,0 +1,157 @@
...
+      }
+    }
+    if (problemCount > 0) {
+      spawnToast(
+        'The schedule "' + schedName + '" contained ' + problemCount + ' unknown course(s), which were removed',
```
Awesome-E, line 59: This message is a bit long for a toast, maybe something more like `Partially imported "${schedName}" (removed ${problemCount} unknown course(s))`

site/src/pages/RoadmapPage/ImportZot4PlanPopup.tsx
```diff
@@ -0,0 +1,157 @@
...
+      );
+    }
+    // Check for validity: the name should be unique among current planners
+    const takenNames = new Set<string>();
+    for (const planner of allPlanData) {
+      takenNames.add(planner.name);
+    }
+    while (takenNames.has(expandedPlanners[0].name)) {
+      // Users can change their planner name easily, so use a simple naming scheme
+      expandedPlanners[0].name += '+';
```
Awesome-E, line 70: Use the same logic as @timobraz did with Copying a roadmap. You should extract his logic to increment the number at the end of a name until it is unique, then use that helper here.

site/src/pages/RoadmapPage/ImportZot4PlanPopup.tsx
```diff
@@ -0,0 +1,157 @@
...
+              <Form.Label className="ppc-modal-form-label">Current Year</Form.Label>
+              <Form.Control as="select" onChange={(ev) => setStudentYear(ev.target.value)} value={studentYear}>
+                <option value="1" selected>
+                  I'm a first year
+                </option>
+                <option value="2">I'm a second year</option>
+                <option value="3">I'm a third year</option>
+                <option value="4">I'm a fourth year</option>
+              </Form.Control>
+            </Form.Group>
```
Awesome-E, line 138: This definitely works, but it could also be read as a sentence, i.e. label = "I am currently a..." + options = "1st year".

The options don't have to be verbose, in fact, I think "1st year" is better than "first year"

site/src/pages/RoadmapPage/ImportZot4PlanPopup.tsx
```diff
@@ -0,0 +1,157 @@
...
+            {busy ? 'Importing...' : 'Import'}
+          </Button>
+        </Modal.Body>
+      </Modal>
+      <Button
+        variant={darkMode ? 'dark' : 'light'}
+        className="ppc-btn import-schedule-btn"
+        onClick={() => setShowModal(true)}
+      >
+        <FileEarmarkArrowDown className="import-schedule-icon" />
```
Awesome-E, line 150: A file icon doesn't really make sense, I think something like the arrow going right into a square (similar to log in) or cloud with down arrow would make more sense

api/src/controllers/zot4planimport.ts
Awesome-E: This logic also won't work, because you need to account for whether this is run during fall quarter. If it's currently fall and someone says they're a first year, then the `startYear` would be the current year, not `startYear - 1`.

Awesome-E (changes requested):

site/src/pages/RoadmapPage/ImportZot4PlanPopup.scss
CadenLee2: I didn't, but since we're now using this in multiple places, I can move the red color into a new `theme.scss` variable

site/src/pages/RoadmapPage/ImportZot4PlanPopup.tsx
CadenLee2: Currently `getFromZot4Plan` returns an empty object, then I return `{ planners: [], transfers: [] } as SavedRoadmap` from the route. I'll change these to both raise errors.

api/src/controllers/zot4planimport.ts
CadenLee2: I set August (month index 7) to be the cutoff point where it's considered Fall, but it's a bit ambiguous. I'm thinking if someone who finished their 1st year is working on their schedule in August they'll consider themselves more a 2nd year

api/src/controllers/zot4planimport.ts
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * Determine a student's start year based on their current year in school
+ * (ex. a first-year's current year is "1")
+ */
+const getStartYear = (studentYear: string): number => {
+  let startYear = new Date().getFullYear();
+  startYear -= parseInt(studentYear);
+  // First-years in Fall start this year, not the previous year
+  if (new Date().getMonth() >= 7) startYear += 1;
```
Awesome-E, line 94: Technically this should work (might want to specify via comment that this is August, not July), but in a perfect world this would get schedules from [the API](https://anteaterapi.com/reference#tag/calendar/GET/v2/rest/calendar/all) and use that to determine what academic year we're in.

This can also be a separate issue since that's sorta separate from the rest of imports. cc @ecxyzzy for thoughts

Awesome-E (commented):
It looks like you might still be working on stuff so I'll leave this comment here for now. Re-request review once you're ready for me to take another look

Awesome-E (approved):
Will raise this as a new issue, but since Zot4Plan only offers 1 summer, the best assumption to make would probably be to do it as a 10 week summer

Awesome-E:
You're good to merge now

PR #562, opened 2025-01-27
Pending searches aborted

### LINKED ISSUE #541: Abort pending searches when the query is changed
Use `AbortController` to abort a pending fuzzy search. Otherwise, a scenario such as the following can occur:
1. User types in a query
2. User adjusts the query (a common scenario could be correcting a typo), but not before the search procedure call is sent
3. A new procedure call is sent for the updated query
4. The new procedure call receives a response first and the results update
5. The old procedure call receives a response second and then the results update again to incorrect ones

I believe this scenario can happen quite frequently if the second call hits the AnteaterAPI cache and the first one misses.

https://trpc.io/docs/client/vanilla/aborting-procedure-calls

There are probably other places in our codebase where we may want to abort procedure calls.

example:

https://github.com/user-attachments/assets/f9df8fc1-091e-4080-b49a-f5fb92362dca

timobraz:
## Description
As noticed by Jacob, sometimes old pending requests overwrite search results. This doesn't match the most recent search result typed in.

Now, the search will call to abort the previous controller whenever a new query is typed in and issues a new trpc query with a novel signal.

## Test Plan
1. Open staging instance
2. Go to devtools network tab and throttle yourself
3. Try searching for a course, quickly delete the search, and type in another course.
4. Observe that the previous search never overwrites the more recent request

## Issues

Closes #541


site/src/component/SearchModule/SearchModule.tsx
```diff
@@ -23,26 +23,52 @@ const SearchModule: FC<SearchModuleProps> = ({ index }) => {
...
     },
     [dispatch, index, search.pageNumber],
   );
 
+  // Cleanup abort controller on unmount
+  useEffect(() => {
+    return () => {
+      if (abortControllerRef.current) {
+        abortControllerRef.current.abort();
+      }
```
Awesome-E, line 68: Can be simplified
```suggestion
      abortControllerRef.current?.abort();
```

site/src/component/SearchModule/SearchModule.tsx
```diff
@@ -23,26 +23,52 @@ const SearchModule: FC<SearchModuleProps> = ({ index }) => {
...
-      dispatch(
-        setResults({
-          index,
-          results: results.map((x) => transformGQLData(index, x.result)) as CourseGQLData[] | ProfessorGQLData[],
-          count,
-        }),
-      );
+      if (abortControllerRef.current) {
+        abortControllerRef.current.abort();
+      }
```
Awesome-E, line 32: same as other comment

Awesome-E (approved):
LGTM – just a tiny thing there if you want to make that change

PR #573, opened 2025-01-28
Tim/autosave roadmap import

### LINKED ISSUE #549: Automatically save PeterPortal roadmap when a Zot4Plan schedule is imported
- After the schedule is imported and the roadmap is updated on the client, then save the roadmap
- If the student is logged in, save it to their account
- Otherwise, save it locally to their device
- In the Zot4Plan import modal, signify to the user that their roadmap will be saved afterwards (the text should say "Import & Save" instead of just "Import")

(This should be the same function as clicking the save button)

timobraz:
## Description
When importing an account from Zot4Plan, it now saves any unsaved changes prior to the import. This prevents loss of progress. The button now also says "Import and Save" instead of just "Import".

## Test Plan
- Open staging instance
- Go to your roadmap  (should work both being logged in or not)
- Make some change to the roadmap such as adding classes.
- Don't save them
- Import a Zot4Plan user.
- Notice that the changes have been saved and persist

## Issues

Closes #549


site/src/pages/RoadmapPage/RoadmapMultiplan.tsx
```diff
@@ -66,6 +66,15 @@ const RoadmapMultiplan: FC = () => {
   // const { name, content } = allPlans.plans[currentPlanIndex];
   const { name } = allPlans.plans[currentPlanIndex];
 
+  useEffect(() => {
+    const lastActiveIndex = localStorage.getItem('lastActiveRoadmapIndex');
+    if (lastActiveIndex !== null && allPlans.plans[parseInt(lastActiveIndex)]) {
+      console.log('lastActiveIndex', lastActiveIndex);
```
Awesome-E, line 72: Lint warnings in new code

Awesome-E (commented):
Just the lint warning

site/src/pages/RoadmapPage/ImportZot4PlanPopup.tsx
```diff
@@ -63,7 +66,7 @@ const ImportZot4PlanPopup: FC = () => {
     try {
       // Use the backend route to try to obtain the formatted schedule
       await obtainImportedRoadmap(scheduleName, studentYear);
-      // Success; hide the modal
+      saveRoadmap();
```
Awesome-E, line 69: You seem to call the save function, but the global state doesn't seem to be updated yet. Therefore, reloading the page after an import is not showing the updated planner.

screenshot of network request after importing a new "test_migration" schedule from z4p:
image

Awesome-E (changes requested):

Awesome-E (approved):
Works for me now!

PR #576, opened 2025-01-28
Adjust padding on main wrappers

Awesome-E:
This should make scrollbars less cramped, and scrolling containers will now get cut off by the bottom of the page instead of by the padding

I know the scrollbars are still not perfect, but they should at least be improved, and the contents should be cut off by the page instead of "arbitrarily"

## Screenshots

Before|After
:-:|:-:
before | after

Before|After
:-:|:-:
image | image

## Test Plan

-[ ] Verify that changes look as expected both on the Roadmap and Search pages


charlieweinberger:
Not sure if I agree that these changes are better than what we currently have.

For the course/prof search scrollbar, the scrollbar is now right inbetween the course/prof list and the course/prof info, but since its the scrollbar for the list, it should visually be closer to the list. Also, if the scroll bar is going to be to the right of the search bar (as opposed to under it), then I believe the scroll bar should go to the top of the search bar, not start below it. 
Screenshot 2025-01-30 at 11 19 25 AM

For the roadmap scrollbar, the first issue as above also applies here: the scrollbar should be visually closer to the roadmap years on the left than to the search area on the right. But now, the second issue from above is the opposite here: the scroll bar goes all the way to the top of the website, which it also should not to. It should again end at the same level as the top of the other divs next to it.
Screenshot 2025-01-30 at 11 25 27 AM

Another issue is the bottom of each scrollbar. Both scrollbars end at the bottom of the window, unlike previously. However, I'm not a big fan of this. I think that, similar to my idea above about where the top of the scrollbar should be, I think the bottom of the scrollbar should go to the bottom of the element, not the bottom of the window. I get that this is part of what you were trying to remove, but I'd argue that it shouldn't go to the bottom of the page because it's not the only thing being scrolled. Since we also have the course/prof info on the right, which is also scrollable but doesn't go to the bottom of the window, having the left element scroll to the bottom of the window doesn't make sense to me.
Screenshot 2025-01-30 at 11 38 08 AM

There's also some related changes that might be helpful:

For the course/prof search, if we're gonna make the user scroll anyways, do we really need the different pages? I say we have just one page, and when the user scrolls to the bottom of the page, we load in the next page automatically, instead of making them manually select the next page.
Screenshot 2025-01-30 at 11 29 57 AM

For the roadmap, it might be better to move the add year, import transcript, and import zot4plan schedule buttons to the top of the page. I get the logic behind the add year button being at the bottom, but the others dont need to be at the bottom, and having buttons below just doesn't make as much sense to me. This is definitely the least notable idea of any in this comment.
Screenshot 2025-01-30 at 11 32 31 AM

Another thing I'd potentially like to try is having the scroll bars below the search bar/blue header, respectively, but still to the right of the courses/profs/years.

Awesome-E:
> the scroll bar goes all the way to the top of the website, which it also should not to

> Both scrollbars end at the bottom of the window, unlike previously. However, I'm not a big fan of this. I think that, similar to my idea above about where the top of the scrollbar should be, I think the bottom of the scrollbar should go to the bottom of the element, not the bottom of the window

While this is true, I argue that it is far worse for the content itself to be cut off arbitrarily (see screenshots in PR desc) at the padding of the element. As far as I'm concerned, there is not a away around this.


> do we really need the different pages

No, feel free to create an issue to replace click-based pagination with scroll-based pagination. out of scope of this PR

> For the roadmap, it might be better to move the add year, import transcript, and import zot4plan schedule buttons to the top of the page

Yes, having it at the bottom was only the initial design. I agree that Importing is more concerned with the entire roadmap than the end of the roadmap; we just don't have a design for how this would go in the top yet.

> Another thing I'd potentially like to try is having the scroll bars below the search bar/blue header, respectively, but still to the right of the courses/profs/years.

I think it's good that you're thinking about the scroll bars, and would be happy to talk more about it. Let me know what you think about balancing scroll bar position vs padding and content cutoff position. We can talk more about this at some point in the near future

CadenLee2 (approved):
In terms of objective testing, the changes are working for me.
I'm on Windows so the scrollbar naturally looks slightly different (see screenshot), but it works as expected and the page is not cut off.

image

In terms of opinion, I agree that the roadmap scrollbar shouldn't extend all the way to the top of the page, but I didn't actually notice/get disoriented by it unless I paid attention.

Otherwise, everything does feel fine, and I do agree with Charlie that scroll-based pagination makes a lot more sense (for a future issue).

charlieweinberger (approved):

Awesome-E:
Finally got back to this PR after all that time – content on the right should be a sidebar now as we discussed. This will also be useful for when we add UI centered around the redesigned Transfer Experience

CadenLee2 (approved):
Tested:
- Courses and Professors sidebar, including selecting courses and scrolling on both the sidebar and the search results
- Roadmap sidebar, including selecting courses and scrolling on both the sidebar and the roadmap
- Results on mobile

Opinions:
- Padding amounts and the widths of the sidebars all look good
- Definitely feels much more natural to use

Code:
- CSS looks a bit intimidating in the `.header` class but it makes sense why it's necessary to have a lot because there's a lot going on, and it's possible to parse out what it means
- Organization of the CSS classes makes sense
- Is there a reason why the `` `${}` `` was changed to `''+()` string formatting style in [this one place specifically](https://github.com/icssc/peterportal-client/pull/576/files?w=1#diff-d346370d871034780e84d0cf003a1049587ebc3b242be456c9efb3779345b396R187) and nowhere else? Some specific linting thing?

Issue #554, opened 2025-01-17
Hovering over quarter offerings tooltip for CS 147 crashes PeterPortal


Awesome-E:
1. search for CS 147
2. Hover over the emojis that indicate which quarters it was offered in

Expected: I see the graph
actual: application crashes

Awesome-E:
Note: this also applies to courses with a "blank" quarters tooltip

PR #577, opened 2025-01-30
Fixed Range Error caused when tickSize is 0

ellisl10:
## Description

Hovering over the quarter offerings tooltip causes the website to crash due to tickSize calculation. The tickSize is now equal to 1 when it would otherwise be 0.

## Screenshots

After
Screenshot 2025-01-30 at 4 23 56 AM

## Test Plan

Verify that courses with less than 6 past offerings don't crash the webpage.

## Issues

Closes https://github.com/icssc/peterportal-client/issues/554


Awesome-E (approved):

Awesome-E:
You are good to merge this whenever you're ready

PR #584, opened 2025-02-04
Mobile missing prerequisites

### LINKED ISSUE #565: For mobile add course popup, Add Missing Prerequisite Text for courses that don't exist in the planner
the course info when you tap to add a course here should show the "Prerequisites not met: ..." text below the previous offerings
Image

CadenLee2:
Missing prerequisites are now shown when adding a course on mobile, if they exist

## Description

- Added missing prerequisites into the popup for adding courses on mobile, between all prerequisites and quarter offerings
  - When a course is tapped, the new `activeMissingPrerequisites` in the Redux roadmap slice is updated to store that course's missing prerequisites (similar to the existing `activeCourse` data)
  - This information is retrieved in the popup to be displayed
- Made styles for the `IncompletePrerequisiteText` component more reusable
  - In the mobile add course popup, its text size is larger than it is on the course popover
  - The appearance on the course popover has not changed
- Note: noticed and removed the unused `unmatchedPrerequisites` prop in `Course`

## Screenshots

Before:
image

After:
image

## Test Plan

- [ ] Test the staging instance (on mobile and on a small screen), with:
  - [ ] A course with missing prerequisites
  - [ ] A course with prerequisites where none of them are missing
  - [ ] A course with no prerequisites
- [ ] Ensure the course popover's appearance has not changed (because of style changes)

## Issues

Closes #565

### DIFF
--- a/site/src/store/slices/roadmapSlice.ts
+++ b/site/src/store/slices/roadmapSlice.ts
@@ -65,6 +65,8 @@ interface RoadmapSliceState {
   showAddCourse: boolean;
   // Store the course data of the active dragging item
   activeCourse?: CourseGQLData;
+  // Store the missing prerequisites of the active dragging item (for mobile)
+  activeMissingPrerequisites?: string[];
   // Whether or not to show the transfer modal
   showTransfer: boolean;
   // Store transfer course data
@@ -289,6 +291,9 @@ export const roadmapSlice = createSlice({
     setActiveCourse: (state, action: PayloadAction<CourseGQLData>) => {
       state.activeCourse = action.payload;
     },
+    setActiveMissingPrerequisites: (state, action: PayloadAction<string[] | undefined>) => {
+      state.activeMissingPrerequisites = action.payload;
+    },
     setYearPlans: (state, action: PayloadAction<PlannerData>) => {
       state.plans[state.currentPlanIndex].content.yearPlans = action.payload;
     },
@@ -372,6 +377,7 @@ export const {
   deleteYear,
   clearPlanner,
   setActiveCourse,
+  setActiveMissingPrerequisites,
   setYearPlans,
   setAllPlans,
   setInvalidCourses,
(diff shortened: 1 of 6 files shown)


site/src/store/slices/roadmapSlice.ts
```diff
@@ -65,6 +65,8 @@ interface RoadmapSliceState {
   showAddCourse: boolean;
   // Store the course data of the active dragging item
   activeCourse?: CourseGQLData;
+  // Store the missing prerequisites of the active dragging item (for mobile)
```
Awesome-E, line 68: Mobile doesn't exactly "drag", so "Stores missing prerequisites for courses when adding on mobile" would probably be more accurate.

For variable descriptions, we should also prefer to use JSDoc comments above them because then they can be intellisensed and shown as a description when `activeMissingPrerequisites` is in the autocomplete menu

Awesome-E (commented):
This is originally why I wasn't 100% sure on always showing the entire prerequisite text. I think that we should probably show the normal prerequisite text only if there are no missing prereqs, and only show the MissingPrerequisite text otherwise.

image

site/src/store/slices/roadmapSlice.ts
CadenLee2: There are a lot of these description comments above variables in roadmapSlice, all using `//`. Should I just change this one comment to be a JSDoc (would be inconsistent with the rest of the file), or should I change all of them (would be a lot of changes)?

Awesome-E: just change this comment accordingly. We can create a separate GH issue to change the rest.

Awesome-E (approved):
I personally think with the change you made is better because:
- Both ways, you know info about some of its prerequisites
- The info can't switch unless you close it, so there's no visible effect of switching which one is rendered

Awesome-E:
Good for now, but in the future we should parse the ANDs and ORs differently for missing prerequisites.

PR #586, opened 2025-02-05
Replaced incorrect uses of "dependency" to mean "dependent"

### LINKED ISSUE #543: Incorrect use of the word "dependency" in course prerequisite tree
In package resolution parlance, package A is said to be a *dependency* of package B if package A is required to install package B. This is also true in general usage.

As applied to courses we might say a course taken earlier (say ICS 31) is a *dependency* of a course taken later (say ICS 32). The converse relation is "dependent"; ICS 32 is a *dependent* of ICS 31. The relationship between "dependent" and "dependency" is a swap of the subject and object. When dealing with courses in particular, the word "prerequisite" is often used in place of "dependency"; ICS 31 is a *prerequisite* of ICS 32.

Screenshot_20241224_132449

PeterPortal (and, as a result, AntAlmanac) uses "dependency" when it means "dependent". Particularly in the prerequisite tree, suppose we are focusing on course ICS 32. Then the single prerequisite, ICS 31, is shown to the right. Then one might expect the courses to the left, which require ICS 32, to be referred to as "dependents", but they are instead called "dependencies".

This incorrect naming is also present in Anteater API, meaning any change would be frontend-only. But it is still worth considering.

If consensus is reached agreeing on change, I am willing to see through implementing it.

philosolog:
## Description
Changed every occurrence of "dependency" or "dependencies" to "dependent" and "dependents."

## Screenshots
N/A

## Test Plan
Make sure the staging instance works.

## Issues
Closes #543

## Note
AAPI needs to have the term "dependencies" updated to "dependents."

### DIFF
--- a/site/src/helpers/util.tsx
+++ b/site/src/helpers/util.tsx
@@ -81,7 +81,7 @@ function transformCourseGQL(data: CourseAAPIResponse) {
   const course = { ...data } as unknown as CourseGQLData;
   course.instructors = Object.fromEntries(data.instructors.map((instructor) => [instructor.ucinetid, instructor]));
   course.prerequisites = Object.fromEntries(data.prerequisites.map((prerequisite) => [prerequisite.id, prerequisite]));
-  course.dependencies = Object.fromEntries(data.dependencies.map((dependency) => [dependency.id, dependency]));
+  course.dependents = Object.fromEntries(data.dependencies.map((dependency) => [dependency.id, dependency])); // TODO: Change "dependencies" to "dependents" once it is changed in AAPI.
   return course;
 }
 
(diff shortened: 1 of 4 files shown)


site/src/helpers/util.tsx
```diff
@@ -81,7 +81,7 @@ function transformCourseGQL(data: CourseAAPIResponse) {
   const course = { ...data } as unknown as CourseGQLData;
   course.instructors = Object.fromEntries(data.instructors.map((instructor) => [instructor.ucinetid, instructor]));
   course.prerequisites = Object.fromEntries(data.prerequisites.map((prerequisite) => [prerequisite.id, prerequisite]));
-  course.dependencies = Object.fromEntries(data.dependencies.map((dependency) => [dependency.id, dependency]));
+  course.dependents = Object.fromEntries(data.dependencies.map((dependency) => [dependency.id, dependency])); // TODO: Change "dependencies" to "dependents" once it is changed in AAPI.
```
Awesome-E, line 84: use JSDoc's format for consistency
```suggestion
  course.dependents = Object.fromEntries(data.dependencies.map((dependency) => [dependency.id, dependency])); /** @todo Change "dependencies" to "dependents" once it is changed in AAPI */
```

Awesome-E (commented):
Just a small nit

Awesome-E (approved):

PR #587, opened 2025-02-10
Multiple Major Selection

timobraz:
## Description
You can now select multiple majors in the course requirements list. A major can also have a selection of a specialization. 

Requirements are now fetched upon initial load and adding a new major.

You can also switch between the roadmaps without losing the selection of majors.

The loading state is now independent per each major requirements being loaded. 

## Screenshots
image

## Test Plan

- [ ] Add a major without a spec, i.e. Software Engineering. The courses for that major should show
- [ ] Add a second major, this one with a spec, i.e. Computer Science.
- [ ] Switch to the Minor tab, then switch back. The spec should still be empty.
- [ ] Choose a spec for the second major, and its courses should show.
- [ ] Switch to the GE tab, then switch back. The spec you chose should be saved.
- [ ] Reload the page. The specs should be saved, and everything should load properly
- [ ] Remove one, and it should disappear from the list.
- [ ] Remove the other, and it should disappear from the list.

## Issues

_(can't find an issue for this?)_


Awesome-E:
I've just started reviewing this, but a spec must be chosen for the majors before the course list can be shown. This is currently not the case.

image

Additionally, this header should be bold if not already (it might be already since our fonts are weird), and it should have the arrow to collapse (similar to how roadmap years and individual requirements work, but w/o a border).

site/src/pages/RoadmapPage/sidebar/MajorRequiredCourseList.tsx
```diff
@@ -37,23 +38,24 @@ const MajorRequiredCourseList: FC = () => {
...
-  const selectedMajor = useAppSelector((state) => state.courseRequirements.major);
-  const selectedSpec = useAppSelector((state) => state.courseRequirements.specialization);
+  const selectedMajors = useAppSelector((state) => state.courseRequirements.selectedMajors);
 
   const plans = useAppSelector((state) => state.roadmap.plans);
   const planIndex = useAppSelector((state) => state.roadmap.currentPlanIndex);
   const activePlanID = plans[planIndex].id;
 
   /** Must ONLY contain requirements for the selected major/spec. This is used to check whether a major is
    * already loaded, so it must be set to empty if we change the major or spec */
```
Awesome-E, line 48: This comment should be moved to the `courseRequirementsSlice` after `requirements: ProgramRequirement[];` or removed.

site/src/pages/RoadmapPage/sidebar/MajorRequiredCourseList.tsx
```diff
@@ -182,46 +185,84 @@ const MajorRequiredCourseList: FC = () => {
...
-          classNamePrefix="ppc-combobox"
-          placeholder="Select a specialization..."
-          theme={(t) => comboboxTheme(t, isDark)}
-        />
-      )}
-      {selectedMajor && (!selectedMajor.specializations.length || selectedSpec) && (
-        <>{resultsLoading ? loadingIcon : <ProgramRequirementsList requirements={requirements} />}</>
-      )}
+
+      {selectedMajors.map((majorWithSpec) => (
```
Awesome-E, line 220: This is way too much code to have in one component. If I recall correctly, we discussed that the spec selection and requirements display should be extracted into its own component, and that each of those should handle state.

Having all the code through a map makes this difficult to read. I'd name the separate component something like `<SingleMajorCourseList/>` or something of that manner.

site/src/pages/RoadmapPage/sidebar/MajorRequiredCourseList.tsx
```diff
@@ -182,46 +185,84 @@ const MajorRequiredCourseList: FC = () => {
...
+                    majorId: majorWithSpec.major.id,
+                    specialization: updatedSpec,
+                  }),
+                );
+                await fetchRequirements(majorWithSpec.major.id, updatedSpec?.id || null);
+                const updatedMajors = selectedMajors.map((m) =>
+                  m.major.id === majorWithSpec.major.id ? { ...m, specialization: updatedSpec } : m,
+                );
+                saveMajors(updatedMajors);
+              }}
```
Awesome-E, line 251: Inline handlers should not be this long.

site/src/pages/RoadmapPage/sidebar/MajorRequiredCourseList.tsx
```diff
@@ -182,46 +185,84 @@ const MajorRequiredCourseList: FC = () => {
...
+              className="ppc-combobox"
+              classNamePrefix="ppc-combobox"
+              placeholder="Select a specialization..."
+              theme={(t) => comboboxTheme(t, isDark)}
+              onFocus={async () => {}}
```
Awesome-E, line 256: Was there a reason for adding this?

site/src/pages/RoadmapPage/sidebar/MajorRequiredCourseList.tsx
```diff
@@ -182,46 +185,84 @@ const MajorRequiredCourseList: FC = () => {
...
+            />
+          )}
+          {resultsLoading[majorWithSpec.major.id] ? (
+            loadingIcon
+          ) : (
+            <ProgramRequirementsList requirements={majorWithSpec.requirements} />
+          )}
```
Awesome-E, line 263: I know this is a 'prettier' format, but having this across 5 lines makes this hard to read. After this is in a separate component, we should be able to fit it within its line length limit, hopefully.

site/src/pages/RoadmapPage/sidebar/MajorRequiredCourseList.tsx
```diff
@@ -182,46 +185,84 @@ const MajorRequiredCourseList: FC = () => {
...
+            if (!newMajors.find((m) => m.id === id)) {
+              dispatch(removeMajor(id));
+            }
+          });
+          newMajors.forEach((major) => {
+            if (!currentMajorIds.includes(major.id)) {
+              dispatch(addMajor(major));
+              fetchRequirements(major.id);
+            }
+          });
```
Awesome-E, line 206: I suppose this is fine, but I feel like there should be a better way to do it

site/src/pages/RoadmapPage/sidebar/MajorRequiredCourseList.tsx
```diff
@@ -74,105 +76,106 @@ const MajorRequiredCourseList: FC = () => {
...
+        }
+      }
+    };
+
+    loadMajorsAndSpecs();
+
+    return () => {
+      mounted = false;
+    };
+  }, [dispatch, activePlanID, majors, isLoggedIn, fetchRequirements]);
```
Awesome-E, line 172: Lint warning

site/src/pages/RoadmapPage/sidebar/MajorRequiredCourseList.tsx
```diff
@@ -74,105 +76,106 @@ const MajorRequiredCourseList: FC = () => {
...
-      dispatch(setMajor(foundMajor));
-      dispatch(setSpecialization(null));
-      setSavedSpecID(pairs[0]?.specializationId);
-      dispatch(setRequirements([]));
-    });
-  }, [dispatch, activePlanID, majors, isLoggedIn]);
+    const loadMajorsAndSpecs = async () => {
+      try {
+        const pairs = await trpc.programs.getSavedMajorSpecPairs.query(activePlanID);
+        if (!mounted) return; //mounted is false if the component is unmounted, prevents loop rendering issue i had
```
Awesome-E, line 138: This seems like it could be a structural issue. I think if each Major's spec/course list handles this on its own and calls the Redux store functions responsibly, you should be able to self-contain all of these instead of looping through pairs. Definitely easier said than done, but we could make some time to pair on this one.

Awesome-E (changes requested):

site/src/pages/RoadmapPage/sidebar/MajorRequiredCourseList.tsx
timobraz: This causes infinite refetching of the majors since it sets the selectedmajor

Awesome-E (commented):
Ready for someone else to review @CadenLee2 @ellisl10

CadenLee2 (commented):
## Overall
- Tested successfully while logged in and logged out, on mobile and desktop.
- Performed the steps in the Test Plan
  - When I had two tabs open, reloading the page didn't save my SWE and Computer Science, and instead loaded the major from the other tab (probably not an issue, though not sure why this happened)
- Main functionality seems to work well and is intuitive

## Opinions and Small Details
- I think the larger X on the multi major selection box is unnecessary because you can delete the individual majors quickly already, and no one will have so many majors that the larger X saves them any substantial amount of time. It takes up extra space and on mobile increases the chances of accidentally clearing all your majors.
image
- The X icon looks very squished, is there no better X icon?
- The arrow to expand the requirements is not vertically centered at the same central level as the text
image
- Scrolling to the bottom of major requirements now has a small bit of extra padding compared to minor/GEs (both when the bottom major are expanded and collapsed)
image


## Issues

> Requirements are now fetched upon initial load and adding a new major.

This seems to only be the case with majors with specializations; majors without specializations appear to be re-fetched every time. For instance, this is seen in the SWE major in the third step of the Test Plan

### Can't delete on first click after switching tabs while logged in
1. Add Anthropology, B.A. and nothing else (while logged in)
2. Click the Minor tab
3. Click back to the Major tab
4. Click the X on Anthropology. The selection box will load for a second, then Anthropology will reappear.
5. Click the X again, and Anthropology will successfully be deleted.

Note: the same issue happens when clicking the larger X on the whole selection box.

Awesome-E:
> When I had two tabs open, reloading the page didn't save my SWE and Computer Science, and instead loaded the major from the other tab (probably not an issue, though not sure why this happened)

These are saved on select, not when saving the roadmap. So yes that is possible but should never occur in most use cases.

> The arrow to expand the requirements is not vertically centered at the same central level as the text
> The X icon looks very squished, is there no better X icon?

The X comes with the library, and for the arrow, I assumed they were but that can be fixed alongside the ticket for the other X.

> Scrolling to the bottom of major requirements now has a small bit of extra padding compared to minor/GEs (both when the bottom major are expanded and collapsed)

Interesting observation. Can take a look

> The selection box will load for a second, then Anthropology will reappear.

So you wait for everything to load, but removing just causes it to reappear the first time, right?

ellisl10:
Looks good overall, but a small nitpick is that having a major with a long name causes the whole sidebar to change size upon expanding and minimizing the selected major(s)
image
image

site/src/pages/RoadmapPage/sidebar/SingleMajorCourseList.tsx
```diff
@@ -79,7 +79,10 @@ const SingleMajorCourseList: FC<SingleMajorCourseListProps> = ({
   );
 
   const loadSpecRequirements = useCallback(async () => {
-    if (!hasSpecs) return await fetchRequirements(major.id, null);
+    if (!hasSpecs) {
+      if (majorWithSpec.requirements.length > 0) return;
+      else return await fetchRequirements(major.id, null);
+    }
```
Awesome-E, line 85: We're getting a lint warning with the deps array; see if you can fix the lint warning and still achieve the behavior you're mentioning. If so, we are good to merge the PR I think.

Awesome-E: Update: made the commit, double check that it works

CadenLee2 (approved):
- Thanks for catching the lint warning, I'll look into why I didn't see that in my editor. The intended requirements fetching behavior is working still :white_check_mark: 
- I like the removed large X on the selection box
- Style changes addressing the above comments (arrow location, majors with long names, padding) all look good
- Main functionality seems to work like before
  - Did the test plan steps again, both logged in and logged out

ellisl10 (approved):

Awesome-E (approved):
SHIP IT!

PR #595, opened 2025-02-12
Implement minor requirements

### LINKED ISSUE #569: Show Required Courses for a Minor
_From the requirements document:_

> Similar to the “Major” tab, the “Minor” tab should contain a selection box (with search capability) to select any undergraduate minor program as listed in DegreeWorks or the [Catalogue](https://catalogue.uci.edu/undergraduatedegrees/).
>
> Below the selection box for a user’s Minor, there should also be a list of courses required, similar to Majors.

ellisl10:
Implement minor requirements

## Description
Added a required course list for minors and a 'Minor' option in the requirements selector.

## Screenshots

Before
Screenshot 2025-02-12 at 12 22 15 AM

After
Screenshot 2025-02-12 at 12 22 52 AM

## Test Plan

Make sure that minor requirements load and appear as intended in different roadmaps.

## Issues

Closes #569

### DIFF
--- a/site/src/pages/RoadmapPage/sidebar/MinorRequiredCourseList.tsx
+++ b/site/src/pages/RoadmapPage/sidebar/MinorRequiredCourseList.tsx
@@ -4,12 +4,7 @@ import Select from 'react-select';
 import trpc from '../../../trpc';
 import { normalizeMajorName, comboboxTheme } from '../../../helpers/courseRequirements';
 import { Spinner } from 'react-bootstrap';
-import {
-  setRequirements,
-  setSpecialization,
-  setMinor,
-  setMinorList,
-} from '../../../store/slices/courseRequirementsSlice';
+import { setRequirements, setMinor, setMinorList } from '../../../store/slices/courseRequirementsSlice';
 import { useAppDispatch, useAppSelector } from '../../../store/hooks';
 import ThemeContext from '../../../style/theme-context';
 import { MinorProgram } from '@peterportal/types';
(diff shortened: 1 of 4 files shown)


site/src/pages/RoadmapPage/sidebar/MinorRequiredCourseList.tsx
```diff
@@ -0,0 +1,120 @@
...
+        options={minorSelectOptions}
+        // must be NULL, not undefined, to clear the field for unselected values
+        value={minorSelectOptions.find((o) => o.value === selectedMinor) ?? null}
+        isDisabled={minorsLoading}
+        isLoading={minorsLoading}
+        onChange={(data) => {
+          if (data!.value.id === selectedMinor?.id) return;
+          dispatch(setRequirements([])); // set to empty immediately because otherwise it's out of date
+          dispatch(setMinor(data!.value));
+          dispatch(setSpecialization(null));
```
Awesome-E, line 108: spec should only be set for major

site/src/pages/RoadmapPage/sidebar/MinorRequiredCourseList.tsx
```diff
@@ -0,0 +1,120 @@
...
+    if (!activePlanID) return;
+    if (!isLoggedIn) return;
+
+    setMinorsLoading(true);
+    trpc.programs.getMinors.query();
+    setMinorsLoading(false);
+
+    const foundMinor = minors.find((m) => m.id === minors[0]?.id) ?? null;
+    dispatch(setMinor(foundMinor));
+    dispatch(setRequirements([]));
```
Awesome-E, line 82: This looks like it always returns the first minor instead of the one the user saved. We should have a function that lets you save minor to an account, though this is something I'll ask Tim about

Awesome-E: Update: there is no database change as of yet, so it would be possible to copy the logic over for saving a minor to an account.

It's a little bit different though, as you'll be modifying a single entry in the planner table instead of adding a row to `planner_major`. However, given that it is possible to have multiple minors, it might not be worth adding the ability to save until after we change it to multiple minors (or at least merge this branch). Thoughts?

ellisl10: I think we should probably have a save function even for a single minor for now

Awesome-E (approved):
Approving this as it wasn't intended to save Minor to the database yet. We will figure this out after merging. Adding @CadenLee2 for secondary review.

PR #558, opened 2025-01-25
Make the search result cards more compact

Awesome-E:
## Description

- Adds the bookmark button into the course info popover
- Fixes prerequisites not showing up in course search properly (I think an earlier PR of mine broke this)
- Fixes broken popover tooltips for user profile and 3 dots menus on the roadmap page

## Screenshots

Before:
image
image
image

After:
image
image
image

## Test Plan

- [ ] Test that the changes look as expected on staging instance

PR #596, opened 2025-02-12
Extract pluralizing to a helper

### LINKED ISSUE #564: For any text that depends on whether a quantity is 1, extract to a helper
anything that looks like `list.length === 1 ? 's' : ''` should be replaced with a call to a helper function.

This helper function is being created in #558, and this issue just says that any other occurrences of this pattern should use this helper.

CadenLee2:
All logic for pluralizing (ex. `{maxUnits === 1 ? 'unit' : 'units'}`) now uses the helper function (ex. `unit{pluralize(maxUnits)}` from #558

## Description

- Replaced all logic for pluralizing (see above) with the helper function
- Fixed the course popover's unit count text, which originally was plural "units" regardless of whether the course was a singular unit

## Screenshots

(The only visible behavior that should have changed is the course popover's display of units)

## Test Plan

Nothing should have changed other than the course popover. To verify on the staging instance:

- By using classes of different unit types (
    - More than one unit (most classes fit this category)
    - Variable units (e.g. `Mol Bio 243`, 2-4 units)
    - One unit (e.g. `Mol Bio 229`, 1 unit)
- ), sufficiently test that the text on various components is plural/singular as it should be, like:
    - Unit counts in all steps of adding a course on mobile
    - Unit counts on courses in roadmap/search results
    - The totals on the blue header bar (with only 1 course, 1 unit)
    - Year info in the top right of a year
- Test course popover (the pluralizing behavior of units should now be accurate, including for single unit courses)

## Issues

Closes #564

### DIFF
--- a/site/src/component/CoursePopover/CoursePopover.tsx
+++ b/site/src/component/CoursePopover/CoursePopover.tsx
@@ -25,7 +25,7 @@ const CoursePopover: FC<CoursePopoverProps> = ({ course, interactive = true, req
       <div className="popover-name">
         {department + ' ' + courseNumber + ' '}
         <span className="popover-units">
-          ({minUnits === maxUnits ? minUnits : `${minUnits}-${maxUnits}`} {pluralize(maxUnits, 'units', 'unit')})
+          ({minUnits === maxUnits ? minUnits : `${minUnits}-${maxUnits}`} unit{pluralize(maxUnits)})
         </span>
 
         <div className="spacer"></div>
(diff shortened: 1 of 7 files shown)


Awesome-E:
Did you mean to request review instead of assigning the PR?

CadenLee2:
Oh oops, yeah sorry about that

Awesome-E (approved):
I thought there might have been things that use `> 1` for a check but can't find it, so it does look like you got everything!

PR #597, opened 2025-02-19
Fix AP/CC courses not being imported from transcript if a Year 0 Summer exists

### LINKED ISSUE #594: Import Transcript does not import AP courses
## Steps to recreate
- Save a transcript containing AP credits (from webreg, following the directions on PeterPortal) and import it
- Notice that transfer credits does not include any of these AP courses
- The AP courses should be imported

Awesome-E:
## Description

Use the `schools` view to obtain transferred AP/College Courses, as this seems to be at the top regardless of whether a Year 0 summer quarter exists.

## Screenshots

N/A

## Test Plan

- [ ] Go to the staging instance
- [ ] Try importing a transcript where the user has a summer quarter before their first year
- [ ] Try importing a transcript where the user does not have a summer before first year.

## Issues

Closes #594

### DIFF
--- a/site/src/pages/RoadmapPage/ImportTranscriptPopup.tsx
+++ b/site/src/pages/RoadmapPage/ImportTranscriptPopup.tsx
@@ -38,22 +38,25 @@ function toCourseID(course: TranscriptCourse) {
   return (course.dept + course.code).replace(/\s/g, '');
 }
 
-function processRow(row: HTMLElement, transfers: TransferUnitDetails[], quarters: TranscriptQuarter[]): void {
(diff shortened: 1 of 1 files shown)


CadenLee2 (approved):
- [x] Tested with summer quarter before first year
  - All courses imported correctly
  - All AP/transfer credits (including CC courses) imported correctly
- [x] Tested without summer quarter before first year
  - Appears that all courses imported correctly
  - Appears that all AP/transfer credits imported correctly

Awesome-E:
Will be merging in the afternoon

PR #598, opened 2025-02-19
Create tRPC Router for Reading Planners

### LINKED ISSUE #518: Create previously taken courses endpoint for AntAlmanac integration
https://github.com/icssc/AntAlmanac/issues/1028

* Set up a REST endpoint in our API for querying previously taken courses that takes a google account id and the current quarter as params
  * should create a new express router for this endpoint (and any future ones)
* Verify a secret key that is passed in the Authorization header so only AntAlmanac can use this endpoint
* Respond with a string[] of course ids for all courses that come before the specified quarter in the user's planner(s)

Note, our backend is same-origin, need to make sure to allow CORS for just this endpoint so AntAlmanac can access it

Awesome-E:
The tRPC Route `/api/trpc/external.roadmaps.getByGoogleID` will now return user planners when given an auth token and a matching Google User ID

## Screenshots

N/A

## Test Plan

- [ ] Make a request to the staging instance `/api/trpc/external.roadmaps.getByGoogleID?input={"googleUserId":"<google_user_id>"}` with the Bearer token. Ask privately for a sample Google User ID and the bearer token
- [ ] Ensure that the roadmaps for only that user are returned
- [ ] Check that unauthorized requests fail

## Issues

Closes #518

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -44,3 +44,4 @@ jobs:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           ANTEATER_API_KEY: ${{ secrets.ANTEATER_API_KEY }}
+          EXTERNAL_USER_READ_SECRET: ${{ secrets.EXTERNAL_USER_READ_SECRET }}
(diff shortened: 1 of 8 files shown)


MinhxNguyen7:
Do you think there's a practical way for us to import your tRPC router, or is that not really possible?

CC: @alexanderl19

MinhxNguyen7:
I assume the googleUserID is universal, right?

Awesome-E:
> I assume the googleUserID is universal, right?

I believe so. If you have Google Auth on your side, you can let me know your Google User ID in AA and I can see if it matches here.

CadenLee2 (approved):
Tested with the Thunder Client VS Code extension
- With my Google ID and the correct authorization: the results correctly match all the roadmaps saved to my account
- With a presumably nonexistent Google ID and the correct authorization: successfully got a result of `{ "result": { "data": [] } }`
- With no authorization header: failed with the UNAUTHORIZED response
- With incorrect authorization: failed with the UNAUTHORIZED response

MinhxNguyen7 (approved):
Vicky said this works

PR #607, opened 2025-02-20
Hide grades with zero people who got them in grade dist pie chart

charlieweinberger:
## Description

Within the grade distribution for a course is a pie chart that shows how many people got each grade (A, B, etc.) in the course. However, the pie chart shows all possible grades, even if no one got those grades. This creates very slim elements, basically lines, for those grades, which doesn't look great. This one-line commit fixes that by filtering out the grades that don't have any people who got them.

Not totally sold on this commit - I do think it looks better to hide the grades with zero people, but that also means having less information. I think this tradeoff is worth it, but what do you guys think?

### Example 1 

Screenshot 2025-02-19 at 9 00 35 PM Screenshot 2025-02-19 at 9 00 26 PM

### Example 2

Screenshot 2025-02-19 at 9 00 40 PM Screenshot 2025-02-19 at 9 00 48 PM

## Test Plan

- [x] Verify that the grades don't show up if 0 people who got them, but that all other grades do show up.

## Issues

Does not close an existing issue.


Awesome-E:
> Not totally sold on this commit - I do think it looks better to hide the grades with zero people, but that also means having less information. I think this tradeoff is worth it, but what do you guys think?

Given that there is also a bar chart that shows all of the options regardless of whether there are any data points, I think it's fine for the pie to omit ones that have no data.

Awesome-E (approved):

PR #610, opened 2025-02-25
change text highlight color from uci-gold to peterportal-primary-color-2

### LINKED ISSUE #604: Get rid of the yellow text selection color
It just doesn't look good.

charlieweinberger:
## Description

Changed the color of the text highlight from uci-gold to ~~peterportal-primary-color-2 (a light blue). I can also easily change the highlight color to the default browser color, but I kinda like this blue. I tried peterportal-primary-color-1 (the main blue throughout), but it was too dark to easily read text. I don't believe that peterportal-primary-color-2 is used for any text. Thoughts on this color vs. the default highlight color?~~ the default browser text highlight color.

## Screenshots

Before:
Screenshot 2025-02-24 at 11 26 14 PM

After:
Screenshot 2025-02-24 at 11 25 30 PM

For reference, here's what peterportal-primary-color-2 looks like:
Screenshot 2025-02-24 at 11 22 47 PM

## Test Plan

- [ ] Verify that the new color is correctly applied throughout the application
- [ ] Verify that the new color doesn't overlap with any existing text
- [ ] Verify that the new color works on both light mode and dark mode

## Issues

Closes #604

### DIFF
--- a/site/src/index.css
+++ b/site/src/index.css
@@ -36,18 +36,6 @@ code {
   --border-radius: 8px;
 }
 
-::-moz-selection {
-  background: var(--peterportal-primary-color-2);
-  color: black;
-  opacity: 1;
-}
-
-::selection {
-  background: var(--peterportal-primary-color-2);
-  color: black;
-  opacity: 1;
-}
-
 .hide {
   display: none;
 }
--- a/site/src/index.css
+++ b/site/src/index.css
@@ -37,13 +37,13 @@ code {
 }
 
 ::-moz-selection {
-  background: var(--peterportal-uci-gold);
+  background: var(--peterportal-primary-color-2);
   color: black;
   opacity: 1;
 }
 
 ::selection {
-  background: var(--peterportal-uci-gold);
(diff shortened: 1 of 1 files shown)


Awesome-E (changes requested):
Use the default highlight color. People will be used to it because it's what they see on most other sites anyways. In my opinion, there's no strong reason we need the highlight color to be different.

charlieweinberger:
Updated 👍

Awesome-E (approved):
Oh, I see. The reason the "default" might look kinda funny is because semantic UI is overriding selection colors as well in many areas.

We should still merge this – also letting @ethank2222 know that finishing Semantic UI removal will fully fix this.

Awesome-E:
That said, if you do think that there are too many inconsistencies with how selection works (I found out some through text inputs, for example), we can use the PeterPortal color temporarily if you prefer that.

charlieweinberger:
I don't have a strong preference, and since we're working towards removing Semantic UI, I'd prefer to help that along by sticking with the default highlight.

Issue #603, opened 2025-02-19
Bump our "beta" version to 1.2 and generate release notes for Degree Requirements


Awesome-E:
The version in this tag should be updated for a release like this:
<img width="150" alt="Image" src="https://github.com/user-attachments/assets/9de7067d-e6b2-4ce0-ad2b-7c1637694653" />

In this same PR, we could also change the colors to update it to match the Figma:
<img width="180" alt="Image" src="https://github.com/user-attachments/assets/b5c8780f-21e1-4052-a327-2324bda337a0" />

Additionally, there should be a change log for Degree Requirements.

Awesome-E:
completed by #612

PR #612, opened 2025-02-26
Degree requirements changelog

CadenLee2:
Updated version/changelog information for the degree requirements feature

## Description

- Added screenshots and a description to the changelog for the new degree requirements feature
- Incremented the version in the beta tag to 1.2
- Note: the beta tag colors should be updated to match the Figma, but this involves Semantic UI, so that can be done in a separate PR

## Screenshots

image
image

## Test Plan

- [ ] Test that the updated changelog appears (and that the screenshots/release notes are good)
- [ ] Test that the version is 1.2

## Issues

This contributes to closing, but does not close, #603 because the color should also be updated


site/src/component/ChangelogModal/ChangelogModal.tsx
```diff
@@ -1,10 +1,11 @@
 import { useEffect, useState } from 'react';
 import './ChangelogModal.scss';
 import Modal from 'react-bootstrap/Modal';
-import changelogImage from '../../asset/hover-course.png';
+import changelogImage1 from '../../asset/degree-reqs-major.png';
+import changelogImage2 from '../../asset/degree-reqs-minor.png';
 
-const DESCRIPTION = 'Along with visual updates, course info is now found when you hover over the course title';
-const LAST_UPDATED = '01/27/2025';
+const DESCRIPTION = 'You can now view completion of major, minor, and GE requirements on your roadmap';
```
Awesome-E, line 7: for emphasis!
```suggestion
const DESCRIPTION = 'You can now view completion of major, minor, and GE requirements on your roadmap!';
```

site/src/component/ChangelogModal/ChangelogModal.scss
```diff
@@ -31,6 +31,14 @@
...
   }
 
+  .modal-side-by-side {
+    display: flex;
+    flex-direction: row;
+    .modal-img {
+      width: 50%;
+    }
+  }
+
```
Awesome-E, line 41: Gap between the two images should be less than the margin outside of them. I'd suggest using the `gap` property and `padding` instead of `margin` on the images (you can override the images for this changelog

site/src/component/ChangelogModal/ChangelogModal.tsx
```diff
@@ -34,7 +35,10 @@ const ChangelogModal = () => {
       </Modal.Header>
 
       <p className="modal-body">{DESCRIPTION}</p>
-      <img className="modal-img" src={changelogImage} alt="Screenshot or gif of new changes" />
+      <div className="modal-side-by-side">
```
Awesome-E, line 38: Not sure how I feel about there being a new class just for this change log – we should at least leave a comment saying to remove the class if we remove this element from the change log. Also, this will be the first time I concede that Tailwind is better for something.

site/src/component/ChangelogModal/ChangelogModal.tsx
```diff
@@ -34,7 +35,10 @@ const ChangelogModal = () => {
...
+        <img className="modal-img" src={changelogImage1} alt="Screenshot or gif of new changes" />
+        <img className="modal-img" src={changelogImage2} alt="Screenshot or gif of new changes" />
```
Awesome-E, line 40: On mobile this doesn't look good. We should hide the second image if the screen size dips below a certain amount.

site/src/component/AppHeader/AppHeader.tsx
```diff
@@ -55,7 +55,7 @@ const AppHeader: FC = () => {
               position="bottom right"
               trigger={
                 <Label as="a" color="yellow" image>
-                  beta<Label.Detail>v1.1</Label.Detail>
+                  beta<Label.Detail>v1.2</Label.Detail>
```
Awesome-E, line 58: @ethank2222 here's something you might have to keep in mind / deal with when updating the `remove-semantic` branch

Awesome-E (changes requested):

site/src/component/ChangelogModal/ChangelogModal.tsx
charlieweinberger: I agree, this is a rather substantial feature addition and it should be celebrated as such

site/src/component/ChangelogModal/ChangelogModal.scss
charlieweinberger: I agree with this suggestion. Also, if possible, consider rounding the borders of the images to match the rounding of the popup.

charlieweinberger:
On my laptop, the second image overflows out of the modal. Definitely should be fixed.
Screenshot 2025-02-25 at 11 14 47 PM

charlieweinberger:
However, do we really need two images? The only difference between these two images is major vs. minor, but there isn't a meaningful difference in the images in my opinion to warrant both. Maybe this would be necessary if the second image was for GEs instead of minors?

charlieweinberger (changes requested):
I don't think we need an image for both major and minor degree requirements, since the two are very similar. I suggest either replacing the minor degree req image with an image for GEs, or simply removing it and only having the one image for majors.

CadenLee2:
I made some changes which can be viewed on the staging branch (may have to clear local storage to see the changelog), but I'm not sure whether they fix these problems.
> On my laptop, the second image overflows out of the modal

That's a bit strange, it didn't happen to me in Firefox or a Chromium browser. Is it still happening?
> I suggest either replacing the minor degree req image with an image for GEs, or simply removing it and only having the one image for majors

I agree that having two images makes things a bit more awkward, but the problem is that this feature is very vertical, unlike prior changelog features, which means only one image would have a strange aspect ratio. For now, I replaced the minor reqs image with GE reqs, and when the screen is small, only the major reqs show.

What do you guys think about having just one landscape image that includes more of the roadmap but focuses on the major requirements in the sidebar? That could both make things a lot simpler and help contextualize the feature so users are more immediately aware of where it is without reading, like:
Screenshot 2025-02-27 083854

> consider rounding the borders of the images 

This would probably look good if we end up going with two vertical images

charlieweinberger:
> That's a bit strange, it didn't happen to me in Firefox or a Chromium browser. Is it still happening?

Doesn't look like it's still happening. Not sure what why it was caused in the first place.

> What do you guys think about having just one landscape image that includes more of the roadmap but focuses on the major requirements in the sidebar? That could both make things a lot simpler and help contextualize the feature so users are more immediately aware of where it is without reading, like:

I think I slightly prefer this, although I don't have a big preference on this vs. the two images of majors and GEs. Either option is good enough for showing the changes, no need to spend a lot of time thinking about this choice (in my opinion).

> This would probably look good if we end up going with two vertical images

I think it'd look good no matter how many images we have. Since the modal has rounded corners, the images inside should also have rounded corners, imo. However since the rounding isn't much, this also isn't a big deal.

charlieweinberger (approved):
LGTM 👍

Awesome-E (approved):
💪

PR #615, opened 2025-03-04
Course Info Tooltip has Quarter Offering

### LINKED ISSUE #601: Course Info Tooltips Should Have the Quarter Offering Tooltip
This would be especially useful for degree requirements – this would save the user having to open the course page or search for the major/minor/GE-required course every time to see when it was offered.

_Note: "Course Info Tooltip" refers to the popover that appears when you hover over a course title, but we should probably call it a tooltip since that language is more universal_

timobraz:
## Description
Now when hovering over the tooltip for courses, the quarter offering will be shown

## Screenshots
image
image

## Test Plan
Make sure the course tooltip includes the quarter offering in both the searched course and the ones on the roadmap.

## Issues

Closes #601


site/src/component/CoursePopover/CoursePopover.tsx
```diff
@@ -43,6 +44,14 @@ const CoursePopover: FC<CoursePopoverProps> = ({ course, interactive = true, req
...
         <CorequisiteText course={course} />
         <IncompletePrerequisiteText requiredCourses={requiredCourses} />
+        {course.terms && course.terms.length > 0 && (
+          <div className="quarter-offerings">
+            <b>Quarter Offerings:</b>
+            <div className="quarter-indicator-wrapper">
+              <CourseQuarterIndicator terms={course.terms} size="sm" />
+            </div>
+          </div>
+        )}
```
Awesome-E, line 54: Take a look at `AddCoursePopup.tsx` lines 83-86. Would it be worth it to combine that with these lines in a separate component and just call that component in both places? That should also let the style be shared between the two.

timobraz:
This is what the component now looks like (taken from AddCoursePopup)
image

Awesome-E (approved):

PR #618, opened 2025-03-08
Update PeterPortal's Theme colors when the system theme changes

### LINKED ISSUE #611: When the system color scheme changes, PeterPortal should change if set to system theme
You can use `window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', ({ matches }) => ...)` to detect when the system color scheme changes.

https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/change_event

We should use this to automatically switch between light and dark themes when the user selects "System Theme" as their preference instead of having to reload the page.

Awesome-E:
## Description

Adds an event listener to detect when the user changes their system color scheme, and update PeterPortal's theme accordingly if they prefer system color scheme.

## Test Plan

- [ ] Set your PeterPortal theme to light. Change your system color scheme a few times. Nothing should happen.
- [ ] Repeat for dark theme
- [ ] Set your PeterPortal theme to System. Then, change your system color scheme a few times. Each time you change your system color scheme, PeterPortal should update.

## Issues

Closes #611

### DIFF
--- a/site/src/App.tsx
+++ b/site/src/App.tsx
@@ -67,6 +67,14 @@ export default function App() {
     }
   }, []);
 
+  useEffect(() => {
+    const setSystemTheme = () => setThemeState('system');
+    const matcher = window.matchMedia('(prefers-color-scheme: dark)');
+
+    if (usingSystemTheme) matcher.addEventListener('change', setSystemTheme);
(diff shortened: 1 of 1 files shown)


charlieweinberger (approved):
LGTM

PR #629, opened 2025-04-04
Store expanded degree requirement groups

### LINKED ISSUE #624: Store which sections have been collapsed/expanded in degree requirements
When the user expands or collapses course requirement groups, this should be saved to the global state so that switching back to the tab doesn't reset it to the default state.

i.e. clicking this should save it:
Image

CadenLee2:
When you expand or collapse groups for major, minor, and GE requirements, it is now saved to global state, so it is preserved when switching tabs.

## Description

- Added a variable into the course requirements slice, `expandedGroups`, that stores all the unique keys of the requirement groups that have been opened/expanded
- Each requirement group now has a unique string key, called `storeKey`, and uses it to get/set whether it is opened/expanded in the store, rather than in its local state

Visualization of how these keys are determined/used:
```
courseRequirementsSlice                Displayed requirements for major 123ABC     Their storeKeys
  expandedGroups: {              ------  v Requirement A                             major-123ABC-0
                                 |         > Sub Requirement 1                       major-123ABC-0-0
                                 | ----    v Sub Requirement 2                       major-123ABC-0-1
                                 | |         [course 1] [course 2] [course three]
    "major-123ABC-0": true,   ---- |     > Requirement B                             major-123ABC-1
    "major-123ABC-0-1": true, ------ --  v Requirement C                             major-123ABC-2
                                     |     [course 1] [course 2]
    "major-123ABC-2": true,   --------   > Requirement D                             major-123ABC-3


                                       Displayed requirements for major 789Z       Their storeKeys
                                         > Requirement A                             major-789Z-0
    "major-789Z-0-0": true    ---------     (hidden but expanded)                    major-789Z-0-0
  }
```

- New behavior resulting from this includes:
  - If you open a group that is nested inside another one (i.e. GE requirements -> the second requirement, called "Select 1 of the following" -> open an inner group), then close and reopen the outer one, the opened state of the inner one is now preserved
  - If you switch tabs between major, minor, and GE requirements, the opened states are all perserved for each, independently
  - The opened states are preserved independently for each major of multiple majors (this should also work for multiple minors in the future with no changes)
  - If a major is added, some groups are expanded, the major is removed, and the major is re-added, the opened states will stay the same (specializations have no impact on this)
- Passing down the new unique group `storeKey`s as props resulted in a lot of code changes in the program requirements list components
  - I would appreciate a review of the code and suggestions on how to make the process more concise/elegant, or whether a completely different system of identifying groups would be better altogether
  - For instance, `CourseRequirement` and `GroupRequirement` use the same code for getting/setting whether they are opened. Should they be refactored in some way?
  - The actual React `key` props are the same as they were, just `storeKey` was added, is this okay?

## Screenshots

Below: what the Major tab looks like just after having switched back from the Minor tab
image

## Test Plan

- Test all new behavior described (beneath the visualization above)
- Test that existing related behavior still works as intended

## Issues

Closes #624

### DIFF
--- a/site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.tsx
+++ b/site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.tsx
@@ -19,6 +19,7 @@ import { setActiveCourse, setActiveCourseLoading, setShowAddCourse } from '../..
 import { useAppDispatch, useAppSelector } from '../../../store/hooks';
 import { Spinner } from 'react-bootstrap';
 import { ProgramRequirement, TypedProgramRequirement } from '@peterportal/types';
+import { setGroupExpanded } from '../../../store/slices/courseRequirementsSlice';
 
 interface CourseTileProps {
   courseID: string;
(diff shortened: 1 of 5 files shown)


site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.tsx
```diff
@@ -183,7 +206,13 @@ const GroupRequirement: FC<GroupRequirementProps> = ({ data, takenCourseIDs }) =
       )}
       {open &&
         data.requirements.map((r, i) => (
-          <ProgramRequirementDisplay key={i} requirement={r} nested takenCourseIDs={takenCourseIDs} />
+          <ProgramRequirementDisplay
+            key={i}
+            storeKey={storeKey + '-' + i}
```
Awesome-E, line 211: minor nit – maybe use a template string?
```suggestion
            storeKey={`${storeKey}-${i}`}
```

Awesome-E (approved):
:shipit:

PR #630, opened 2025-04-07
Programmatically shrink long course names in course requirements tiles

### LINKED ISSUE #591: Long Course Names in Requirement Lists should shrink their text
So they don't wrap or go outside the tile. Example Design:

Image

This can be achieved with a simple function that decreases the font size based on the length of the course ID (when spaces are removed)

Awesome-E:
## Description

This PR makes it so that long course names have a smaller font size so that all course names fit in a course tile instead of wrapping to the next line.

This behavior is not perfect, and if it's serious enough, we can think about scaling based on rendered width should that be necessary. However, I do believe this is good enough at least for the time being.

## Screenshots

Before|After
:-:|:-:
image|image

## Test Plan

- [ ] Go to staging instance
- [ ] Open GE III and ctrl+f for `CRM/LAWH80` and `CHC/LAT164A`. Make sure these don't wrap lines or exhibit any other strange visual behavior.
- [ ] Check across some other required course lists and make sure that course tiles render as you would expect them to with this change.

Do note that text scaling is not perfect, and you might get cases like `CHC/LAT164A` and `CRM/LAWH80` not being scaled to the same width.

## Issues

Closes #591

### DIFF
--- a/site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.scss
+++ b/site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.scss
@@ -80,7 +80,7 @@ $complete-color: green;
   display: flex;
   justify-content: center;
   align-items: center;
-  padding: 5px 10px;
+  padding: 5px 2px;
   background-color: globals.$primary-blue-1;
   color: white;
   font-weight: bold;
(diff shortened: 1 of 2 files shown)


CadenLee2 (approved):
- Tried the courses in the test plan as well as a few other courses and it works well

- In cases with consistently wide letters (like `WWWWWWWWW`) it doesn't work because the sizing is based on the string length, but obviously we won't have courses like that so this is good enough; the longest one I could find was:
image
(which is perfectly fine)

- Code is simple and makes sense (edit: it will break with long names by producting a negative font size which reverts to the default, but again this should never be an issue)

PR #638, opened 2025-04-12
Customized react-select theme, simplified overall color palette

### LINKED ISSUE #600: Degree Requirements selection box theme colors should be more consistent with the rest of the app
Currently in degree requirements, the styles are mostly just the default react-select library styles. we should override them to use our theme colors where applicable and have a more robust color pallet within our global CSS.

charlieweinberger:
The current selection boxes in degree requirements use the default react-select blues. In this PR, I updated those blues to more closely match with the rest of the app. I also heavily modified the current color palette defined in index.css, as well as updating said colors throughout the codebase, with the goal of simplifying and standardizing our colors. I recognize that there are many changes in this PR, some of which are my own opinions, so I welcome feedback.

## Description

- Updated comboboxTheme() in `courseRequrements.ts`, to apply `--blue-primary`, `--blue-secondary`, and `--blue-tertiary` to the selection box theme, for the selection box border, active state, and hover states, respectively.
- Simplified and standardized color palette
    - Moved `ProgramRequirementsList.css`'s `$complete-color` into `index.css` and renamed it `--peterportal-primary-green`
    - Folded `global.css`'s `$primary-blue-1` into `index.css`'s `--peterportal-primary-color-1`
    - Created `--blue-primary`, intended to replace `--peterportal-primary-color-1` and `--zot-blue`
    - Removed the definition of `--peterportal-primary-color-1` in `theme.scss` that overwrote itself to be `#268cd3` in dark mode
    - Created `--blue-secondary` and --blue-tertiary` for the degree requirements selection box theme, which have the same hue and saturation as `--blue-primary` but differing lightnesses (multiples of 16 up/down for light/dark mode).
    - Created `--blue-accent`, intended to replace `--peterportal-primary-color-2` and `--aldrich-blue`
    - Removed `--peterportal-light-gray` and `--peterportal-uci-gold`, since they were unused (`--uci-gold` was previously only used for the text selection highlight color)
    - Folded `--petr-gray` into `--peterportal-mid-gray,` since they're so similar
    - Replaced `rgba(globals.$primary-blue-1, 0.5)` and `rgba(globals.$primary-blue-1, 0.8)` in `ProgramRequirementsList.css` with `--blue-secondary`.
    - Renamed both `--peterportal-primary-color-1` and `--zot-blue` to `--blue-primary`
    - Renamed both `--peterportal-primary-color-2` and `--aldrich-blue` to `--blue-accent`
    - Renamed `--peterportal-secondary-green/yellow/orange/red` to `--green/yellow/orange/red-secondary`
    - Add a dark mode variant for `--green-secondary`, and slightly modified the light mode variant to match.
    - Renamed `--peterportal-mid-gray` to `--mid-gray`
    - Renamed `--peterportal-gray-blue` to `--gray-blue`

### Future Issues

- Combine all custom colors into one file (current defined in `index.css`, `style/theme.scss`, and other random files, such as `pages/Schedule/Schedule.scss` and `pages/RoadmapPage/RoadmapMultiplan.scss`)
- Add dark mode versions for rating colors
- Rename `--mid-gray` to `--gray`, `--gray-blue` to `--light-gray`, and/or `--ring-road-white` to `--white`
- Consider using the dark mode version of `--blue-primary` for dark mode
- Decide how we should handle colors that come from semantic-ui or bootstrap (i.e. aren't defined directly in this codebase), and how that relates to the potential future use of MUI, when merging with AntAlmanac
- Fix current bug where after a specialization is selected, the drop down box still remains blue until the cursor clicks somewhere else

Edit #1: Updated description to match most recent commit.
Edit #2: Updated description to match most recent commit.
Edit #3: Updated description to include the suggestion from Charlie Z as a future issue, and rephrased other future issues.
Edit #4: Updated description to include the addition of a dark mode variant for `--green-secondary` and the modification to it's light mode.

## Screenshots

Before:
Screenshot 2025-04-12 at 3 26 06 AMScreenshot 2025-04-12 at 3 26 35 AM

After:

PeterPortal-#600-Screenshot-light-mode
PeterPortal-#600-Screenshot-dark-mode

## Test Plan

- [x] Combobox
    - [x] Test that all aspects of the degree requirements select dropdown is the correct color, including the outline of the box and hover/active states
    - [x] Test that the same is true for the minor/specialization tabs
    - [x] Test if it is customized on both light and dark mode
    - [x] Test if it looks as expected on mobile
    - [x] Give opinions on if the colors for light and dark mode look good
- [x] Color palette modification
    - [x] Confirm that all instances of all colors were successfully replaced/renamed in the codebase, such that no color is undefined or otherwise defined in such a way that would interfere with the user interface.
    - [x] Give opinions on the shapes of --blue-primary, -secondary, and -tertiary
    - [x] Give opinions on the new names for every color

## Issues

Closes #600

### DIFF
--- a/site/src/helpers/courseRequirements.ts
+++ b/site/src/helpers/courseRequirements.ts
@@ -45,7 +45,7 @@ export const LOADING_COURSE_PLACEHOLDER: CourseGQLData = {
 export const comboboxTheme = (theme: Theme, darkMode: boolean) => {
   const themeCopy = { ...theme, colors: { ...theme.colors } };
 
-  // These need to match the CSS variabels defined in index.css
+  // These need to match the CSS variables defined in index.css
   const cssVars = {
     '--blue-primary': '#2484c6',
     '--blue-secondary': darkMode ? '#185580' : '#5babe1',
--- a/site/src/helpers/courseRequirements.ts
+++ b/site/src/helpers/courseRequirements.ts
@@ -43,23 +43,29 @@ export const LOADING_COURSE_PLACEHOLDER: CourseGQLData = {
 };
 
 export const comboboxTheme = (theme: Theme, darkMode: boolean) => {
-  if (!darkMode) return theme;
-
   const themeCopy = { ...theme, colors: { ...theme.colors } };
-  const neutralIncrements = [0, 5, 10, 20, 30, 40, 50, 60, 70, 80, 90];
-  Object.entries(theme.colors).forEach(([key]) => {
-    if (key.startsWith('neutral')) {
-      const index = neutralIncrements.indexOf(parseInt(key.replace('neutral', '')));
-      const opposite = ('neutral' + neutralIncrements.at(-1 - index)) as keyof Theme['colors'];
-      themeCopy.colors[key as keyof Theme['colors']] = theme.colors[opposite];
-    }
-  });
 
-  themeCopy.colors.danger = theme.colors.dangerLight;
-  themeCopy.colors.dangerLight = theme.colors.danger;
+  // These need to match the CSS variabels defined in index.css
+  const cssVars = {
+    '--blue-primary': '#2484c6',
+    '--blue-secondary': darkMode ? '#185580' : '#5babe1',
+    '--blue-tertiary': darkMode ? '#0b283c' : '#a0ceee',
+  };
 
-  themeCopy.colors.primary = theme.colors.primary75;
-  themeCopy.colors.primary25 = '#343a40';
+  themeCopy.colors.primary = cssVars['--blue-primary']; // box border
+  themeCopy.colors.primary50 = cssVars['--blue-secondary']; // active
+  themeCopy.colors.primary25 = cssVars['--blue-tertiary']; // hover
+
+  if (darkMode) {
+    const neutralIncrements = [0, 5, 10, 20, 30, 40, 50, 60, 70, 80, 90];
+    Object.entries(theme.colors).forEach(([key]) => {
+      if (key.startsWith('neutral')) {
+        const index = neutralIncrements.indexOf(parseInt(key.replace('neutral', '')));
+        const opposite = ('neutral' + neutralIncrements.at(-1 - index)) as keyof Theme['colors'];
+        themeCopy.colors[key as keyof Theme['colors']] = theme.colors[opposite];
+      }
+    });
+  }
 
   return themeCopy;
 };
--- a/site/src/helpers/courseRequirements.ts
+++ b/site/src/helpers/courseRequirements.ts
@@ -59,10 +59,11 @@ export const comboboxTheme = (theme: Theme, darkMode: boolean) => {
   themeCopy.colors.dangerLight = theme.colors.danger;
 
   themeCopy.colors.primary = theme.colors.primary75;
-  themeCopy.colors.primary25 = '#343a40'; // same as bootstrap dark
+  themeCopy.colors.primary25 = '#343a40'; // replace this with var(--bootstrap-dark)
 
   return themeCopy;
 };
+
 /**
  * Groups consectutive single-course requirements into one group requirement where all courses must be completed
  * @param requirements The raw course requirements, as returned from the API
--- a/site/src/helpers/courseRequirements.ts
+++ b/site/src/helpers/courseRequirements.ts
@@ -59,7 +59,7 @@ export const comboboxTheme = (theme: Theme, darkMode: boolean) => {
   themeCopy.colors.dangerLight = theme.colors.danger;
 
   themeCopy.colors.primary = theme.colors.primary75;
-  themeCopy.colors.primary25 = '#343a40'; // replace this with var(--bootstrap-dark)
+  themeCopy.colors.primary25 = '#343a40';
 
   return themeCopy;
 };
--- a/site/src/helpers/courseRequirements.ts
+++ b/site/src/helpers/courseRequirements.ts
@@ -45,16 +45,14 @@ export const LOADING_COURSE_PLACEHOLDER: CourseGQLData = {
 export const comboboxTheme = (theme: Theme, darkMode: boolean) => {
   const themeCopy = { ...theme, colors: { ...theme.colors } };
 
-  // These need to match the CSS variables defined in index.css
-  const cssVars = {
-    '--blue-primary': '#2484c6',
-    '--blue-secondary': darkMode ? '#185580' : '#5babe1',
-    '--blue-tertiary': darkMode ? '#0b283c' : '#a0ceee',
+  const getCssVariable = (variableName: string) => {
+    const bodyStyles = getComputedStyle(document.body);
+    return bodyStyles.getPropertyValue(variableName).trim();
   };
 
-  themeCopy.colors.primary = cssVars['--blue-primary']; // box border
-  themeCopy.colors.primary50 = cssVars['--blue-secondary']; // active
-  themeCopy.colors.primary25 = cssVars['--blue-tertiary']; // hover
+  themeCopy.colors.primary = getCssVariable('--blue-primary'); // box border
+  themeCopy.colors.primary50 = getCssVariable('--blue-secondary'); // active
+  themeCopy.colors.primary25 = getCssVariable('--blue-tertiary'); // hover
 
   if (darkMode) {
     const neutralIncrements = [0, 5, 10, 20, 30, 40, 50, 60, 70, 80, 90];
--- a/site/src/helpers/toastify.ts
+++ b/site/src/helpers/toastify.ts
@@ -8,7 +8,7 @@ export default function spawnToast(text: string, error = false, style?: Record<s
     gravity: 'bottom',
     position: 'right',
     style: {
-      background: error ? '#D22B2B' : 'var(--peterportal-primary-color-1)',
+      background: error ? '#D22B2B' : 'var(--blue-primary)',
       fontWeight: 'bold',
       ...style,
     },
--- a/site/src/index.css
+++ b/site/src/index.css
@@ -42,7 +42,6 @@ code {
 
 @media (prefers-color-scheme: dark) {
   :root {
-    --blue-primary: #268cd3; /*! Note: different from the color above */
     --blue-secondary: #185580;
     --blue-tertiary: #0b283c;
     --rating-4: #3e7f41;
--- a/site/src/index.css
+++ b/site/src/index.css
@@ -40,7 +40,6 @@ code {
 
   --peterportal-gray-blue: #f5f6fc;
   --peterportal-mid-gray: #8d8d8d;
-  --petr-gray: #808080;
   --ring-road-white: #ffffff;
 
   --border-radius: 8px;
--- a/site/src/index.css
+++ b/site/src/index.css
@@ -18,34 +18,68 @@ code {
   height: 100vh;
 }
 
+/*
+
+Notes for draft:
+
+Already done in this PR:
+- Moved ProgramRequirementsList.css's "$complete-color" into index.css, and renamed it to "--peterportal-primary-green"
+- Folded global.scss's "$primary-blue-1" into --peterportal-primary-color-1 in index.css
+- Created --blue-primary, intended to replace --peterportal-primary-color-1 and --zot-blue
+- Created --blue-secondary and --blue-tertiary for the degree requirements selection box theme
+- Created --blue-accent, intended to replace --peterportal-primary-color-2 and --aldrich-blue
+- Removed --peterportal-light-gray and --peterportal-uci-gold, since they were unused (--uci-gold was previously only used for the text selection highlight color)
+
+Still to do in this PR:
+- Create an inverted version of --peterportal-secondary-green for dark mode (take opposide HSL)
+- Update comboboxTheme() in courseRequrements.ts, to apply new blues to selection box theme. Make default theme "baseTheme" for light mode with theme colors from index.css as 25%, 50%, and 100% (and 75% if necessary). For dark mode, override with dark mode colors (does this automatically happen?)
+- In ProgramRequirementsList.css, replace rgba(var()) with CSS variables, since rgba() doesn't work with CSS variables 
+- In ProgramRequirementsList.css, overwrite dark mode for --peterportal-secondary-green
+
+Future issues:
+- Replace both --peterportal-primary-color-1 and --zot-blue with --blue-primary in the entire codebase
+- Replace both --peterportal-primary-color-2 and --aldrich-blue with --blue-accent in the entire codebase
+- Rename --peterportal-secondary-green/yellow/orange/red to --rating-1/2/3/4, and update the entire codebase to match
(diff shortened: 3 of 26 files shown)


site/src/helpers/courseRequirements.ts
```diff
@@ -43,26 +43,33 @@ export const LOADING_COURSE_PLACEHOLDER: CourseGQLData = {
...
+    '--blue-primary': '#2484c6',
+    '--blue-secondary': darkMode ? '#185580' : '#5babe1',
+    '--blue-tertiary': darkMode ? '#0b283c' : '#a0ceee',
+  };
 
-  themeCopy.colors.primary = theme.colors.primary75;
-  themeCopy.colors.primary25 = '#343a40'; // same as bootstrap dark
+  themeCopy.colors.primary = cssVars['--blue-primary']; // box border
+  themeCopy.colors.primary50 = cssVars['--blue-secondary']; // active
+  themeCopy.colors.primary25 = cssVars['--blue-tertiary']; // hover
```
Awesome-E, line 57: is it not possible to use CSS Variables when setting properties in `themeCopy`? At the very least, we should programmatically get the color from the body instead of hard-coding them here.

site/src/index.css
```diff
@@ -19,23 +19,35 @@ code {
...
+  --blue-primary: #2484c6;
+  --blue-secondary: #5babe1;
+  --blue-tertiary: #a0ceee;
+  --blue-accent: #74d1f6;
+
+  --blue-primary-50-percent: #2484c680;
+  --blue-primary-80-percent: #2484c6cc;
+
+  --green-primary: green;
+  --rating-4: #81c284;
```
Awesome-E, line 31: I think you should stick with one naming convention. I know we discussed both options, but I could have specified that it'd be better to be consistent between these
```suggestion
  --green-primary: green;
  --green-secondary: #81c284;
```
Because the rating-4 color is essentially a secondary green, we should name it as such if the primary one is going to be called `green-primary`

Awesome-E: You would have to change this for other ratings as well, of course.
And for now, I do think it's fine if you change `--rating-3` to be `--yellow-secondary` and not have a `--yellow-primary`.

site/src/index.css
```diff
@@ -19,23 +19,35 @@ code {
...
-  --peterportal-gray-blue: #f5f6fc;
-  --zot-blue: var(--peterportal-primary-color-1);
-  --aldrich-blue: var(--peterportal-primary-color-2);
+  --blue-primary: #2484c6;
+  --blue-secondary: #5babe1;
+  --blue-tertiary: #a0ceee;
+  --blue-accent: #74d1f6;
+
+  --blue-primary-50-percent: #2484c680;
+  --blue-primary-80-percent: #2484c6cc;
```
Awesome-E, line 28: I don't defining specific opacities is not the right move. It makes more sense to use either a sass variable to support the rgba(hex, alpha) or just use the secondary color instead of using a different opacity.

In our meeting, we decided to use `--blue-secondary` to replace the 50% opacity of the primary color

site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.scss
```diff
@@ -1,16 +1,14 @@
...
-
 .program-requirements {
   display: flex;
   flex-direction: column;
   gap: 16px;
 }
 
 .group-requirement {
-  border: 2px solid rgba(globals.$primary-blue-1, 0.5);
+  border: 2px solid var(--blue-primary-50-percent);
```
Awesome-E, line 11: See comment on opacity vs using decreasing importance (50% opacity == primary → secondary)

Awesome-E (changes requested):
Getting close, but I noticed a few things.

site/src/helpers/courseRequirements.ts
charlieweinberger: I wasn't able to easily get CSS variables to work properly in both light and dark mode. The code below does properly use the CSS variable names, but it matches the system theme, not the theme of peterportal (as defined in the top-right dropdown). I'm not sure if this is intentional, and I'm not sure how to fix it for just this function.

```
  const getCssVariable = (variableName: string) => {
    const rootStyles = getComputedStyle(document.documentElement);
    return rootStyles.getPropertyValue(variableName).trim();
  }
  
  themeCopy.colors.primary = getCssVariable('--blue-primary'); // box border
  themeCopy.colors.primary50 = getCssVariable('--blue-secondary'); // active
  themeCopy.colors.primary25 = getCssVariable('--blue-tertiary'); // hover
  ```

Comment (charlieweinberger) [reply site/src/helpers/courseRequirements.ts:57]: This does seem to be an issue elsewhere in the code. `--green-secondary` has both a light more and dark mode variant, as defined in `index.css` and it too is matching the system theme, not the peterportal theme. I don't quite understand how our app differentiates between light and dark mode - what is going on, and how can we fix this discrepancy?

File: site/src/index.css (hunk 3 above)
Comment (charlieweinberger) [reply site/src/index.css:28]: I agree that defining css variables for specific opacities isn't great. I'll fix that right now.

However, I don't think `--blue-secondary` is a good replacement for 50% of `--blue-primary`, since they're rather different shades of blue, and the secondary color just does not look good IMO as the border of the dropdowns.

50% of primary blue is on the left, secondary blue is on the right:
Screenshot 2025-04-13 at 12 25 42 AMScreenshot 2025-04-13 at 12 25 49 AM

File: site/src/index.css (hunk 2 above)
Comment (charlieweinberger) [reply site/src/index.css:31]: Agreed, fixed

File: site/src/index.css (hunk 3 above)
Comment (charlieweinberger) [reply site/src/index.css:28]: Just found and implemented a way to use the css variable `--blue-primary` directly at 50% or 80% opacity: with `color-mix()`. Seems like it's a newer CSS feature, but still supported on all major browsers. Instead of `rgba(globals.$primary-blue-1, 0.5)` or `var(--blue-primary-50-percent)`, we can do `color-mix(in srgb, var(--blue-primary) 50%, transparent);`, and it'll work the same, using the same primary blue variable as everywhere else and without needing a SCSS variable. I prefer this method over a use SCSS variable and `rgba(var())` because this allows us to keep the color defined in just `index.css`.

File: site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.scss (hunk 4 above)
Comment (charlieweinberger) [reply site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.scss:11]: See comments above.

**Comment by zcharz** [comment]
maybe a minor issue, but in my opinion the color names would be better off as --primary-blue rather than --blue-primary for example, and follow the same format of --usage-color rather than --color-usage. this would allow future design changes to more transparently modify the colors

**Comment by zcharz** [comment]
after a specialization is selected/clicked, the drop down box still remains blue until the cursor clicks somewhere else

**Comment by charlieweinberger** [comment]
> maybe a minor issue, but in my opinion the color names would be better off as --primary-blue rather than --blue-primary for example, and follow the same format of --usage-color rather than --color-usage. this would allow future design changes to more transparently modify the colors

I prefer --blue-primary, and as far as I can tell, --color-usage is the preferred order by most css frameworks.

**Comment by charlieweinberger** [comment]
> opinions on the colors: darker/more "colored" colors feels a bit muted and gloomy, in dark mode the darker blue doesn't contrast enough with the black text

I can see how this could be the case. Can you provide screenshots of exactly which combinations and color & text look off? Also, I don't think that the dark blues should ever have black text... can you also send a screenshot of where that happens? That might be a bug related to whether colors use the light vs. dark mode variants, which I mentioned in one comment response to one of Ethan's suggestions.

**Comment by charlieweinberger** [comment]
> after a specialization is selected/clicked, the drop down box still remains blue until the cursor clicks somewhere else. i presume this isn't intended

I agree that that doesn't seem like it should happen, so I'd call it a bug, and it should be fixed. However, I think that's slightly out of the scope of this PR, since this PR is strictly related to the colors of the select boxes, not when they're actively selected.

File: site/src/helpers/toastify.ts [hunk 5]
```diff
@@ -8,7 +8,7 @@ export default function spawnToast(text: string, error = false, style?: Record<s
     gravity: 'bottom',
     position: 'right',
     style: {
-      background: error ? '#D22B2B' : 'var(--peterportal-primary-color-1)',
+      background: error ? '#D22B2B' : 'var(--blue-primary)',
```
CadenLee2, line 11: Is there a reason the error color is specified here without a variable? The color `#D22B2B` is a bright red much like the already-existing `var(--warning-red)` (light mode: `#ce0000`, dark mode: `red`). Considering that the error toast shouldn't appear that often anyway, I think it could be replaced for consistency?

Here's what it would look like if it was `var(--warning-red)`:
image

CadenLee2:
I also find it uncomfortable that in dark mode the text for a selected spec is black over the blue background, while the other places with the same/similar blue background (such as a button or a selected tab, see above in the screenshot) have white text. In light mode it looks good
image

site/src/index.css
Awesome-E: > However, I don't think --blue-secondary is a good replacement for 50% of --blue-primary, since they're rather different shades of blue, and the secondary color just does not look good IMO as the border of the dropdowns.

Secondary in dark mode should be overridden to a darker shade of blue

> we can do `color-mix(in srgb, var(--blue-primary) 50%, transparent);`

if that works, I like that as well. In favor of this or using secondary (which should be overridden in dark mode regardless) over SCSS variables.

site/src/helpers/courseRequirements.ts
Awesome-E: > but it matches the system theme, not the theme of peterportal

that's a good point. I believe that the `<body>` element gets an attribute `data-theme="dark"` when in dark mode, so you can use that and get the computer style of the `body` element. Then, you can get the values that way.

site/src/index.css
```diff
@@ -19,23 +19,32 @@ code {
...
 }
 
+@media (prefers-color-scheme: dark) {
+  :root {
+    --blue-secondary: #185580;
+    --blue-tertiary: #0b283c;
+    --green-secondary: #3e7f41;
+  }
+}
+
```
Awesome-E, line 47: In response to the other comment about matching peterportal theme instead of system theme
```suggestion
body[data-theme="dark"] {
  --blue-secondary: #185580;
  --blue-tertiary: #0b283c;
  --green-secondary: #3e7f41;
}
```

Note that this would override it on the body rather than the `:root`, but that should not be a big deal in this case. However, it is something that should be kept in mind.

Awesome-E (commented):
Waiting for responses to other reviews, then I will check code again and test on staging. Otherwise, the code itself looks good to me.

site/src/index.css
charlieweinberger: You're right, I realized that the image above shows --blue-secondary's light mode variant, possibly because of the issue with light vs. dark mode we discussed above. Once I fix that issue, I'll consider replacing this 50% opacity color with secondary.

site/src/helpers/courseRequirements.ts
charlieweinberger: Sounds good, I'll try implementing this soon (probably after my 162 exam tmrw)

charlieweinberger:
> I also find it uncomfortable that in dark mode the text for a selected spec is black over the blue background, while the other places with the same/similar blue background (such as a button or a selected tab, see above in the screenshot) have white text. In light mode it looks good image

Charlie (the other one) also mentioned this, and I agree that the black text in dark mode on the blue background doesn't look great. I'll look into this.

site/src/index.css
charlieweinberger: Sounds good. Thoughts on leaving a comment within `index.css`, saying `/* Keep in mind that these colors are defined in :root but overridden in body */`, right above like 40? Or should we just remember ourselves and leave it out to keep the code cleaner

site/src/index.css
charlieweinberger: Just replaced both instances of `color-mix` with `--blue-secondary`, since the secondary color is almost exactly the same, so there's no need for the added complexity of `color-mix`.

site/src/helpers/courseRequirements.ts
charlieweinberger: nvm, just implemented it rn, color override seems to work now

charlieweinberger:
As far as I know, all bug fixes/code improvements have been implemented. The only remaining question is, are the dark mode variants of `--blue-primary`, `--blue-secondary`, and `--blue-tertiary` too dark/not contrasting enough with the selection items in the dropdown menus on dark mode? I somewhat agree, but also think it's not necessary to improve. @Awesome-E what do you think?

Awesome-E:
> Charlie (the other one) also mentioned this, and I agree that the black text in dark mode on the blue background doesn't look great. I'll look into this.

You can forgo that for this PR since Multiple Minors will effectively eliminate this state.

Awesome-E:
> The only remaining question is, are the dark mode variants of --blue-primary, --blue-secondary, and --blue-tertiary too dark/not contrasting enough with the selection items in the dropdown menus on dark mode?

I think the contrast is low, but to me the issue also has to do with the text box not having enough contrast from the background. I think that would be something for a different PR though

site/src/index.css
```diff
@@ -19,23 +19,31 @@ code {
...
+  --mid-gray: #8d8d8d;
+  --gray-blue: #f5f6fc;
   --ring-road-white: #ffffff;
-  --petr-gray: #808080;
+
   --border-radius: 8px;
 }
 
+/* Keep in mind that these colors are defined in :root but overridden in body */
```
Awesome-E, line 40:
```suggestion
/* User theme is not accessible to :root, so we override them in body */
```

site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.scss
```diff
@@ -80,8 +78,8 @@ $complete-color: green;
   display: flex;
   justify-content: center;
   align-items: center;
-  padding: 5px 2px;
-  background-color: globals.$primary-blue-1;
+  padding: 5px 10px;
```
Awesome-E, line 81: I assume this happened because of a merge conflict, but please revert this line

Awesome-E: "this line" refers to just the padding. for some reason GH did not highlight correctly

zcharz (approved):

charlieweinberger:
@Awesome-E @CadenLee2 @zcharz Just putting this in writing here: the only question I still have for this PR is what the updated color of `--green-secondary` should be in dark mode. It's currently `#3e7f41`, but I'm thinking of changing it to either `#1c4a1c` or `#206020`. What you you guys think?

`#3e7f41`
Screenshot 2025-04-14 at 4 21 26 PM

`#1c4a1c`
Screenshot 2025-04-14 at 4 22 12 PM

`#206020`
Screenshot 2025-04-14 at 4 22 31 PM

charlieweinberger:
Final change: updated the light and dark mode variants of `--green-secondary` to `#87c587` and `#295629`, respectively.

zcharz (approved):

PR #643, opened 2025-04-13
fix missing prerequisites on major/minor/ge tabs on roadmap sidebar course hover

### LINKED ISSUE #614: Program Requirements (Major, Minor, GE) don't show missing prerequisite text
On the search sidebar, hovering a course for info or tapping to add it should show missing prerequisites. Currently, it does not.

Device|Expected|Actual
:-:|:-:|:-:
Desktop|Image|Image
Mobile|Image|Image

zcharz:
## Description

Mobile: calculate missing prerequisites and update information using dispatch() within insertCourseOnClick()

PC: retrieve completed course list and calculate missing prerequisites within the CoursePopover component, such that the information is only retrieved on hover

Notably, the prerequisite update for mobile is within ProgramRequirementsList itself, while for PC the logic is added to the popover component. This is because for the PC component, adding the logic in ProgramRequirementsList by feeding the information into CourseNameAndInfo would require the roadmap state to be read for every coursetile before they are hovered on.

Also moved logic of retrieving roadmap data and transfer courses to calculate all cleared courses to a hook, made helper function which calculates the missing prerequisites of a course given the course and set of cleared courses.

## Screenshots

BEFORE

image

image

AFTER

image

image

## Test Plan

Check for correct missing prerequisites listed for when no courses are in the roadmap, when some prerequisites are fulfilled, and when all prerequisites are completed.

## Issues

Closes #614

### DIFF
--- a/site/src/component/CoursePopover/CoursePopover.tsx
+++ b/site/src/component/CoursePopover/CoursePopover.tsx
@@ -12,6 +12,8 @@ import {
   PreviousOfferingsRow,
 } from '../CourseInfo/CourseInfo';
 import { Spinner } from 'react-bootstrap';
+import { useClearedCourses } from '../../hooks/planner';
+import { getMissingPrerequisites } from '../../helpers/planner';
 
 interface CoursePopoverProps {
   course: CourseGQLData | string;
@@ -25,6 +27,11 @@ const CoursePopover: FC<CoursePopoverProps> = ({ course, interactive = true, req
       <Spinner animation="border" />
     </div>
   );
+
+  const clearedCourses = useClearedCourses();
+  if (typeof course !== 'string') {
+    requiredCourses = getMissingPrerequisites(clearedCourses, course);
+  }
 
   if (typeof course !== 'string') {
     const { department, courseNumber, minUnits, maxUnits } = course;
--- a/site/src/pages/RoadmapPage/Course.tsx
+++ b/site/src/pages/RoadmapPage/Course.tsx
@@ -29,6 +29,7 @@ interface CourseNameAndInfoProps {
   requiredCourses?: string[];
   /** Whether to always collapse whitespace in the course name */
   alwaysCollapse?: boolean;
+  // change prev to isTile
 }
 export const CourseNameAndInfo: React.FC<CourseNameAndInfoProps> = (props) => {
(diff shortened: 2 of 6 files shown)


site/src/component/CoursePopover/CoursePopover.tsx
```diff
@@ -26,6 +28,11 @@ const CoursePopover: FC<CoursePopoverProps> = ({ course, interactive = true, req
     </div>
   );
 
+  const clearedCourses = useClearedCourses();
+  if (typeof course !== 'string') {
+    requiredCourses = getMissingPrerequisites(clearedCourses, course);
+  }
```
Awesome-E, line 34: same if condition as the next one; combine the two.

Additionally, if we're always going override the `requiredCourses` array when the data is done loading, then I think we should just change `requiredCourses` to be a local variable instead of a prop passed in.

site/src/pages/RoadmapPage/Course.tsx
```diff
@@ -29,6 +29,7 @@ interface CourseNameAndInfoProps {
   requiredCourses?: string[];
   /** Whether to always collapse whitespace in the course name */
   alwaysCollapse?: boolean;
+  // change prev to isTile
```
Awesome-E, line 32: Hanging comment; decide whether you want to make the change, and then remove this comment.

site/src/pages/RoadmapPage/Course.tsx
```diff
@@ -104,6 +105,7 @@ export const CourseNameAndInfo: React.FC<CourseNameAndInfoProps> = (props) => {
         <a className="name" href={courseRoute} target="_blank" rel="noopener noreferrer" onClick={handleLinkClick}>
           {courseID}
         </a>
+        {/* use isTile here to NOT show exclaim on non-search tabs */}
```
Awesome-E, line 108: hanging comment

site/src/pages/RoadmapPage/Course.tsx
Awesome-E, line 1: Did you mean to push additional changes to this file? The logic seems like it might work just fine, but there are some comments that suggest you would still do something else with it.

Awesome-E (changes requested):
Good work so far! Just a few questions regarding the code style, but you're pretty close!

Issue #621, opened 2025-03-22
User-selected minors should be saved to the user's account


Awesome-E:
If the user selects a minor, currently it only shows for the current session (reloading will make it go away). It should be saved to the user's account when they add/remove a minor. This issue should be closed in conjunction with #620

ellisl10:
Closed by 644

PR #644, opened 2025-04-13
Can now save multiple minors to account

### LINKED ISSUE #620: User Should be Able to Select Multiple Minors
Currently, the user can select multiple majors. However, it is also possible for someone to have more than one minor in their degree, and our interface does not allow for that.

Major Selection|Minor Selection
:-:|:-:
Image|Image

We should make it so that a user can select multiple minors. It should function almost exactly the same as it does for Majors. Work should be modeled after the `<MajorRequiredCourseList/>` and `<SingleMajorCourseList/>` components, and the header used in the `<SingleMajorCourseList/>` component (lines 141-144) should be extracted to its own helper component that is used for both majors and minors.

Additionally, we can improve the naming of the components:
- `MajorRequiredCourseList` → `MajorSelector`
- `SingleMajorCourseList` → `MajorCourseList`
- `MinorRequiredCourseList` → `MinorSelector`
- _create a new component_ called `MinorCourseList`

All logic related to only one minor, i.e. fetching requirements, should be done inside the helper component `MinorCourseList`, and the parent should only handle selection of minors and rendering components to display the contents of each minor. For storing the data in the global state, please refer to existing logic in the `MajorRequiredCourseList`/`MajorSelector` component, the `SingleMajorCourseList`/`MajorCourseList` component, and the `courseRequirementsSlice.ts`.

Similar to how majors did it, we should replace `setMinor` with `addMinor` and `removeMinor` (thankfully, there is no `setSpecialization` to deal with :D)

ellisl10:
Save multiple minors to account

## Description

Created new component MinorCourseList and renamed MinorRequiredCourseList to MinorSelector (renamed MajorsCourseList and MajorSelector similarly). The selector handles selection of minors while MinorCourseList actually fetches requirements.

Additionally, created new functions to save minors to account.

## Screenshots

Before 
Screenshot 2025-04-13 at 1 36 29 AM

After
Screenshot 2025-04-13 at 1 35 09 AM

## Test Plan

Verify that minors persist between roadmaps and after refreshing the page.
## Issues

Closes #620 and #621


charlieweinberger:
Selecting multiple minors works well, as does saving minors to my account so they don't get deleted when I reload.

However, I notice that every time I add or delete a minor, the select box becomes grayed-out and shows loading dots for a split second. Is this intentional? I don't notice this for majors, only for minors.

Awesome-E:
- Not sure why deployment isn't triggering
- Based on the branch conflicts, it looks like you might want to `git merge main MultipleMinors` and see if that fixes any issues that Charlie pointed out.

api/src/controllers/programs.ts
```diff
@@ -39,6 +39,11 @@ const zodMajorSpecPairSchema = z.object({
   ),
 });
 
+const MinorProgramSchema = z.object({
```
Awesome-E, line 42: Naming should be consistent with the Majors/Spec pairs

```suggestion
const zodMinorProgramSchema = z.object({
```

Alternatively, you can change `zodMajorSpecPairSchema` to be `majorSpecPairSchema` and make the first letter of `MinorProgramSchema` lowercase.

api/src/controllers/programs.ts
```diff
@@ -99,6 +117,13 @@ const programsRouter = router({
     if (rowsToInsert.length) await db.insert(plannerMajor).values(rowsToInsert);
   }),
   /** @todo add `setPlannerMinor` (or similarly named) operation for updating a minor */
+  saveSelectedMinor: publicProcedure.input(MinorProgramSchema).mutation(async ({ input }) => {
+    const { plannerId, minorIds } = input;
+    await db.delete(plannerMinor).where(eq(plannerMinor.plannerId, plannerId));
+
+    const rowsToInsert = minorIds.map((minorId) => ({ plannerId, minorId }));
+    if (rowsToInsert.length) await db.insert(plannerMinor).values(rowsToInsert);
```
Awesome-E, line 125: This looks like it'd work, but for multiple, we should add/remove individual rows instead of overriding every row every time. Since it's also the case for majors, let's file a new issue for this.

site/src/pages/RoadmapPage/sidebar/MajorSelector.tsx
Awesome-E, line 1: This is a good file name update

site/src/store/slices/courseRequirementsSlice.ts
```diff
@@ -11,15 +11,20 @@ export interface MajorWithSpecialization {
 
 type ExpandedGroupsList = { [key: string]: boolean | undefined };
 
+export interface minorRequirements {
```
Awesome-E, line 14: Interface should be `PascalCase`, not `camelCase`
```suggestion
export interface MinorRequirements {
```

site/src/pages/RoadmapPage/sidebar/MinorSelector.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const isLoggedIn = useIsLoggedIn();
+  const minors = useAppSelector((state) => state.courseRequirements.minorList);
+  const selectedMinors = useAppSelector((state) => state.courseRequirements.selectedMinors);
+
+  const plans = useAppSelector((state) => state.roadmap.plans);
+  const planIndex = useAppSelector((state) => state.roadmap.currentPlanIndex);
+  const activePlanID = plans[planIndex].id;
+
+  const [minorsLoading, setMinorsLoading] = useState(false);
+  // const [resultsLoading, setResultsLoading] = useState(false);
```
Awesome-E, line 33:
```suggestion
```

Awesome-E (changes requested):
Mostly working! Tiny code changes, and also for some reason it looks like if you switch tabs it reloads Minor-required courses every time instead of reading from the cache despite the minor itself being saved.

CadenLee2 (commented):
- Tested multiple minors when logged in and logged out
- Code changes: I agree with Ethan's comments

Awesome-E (approved):

Awesome-E:
💪

PR #649, opened 2025-04-16
Update DB schema for transfer credits

CadenLee2:
Update the database schema to support the transfer credits enhancements by better organizing transferred items

## Description

Updating DB schema (`schema.ts`) by:
- Renaming the transferred course table to transferred misc (for miscellaneous transfers which could not be converted to an AP exam, valid course, or GE)
- Creating a transferred AP table
- Creating a transferred GE table
- Creating a transferred course table (different table from the original transferred course table that was renamed, since this new one has more constraints)

⚠️ The migration process has not yet occurred (other than testing on a temporary database) and this PR **should not be merged until we are ready**, in order to prevent issues (probably Friday). **Do not deploy to the staging instance yet**.

Also note that an incorrectly capitalized `planner_Id` was fixed to `planner_id`

## Test Plan

Check that the new schema (`schema.ts`) looks good

## Issues

Does not close the transfer credits data migration issue because a script is needed to migrate all the data fully
Closes #


charlieweinberger (approved):
I can't see anything particularly wrong about this, although I also can't say I know much about how our database works in practice, so I can't help much in this review (until I learn more, which can happen soon, but just hasn't yet).

Issue #628, opened 2025-04-04
AP Credits that clear GEs or major requirements are not clearing them


ningkaiyang:
Even though I have entered my AP credits correctly exactly as shown on my transcript:

<img width="495" alt="Image" src="https://github.com/user-attachments/assets/70a0d58c-9931-4156-b920-fd1c917f0dc8" />

They aren't clearing any GEs or equivalent courses, which is not a huge deal but feels like a relatively doable thing to implement.

<img width="1256" alt="Image" src="https://github.com/user-attachments/assets/28abfc18-bc8c-4459-bdad-8155d4ac15eb" />

<img width="339" alt="Image" src="https://github.com/user-attachments/assets/71c80242-d7dd-4ca0-8bb0-f9ee4978e8b0" />

<img width="1263" alt="Image" src="https://github.com/user-attachments/assets/35f60a13-8e8c-4080-956a-79bc6db83085" />

<img width="344" alt="Image" src="https://github.com/user-attachments/assets/746f2762-3f53-4dab-88bf-85f625aed015" />

<img width="1053" alt="Image" src="https://github.com/user-attachments/assets/bd5d4198-c3aa-40de-a701-c21714901922" />

<img width="340" alt="Image" src="https://github.com/user-attachments/assets/8388dbce-a867-4444-8bf4-7c3d649feef2" />

Awesome-E:
This is planned to be fixed with a redesigned "Transfer Credits" experience in the next couple of months. Currently, like you stated, AP Exams only clear course prerequisites and not GE or Major Requirements.

Issue #550, opened 2025-01-10
Import AP exams from Zot4Plan when importing 4-year plans


CadenLee2:
Zot4Plan has an AP exam feature that allows users to list AP exams and their corresponding units (similar to PeterPortal's Transfer Credits system).
- When users import their schedule from Zot4Plan (see #547), the AP exams in their Zot4Plan schedule should also be imported to the Transfer Credits list
- The added exams should have the correct names (ex. the "AP Computer Science A" exam, represented by a certain ID in Zot4Plan, should be converted to "AP COMP SCI A") so that they work correctly for prerequisite checks
- The added exams should have the same unit counts as they did in the Zot4Plan schedule
- In the modal for importing, probably indicate to users that any AP exams will also be imported (because unlike with the main roadmap, the fact that the AP exams/Transfer Credits has updated is not obvious at first glance)

PR #653, opened 2025-04-18
Redesign the "Transfer Credits" Experience

### LINKED ISSUE #635: Transferred Credits Data Conversion Pipeline
Because we are separating transferred credits into different sections, the data needs to be stored differently.

First, we need new tables in the database for transferred courses, transferred AP Exams, and transferred GEs. Second, user data needs to be converted from the old format into the new format automatically. We can have this as a user-triggered action, or we can run the script automatically on all user data.

For every entry in the current user transfers:
1. We will want to check if it is an AP Exam. We can probably use a naïve approach like simply checking whether the transferable item starts with `"AP "` (with the space). If so, do a lookup against the response from [Anteater API's AP Exam Names](https://anteaterapi.com/reference#tag/ap-exams/GET/v2/rest/apExams) and assume the minimum score that gets the same number of units.
2. Do a course lookup on the rest of the items. For each one that is a course, add it with the user's original inputted number of units rather than the number of units associated with the course.
3. Don't do anything with items that are neither a course nor AP exam. We will probably have a way to show unsuccessful automatic conversions and display them so that the user can still manually add them then delete the unconverted item.

This issue may require more than 1 PR to close.

Awesome-E:
## Description

This PR is going to be merged once all sections of the redesigned Transfer Credits menu are complete.

## Screenshots

image

## Test Plan

- [ ] Verify that all changes work on staging xD

## Issues

Closes #635, #550, #628


site/src/pages/RoadmapPage/transfers/GESection.tsx
```diff
@@ -0,0 +1,144 @@
...
+}
+
+const GEInput: FC<GEInputProps> = ({ value, handleUpdate, valueType }) => {
+  const handleKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.key === 'Enter') (event.target as HTMLInputElement).blur();
+  };
+
+  const onBlur = (e: React.FocusEvent<HTMLInputElement>) => {
+    if (e.target.value === value.toString()) return;
+    if (isNaN(e.target.valueAsNumber) || e.target.valueAsNumber < 0) {
```
charlieweinberger, line 37:
```suggestion
    if (isNaN(e.target.valueAsNumber) || e.target.valueAsNumber < 0 || (valueType === 'numberOfCourses' && !Number.isInteger(e.target.valueAsNumber))) {
```

Awesome-E: Put up a PR for this

Awesome-E: Applied fix in commit https://github.com/icssc/peterportal-client/pull/653/commits/77ff895c1931feead2f2e6cc195d8825e070e270

ecxyzzy (commented):
chat this is fire. fantastic job to everyone involved

--

One small UX snag I noted is that in the transferred courses search bar, there's a dropdown icon that is highlighted when hovering over it. The issue is that there is no actual dropdown until a search term is provided, and it is the inputting of the search term that causes the dropdown to be displayed. This feels a bit semantically icky to me, but I'm not sure if there's a good alternative.

Also slight typography overrun in GEs:
image

ecxyzzy:
Current data points on the typography overrun:
Windows Firefox (Eddy): ❌
Windows Firefox (Ethan): ✅
Windows Chrome (Eddy): ❌
Windows Chrome (Ethan): ❌
Linux Brave (Dante): ❌
Linux Firefox (Eddy): ✅

CadenLee2 (commented):
Tested staging, looks good overall.

Details:
- It seems they recently took down the placement page on the Testing Center website, so the credit by exam link we included now directs to [a "No Results Found" page](https://testingcenter.uci.edu/programs/placement-testing/). Maybe we just make the link go straight to <https://testingcenter.uci.edu/> since the specific page is gone?
- I agree that the course search dropdown looks a bit weird since it does nothing when clicked at first, but that's probably not a critical change right now

charlieweinberger:
Looks good IMO! Other than the credit by exam link issue that Caden mentioned, I don't see anything that absolutely needs to be fixed before pushing this PR. Small improvements can come later, but for now this is great!

Awesome-E:
> Maybe we just make the link go straight to https://testingcenter.uci.edu/ since the specific page is gone?

sounds good, made that change. it is kinda weird that they now have that same link on the homepage that is broken, but that is no longer our concern

PR #654, opened 2025-04-18
recolor roadmap course missing prerequisite color

### LINKED ISSUE #640: Make coloring consistent for courses with missing prerequisites
Currently in search, it shows a red warning icon, and in the roadmap, the whole name and the unit count are in orange

Image Image

We should make the two consistent, most likely by making just the course **name** and the **warning icon** red, and leave the units unchanged in both cases.

zcharz:
## Description
currently, there are inconsistencies with the roadmap and search bar where if a course has incomplete prerequisites, on the search bar the name and units are blue, but the icon is red

on the roadmap, however, the name, icon, and units are all orange

that is, the colors are different, and which elements are colored are also different

changed where the color is applied and the logic for applying color on missing prerequisites, now the units is always displayed as blue, the course name and icon are red

changing the logic had the side effect of fixing another inconsistency where when dragging a course which has incomplete prerequisites, the course name and icon would appear as blue while being dragged. now, the course title and icon will be colored while being dragged as well

## Screenshots
before:
image
image

after:
image
image
image

## Test Plan
- [ ] course name + icon red
- [ ] trashcan red only on hover
- [ ] consistent on sidebar + roadmap

## Issues
Closes #640


Awesome-E:
Can you make the border of a dragged course the normal color instead of `orange` as well?

site/src/pages/RoadmapPage/Course.scss
```diff
@@ -72,7 +76,7 @@
 }
 
 .course-delete-icon {
-  color: red;
+  color: #ce0000;
```
Awesome-E, line 79: Is there a reason this was changed from `red`? I see you added something else with that same color instead of keeping the `red`

site/src/pages/RoadmapPage/Course.scss
```diff
@@ -38,10 +38,14 @@
   }
 }
 
-.invalid {
+.invalid-shadow {
   box-shadow: 0px 0px 4px orange;
```
Awesome-E, line 42: Change the shadow to the same red as well

site/src/pages/RoadmapPage/Course.scss
```diff
@@ -38,10 +38,14 @@
...
 }
 
+.invalid-name {
+  color: #ce0000;
```
charlieweinberger, line 46: Replace this with `color: var(--warning-red);`, to use the global warning color. For reference, `--warning-red`, defined in `src/style/theme.scss`, is `#ce0000` in light mode and `red` in dark mode. Regardless of whether or not these shades of red look great, this particular warning should use the global warning color nonetheless, and the global warning color can be modified if necessary.

site/src/pages/RoadmapPage/Course.scss
charlieweinberger, line 79: Replace this with `color: var(--warning-red);`, to use the global warning color. See my comment on line 46 for an explanation.

charlieweinberger: I'd recommend replacing this line with `color: var(--warning-red);`, to use the global warning color. See my comment on line 46 for my reasoning.

site/src/pages/RoadmapPage/Course.scss
charlieweinberger: I agree that the shadow should be red, specifically `--warning-red;`.

charlieweinberger (changes requested):
See my line-specific code comments for requested changes, all focusing on the use of ` --warning-red`. For reference, `--warning-red` is the global warning color, defined in `src/style/theme.scss`, and is `#ce0000` in light mode and `red` in dark mode. These prerequisite warnings should use the global warning color to match the overall theme/other warnings. If you have any concerns with those warning shades of red, they should (at least in my opinion) be modified globally, not locally for this warning.

Also, in my opinion, the unit count in courses with a warning should also be red, because when just the course title is red, there isn't enough difference between it and courses without warnings. However, if you do this, but still want to be consistent, then you have to consider changing the units color in the sidebar as well. That might also raise the question about changing the color of the border as well of the course, in the sidebar and/or in the roadmap. That's a lot to think about for not THAT important of a feature, so its probably best to just stick with just the name being red and everything else being blue.

Screenshot 2025-04-22 at 12 11 41 PMScreenshot 2025-04-22 at 12 11 21 PM

charlieweinberger:
> Can you make the border of a dragged course the normal blue color instead of `orange` as well? (You may want to `git merge main consistent-missing-prerequisites` first)

You mention here making the border blue, but in your review of line 42 you requested the border to be red. Which of the two do you prefer?

charlieweinberger:
Just created `--red-primary` in`src/index.css`, and replaced all code instances of either `red`, `#ce0000`, or `--warning-red` with `--red-primary`.

charlieweinberger (approved):
Just created `--red-primary` in`src/index.css`, and replaced all code instances of either `red`, `#ce0000`, or `--warning-red` with `--red-primary`. @Awesome-E @zcharz thoughts?

Awesome-E:
Hmm this is interesting – starting to have second thoughts on making all the colors red.

> You mention here making the border blue, but in your review of line 42 you requested the border to be red. Which of the two do you prefer?

I meant to make it blue because that's what the borders are in search and in the roadmap

@charlieweinberger you make a good point in [this comment](https://github.com/icssc/peterportal-client/pull/654#pullrequestreview-2785095019), and I do think that just units in blue does look weird. However, I also think that all red also looks worse than it did with everything else being orange for some reason. I'm not quite sure why that is though. Will have to think about this more unfortunately.

charlieweinberger (changes requested):
My suggestion would be to use `#e24932` for both light and dark modes, and preferably to make the unit count also be red.

Current:
Screenshot 2025-04-25 at 2 04 08 PM
Screenshot 2025-04-25 at 2 03 28 PM

My suggestion:
Screenshot 2025-04-25 at 2 04 30 PM
Screenshot 2025-04-25 at 2 02 01 PM

Awesome-E:
On its own, I'm not against units being in red for the roadmap, but I don't like how that looks in search. Alternatively, we can use the more subtle approach, which is to make everything consistent with how search was before, that is, all blue except for the warning icon.

I wonder if it's also possible to just make the trash can only turn red when you hover it – I think this takes time to get used to, but I do feel a lot less overloaded when there's only red over the parts that need my attention. Curious on others' thoughts on that


***

Here's a direct comparison between what I think are the most reasonable options:
Red title + icon, trash can only on hover|Warning Icon only
:-:|:-:
|image

Red everywhere|Red everywhere w/ trash can
:-:|:-:
image|image

Awesome-E:
or... slightly less consistent, but potentially cleaner. This advantage of this one is that the invalid search results are less jarring, but the disadvantage is obviously less consistency.
image

charlieweinberger:
I see the value in not having the trash cans be red by default, but also if they're the same color as the text, it's less obvious that they're clickable buttons.

I definitely think that the units in the roadmap should be red, but I'm not as sure about the units in the sidebar being red, since they're visually separated from the name by the course offering icons.

I don't really like how the border (or technically box-shadow I think) of the invalid courses in the sidebar are red, I like the consistency of them being blue like the other courses.

Combining these points, I'm in favor of either the 4th or 5th options, leaning towards the 5th option, but with red trash can icons everywhere. I think? It's tricky.

charlieweinberger:
Also, @Awesome-E what do you think about my suggestion to make our primary red less saturated, to contrast less with the deep shade of blue we primarily use?

charlieweinberger (changes requested):
Super close to being done, down to the finishing touches! I just have one suggested change and 1 question, not including the few requested code changes.
1. Can you make it so the trash can buttons don't have a background, and it's just the icon as the button? For example, look at the bookmark icon when hovering a course in the sidebar.
2. Currently, invalid courses have blue borders in the sidebar, red borders while being dragged, and then blue borders again in the roadmap. Are we sure the red border while being dragged is the best choice, given that the borders before and after are blue? I'm undecided - one one hand, the red border is inconsistent, but on the other hand, it does help emphasize that the course is invalid.

Awesome-E:
Blue border

site/src/index.css
```diff
@@ -25,6 +25,8 @@ code {
   --blue-accent: #74d1f6;
 
   --green-primary: green;
+  --red-primary: #ce0000;
```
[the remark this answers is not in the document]
zcharz: the second one is for dark mode from what i understand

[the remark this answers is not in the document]
Awesome-E: Dark mode one is defined on line 48. You should delete the one on line 28 so that line 33 is the only one for light mode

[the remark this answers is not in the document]
charlieweinberger: I see it defined twice for light mode:

Screenshot 2025-05-11 at 6 50 40 PM

site/src/pages/RoadmapPage/Course.tsx
```diff
@@ -97,10 +97,12 @@ const Course: FC<CourseProps> = (props) => {
   const tappableCourseProps = props.addMode === 'tap' ? tapProps : {};
 
   return (
-    <div className={`course ${requiredCourses ? 'invalid' : ''}`} {...tappableCourseProps}>
+    <div className="course" {...tappableCourseProps}>
       <div className="course-card-top">
         <div className="course-and-info">
-          <CourseNameAndInfo data={props.data} {...{ openPopoverLeft, requiredCourses }} />
+          <span className={`${requiredCourses ? 'invalid-name' : ''}`}>
```
Awesome-E, line 103: `invalid-name` suggests that the user somehow entered the name and that it's wrong. I would keep the `invalid` className on the course itself but then change the style rule to be `.course.invalid .course-name` (or whatever the class name is for the element that contains the title and warning icon)

Awesome-E (commented):
Looks good visually but have two small nits with the code

charlieweinberger (approved):
Other than the repeat definition of `--red-primary` as mentioned above, this pr looks good to go!

Awesome-E (approved):
🎉

PR #655, opened 2025-04-20
Added Recent Offerings table component and style updates to SideInfo

### LINKED ISSUE #622: Show a Better Visualization of Past Offerings on the Course Page
Currently, the user must hover over a set of icons to show when a course was previously offered. There are numerous issues with the usability and how intuitive this feature is, one of which is the difficulty of reading this data.

The bar chart, while it may look nice, doesn't tell you if the spring offerings were for the past year, or if that was 10 years ago. It would instead be more useful if we could tell which years the course was offered and which years they were not.

I think that we should start by putting this on the course info page so it's very obvious and upfront, though we can discuss changing how the popup works at a different point in time.

Example Design:

**Notes about the design (please read before implementing):**
- The check and x are there as an initial thought, but I think in the design, it's not immediately obvious that they're different due to the similarity between the check and x shapes. There are a few alternative approaches (i.e. adding color), but I think the best approach is to make it so there's a check where it is offered, and leave the cell blank when it's not offered.
- This design does also remove the ratings, but that's not relevant to this issue. For this issue, we are only adding this table directly below the course tags ("4 units", "lower division", "GE Vb"). This _does_ mean it will appear _**above**_ the average ratings.

Other development notes:
- Take a look at some of our types related to quarter names and/or inspect the automatically generated `anteater-api-types.ts` file. You will likely want to have this handy for either creating a new type like <code>type PastQuarter = \`${number}${quarterName}\`</code> and/or grouping a course's past offerings into years.

fiveminus1:
added recent offerings table to SideInfo, which displays a more intuitive visualization of past course offerings directly on the course page. 

## Description

created a table component RecentOfferings that parses the terms string[] into a dict of {academicYear: [bool, bool, bool, bool]}, then displays this onto a table. added this component to SideInfo.

### Before
image
### After
image

## Test Plan

- [ ] confirm table data matches previous SoC/existing CourseQuarterIndicator popup data 

## Issues

Closes #622


site/src/component/SideInfo/SideInfo.tsx
```diff
@@ -161,6 +162,10 @@ const SideInfo: FC<SideInfoProps> = (props) => {
         </div>
       </div>
 
+      <div className="side-info-recent-offerings">
+        <RecentOfferings terms={props.terms ? props.terms : []} />
```
Awesome-E, line 166: I don't think you need to wrap this in a `div` – you already have a wrapper container inside it. Instead, just use the one that's there already, change the `className` to `recent-offerings`, and combine the styles

site/src/component/RecentOfferings/RecentOfferings.scss
```diff
@@ -0,0 +1,34 @@
...
+  min-width: 400px;
+}
+
+.recent-offerings-table td,
+th {
+  min-width: 50px;
+  padding-left: 10px;
+  padding-top: 3px;
+  padding-bottom: 3px;
+}
```
Awesome-E, line 17: The emojis and check marks are not the same width, so you should center these.

Additionally, the style rule should be `.recent-offerings-table :is(td, th)` because otherwise you're saying `.recent-offerings-table td` and all `th` elements

site/src/component/RecentOfferings/RecentOfferings.tsx
```diff
@@ -0,0 +1,75 @@
...
+    let quarterIndex = -1;
+    if (quarter.startsWith('Fall')) quarterIndex = 0;
+    else if (quarter.startsWith('Winter')) quarterIndex = 1;
+    else if (quarter.startsWith('Spring')) quarterIndex = 2;
+    else if (quarter.startsWith('Summer')) quarterIndex = 3;
+    else continue;
+
+    // if the course is not described as a "Fall" course, it should be listed as the previous academic year
+    // e.g. "Winter 2023" should be in "2022-2023", but "Fall 2023" should be in "2023-2024"
+    const academicYear = quarterIndex === 0 ? `${year}-${year + 1}` : `${year - 1}-${year}`;
```
Awesome-E, line 25: I prefer the term `startYear` instead of `academicYear`, and I think that's also what we use for Zot4Plan imports.

site/src/component/RecentOfferings/RecentOfferings.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+    if (!offerings[academicYear]) {
+      offerings[academicYear] = [false, false, false, false]; // Fall, Winter, Spring, Summer
+    }
```
Awesome-E, line 29: Can be simplified into nullish assignment if you wish, but whether that's better is subjective. You don't have to accept this suggestion

```suggestion
    // Initialize Fall, Winter, Spring, Summer of the year to false
    offerings[academicYear] ??= [false, false, false, false];
```

site/src/component/RecentOfferings/RecentOfferings.tsx
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  return offerings;
+}
+
+const RecentOfferings: FC<RecentOfferingsProps> = (props) => {
+  //show in order of fall, winter, spring, summer
+  const termsInOrder = props.terms.slice().reverse();
+
+  const offerings = parseOfferings(termsInOrder);
```
Awesome-E, line 41: Good helper function :+1:

Awesome-E (changes requested):
Definitely good progress on this, I like how it's coming along! Overall code structure is good with a few nitpicks.

As for style, I think you could also apply `transform: scale(1.5)` to `.recent-offerings-table .bi-check` since they're a bit small by default. Additionally, I think you should center the texts and use symmetrical padding `padding: 3px 10px` (top/bottom, left/right) instead of defining some of them and doing that separately. (see review comments)

here's what it'd look like if you center the texts and scale up the check marks.
image

site/src/component/RecentOfferings/RecentOfferings.tsx
fiveminus1: Was debating this, but I think the comment cements it for clarity. I'll keep the change, thanks!

site/src/component/SideInfo/SideInfo.tsx
```diff
@@ -161,6 +162,8 @@ const SideInfo: FC<SideInfoProps> = (props) => {
         </div>
       </div>
 
+      {props.terms && <RecentOfferings terms={props.terms ? props.terms : []} />}
```
Awesome-E, line 165: The `&&` and `props.terms ? props.terms : []` are redundant. Assuming you don't want to render the recent offerings table if there are none:
```suggestion
      {props.terms && <RecentOfferings terms={props.terms} />}
```

Awesome-E (approved):

PR #656, opened 2025-04-21
feat: posthog integration

### LINKED ISSUE #646: Integrate PostHog
# Overview

Integrate PostHog via `posthog-js` with [heatmaps](https://posthog.com/docs/toolbar/heatmaps).

# Scope

This issue should not remove Google Analytics.

It should install PostHog and test to ensure page views are being sent to the dashboard.

ecxyzzy:
## Description
Per title.

Note: This PR checks the key and host into source control for local development. This should be fine as they are not sensitive. Deploy time variables are still handled by GitHub variables.

## Test Plan

Verified locally that pageview event is sent to dashboard.

Also added PeterPortal URL to dashboard allowlist.

## Issues

Closes #646.


site/src/index.tsx
```diff
@@ -15,7 +16,13 @@ import { Provider } from 'react-redux';
 ReactDOM.createRoot(document.getElementById('root')!).render(
   <React.StrictMode>
     <Provider store={store}>
-      <App />
+      <PostHogProvider
+        apiKey={import.meta.env.VITE_PUBLIC_POSTHOG_KEY}
+        options={{ api_host: import.meta.env.VITE_PUBLIC_POSTHOG_HOST, autocapture: true, enable_heatmaps: true }}
+      >
+        {' '}
```
Awesome-E, line 23: Is this necessary?
```suggestion
```

Awesome-E (commented):
Just one quick question - and it looks like this will only be on prod, which makes sense.

site/src/index.tsx
ecxyzzy: I think I initially codevomited over index.tsx and prettier decided to introduce this "fix" - just removed it and it didn't try to re-add it so will push that change

Awesome-E (approved):
LGTM!

Issue #14, opened 2020-12-29
Semantic-UI  →  Bootstrap


uci-mars:

ethanwong16:
Semantic-UI may not be maintained anymore in the future

uci-mars:
@ethanwong16 yessir

PR #660, opened 2025-04-25
Remove semantic

ethank2222:
## Description

Removed dependency on semantic-ui

Overview: Semantic UI has been experiencing a general lack of maintenance, and has become unnecessary for our codebase. In the future, we want to use Bootstrap and custom styles for frontend design, rather than relying on legacy dependencies. To remove all instances semantic ui, we need to remove Button, Divider, Dropdown, Popup, Label, and List objects.

- Button: Changed to Bootstrap buttons
- Divider: Replaced with custom spacing formatting
- Dropdown: Changed to Bootstrap dropdowns
- Popup: Changed to Bootstrap popover
- Label: Removed all instances
- Grid: Removed all instances
- Icon: Removed all instances

## Screenshots

**All frontend designs should remain the same**

Button, Label, Popup, Dropdown, Divider:
image
image

Icon:
image
image

Grid, Popup, Button:
image
image

## Test Plan

Verify all pages are not noticeably different (light and dark mode)

## Issues

Closes [#14](https://github.com/icssc/peterportal-client/issues/14)


site/src/component/AppHeader/Profile.tsx
```diff
@@ -54,23 +53,24 @@ const Profile = () => {
...
+                <Sticky />
               </div>
               Your Reviews
             </NavLink>
           </li>
           <li>
             <button className="theme-button profile-popover__link" onClick={() => setTab('theme')}>
               <div>
-                <Icon name={usingSystemTheme ? 'laptop' : darkMode ? 'moon outline' : 'sun outline'} size="large" />
+                {/* <Icon name={usingSystemTheme ? 'laptop' : darkMode ? 'moon outline' : 'sun outline'} size="large" /> */}
```
Awesome-E, line 64: this might've been me 😅, but we should remove commented code
```suggestion
```

site/src/component/AppHeader/Profile.tsx
```diff
@@ -95,7 +95,8 @@ const Profile = () => {
               onClick={() => setTheme('light')}
             >
               <div>
-                <Icon name="sun outline" size="large" />
+                {/* <Icon name="sun outline" size="large" /> */}
```
Awesome-E, line 98:
```suggestion
```

site/src/component/AppHeader/Profile.tsx
```diff
@@ -106,7 +107,8 @@ const Profile = () => {
               onClick={() => setTheme('dark')}
             >
               <div>
-                <Icon name="moon outline" size="large" />
+                {/* <Icon name="moon outline" size="large" /> */}
```
Awesome-E, line 110:
```suggestion
```

site/src/component/AppHeader/Profile.tsx
```diff
@@ -117,7 +119,8 @@ const Profile = () => {
               onClick={() => setTheme('system')}
             >
               <div>
-                <Icon name="laptop" size="large" />
+                {/* <Icon name="laptop" size="large" /> */}
```
Awesome-E, line 122:
```suggestion
```

site/src/component/PPCOverlayTrigger.tsx
Awesome-E, line 1: Nothing wrong with Ethan's code here, but I did make this comment while we worked on it together –

This is a great example of stuff we would avoid by standardizing components with MUI... bootstrap just had horrible defaults for this and we had to pull logic together from separate parts of our app to get the behavior we wanted.
(cc @alexanderl19 @ecxyzzy)

site/src/component/Review/Review.tsx
```diff
@@ -57,6 +59,20 @@ const Review: FC<ReviewProps> = (props) => {
     sortedReviews = reviewData.slice(0);
   }
 
+  // calculate frequencies of professors or courses in list of reviews
```
Awesome-E, line 62: Note for other reviewers: This reordering fixes the issue where when filtering by a certain prof/course for reviews, other profs/courses would always show that they had zero reviews, even when they didn't.

site/src/component/SideBar/SideBar.tsx
Awesome-E, line 1: Still not the biggest fan of this change from a visual standpoint... ideally whatever we do next with UI and unified tooling will let us use icons that are better than the bootstrap ones

site/src/pages/CoursePage/index.tsx
```diff
@@ -63,15 +62,15 @@ const CoursePage: FC = () => {
               <div>
                 <h2>🌲 Prerequisite Tree</h2>
               </div>
-              <Divider />
+              <br />
```
Awesome-E, line 65: Maybe we should consider using `<hr/>` instead of `<br/>` to replace divider, since I believe there is a line currently:

image

Awesome-E: This is subjective though, so if you don't find this necessary then I guess it's fine.

Awesome-E:
My review is mainly a sanity check – we can wait for @charlieweinberger's review or also have @CadenLee2 review if he has time

CadenLee2 (commented):
## Objective changes
- Tested on staging: anywhere with changed components for professors/courses pages, review menu/confirmation, roadmap
- All changes seem to be functional and I don't see any objective issues

## Opinions
- I like the refactoring of the PPC Overlay Trigger
- Overall, since I'm so used to the current appearance of PeterPortal, it was a bit disorienting to see the new icons and font size. Other than the left sidebar, the new icons seem okay. However, especially on the roadmap page, having the larger font makes everything seem much bigger and bolder. If we compare it to other websites at the same zoom level, things feel unnaturally larger, especially on a small desktop screen (on mobile it seems okay). I understand that the smaller font size was a result of Semantic, but I think I may have preferred it.
- See comments for specific opinions

Image: comparing GitHub and the staging instance of PeterPortal, both at the same zoom level. Notice the font size difference, which makes PeterPortal look like it's been zoomed in
image

site/src/component/PPCOverlayTrigger.tsx
charlieweinberger: I agree, that if AntAlmanac and PeterPortal are going to merge like planned, there should be more communication between the two development teams about standardizing icons and UI/UX design choices as a whole. Possibly also the role of the UI/UX team to make sure that the two projects (and possibly other projects) have a consistent feel between them.

site/src/pages/CoursePage/index.tsx
charlieweinberger: I can go either way on the dividers (keeping them vs. removing them), as long as they're consistent between elements, i.e. the prereq tree, SoC, grade dist, and reviews either all do or all don't have dividers. As of now, I prefer them all having dividers, but I haven't gone through this entire PR yet so my opinion may change.

site/src/component/AppHeader/Profile.tsx
```diff
@@ -167,7 +164,7 @@ const Profile = () => {
         <a href={`/api/users/auth/google`}>
           <Button variant={darkMode ? 'dark' : 'light'}>
             <span>
-              <Icon name="sign in" className="login-icon" />
+              <BoxArrowInRight />{' '}
             </span>
             Log In
           </Button>
```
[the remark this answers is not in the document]
charlieweinberger: I agree that the button not being centered is strange, good call.

The issue with `ppc-btn` is indeed out of the scope of this PR, and is actually something I've been working on in my own time to remedy. I'm still figuring out the details tho, and it's definitely not as important as this PR or other PRs we're going on, so it might take a while before being fixed on main.

site/src/component/AppHeader/AppHeader.scss
```diff
@@ -107,3 +107,36 @@
...
+  align-items: center;
+  color: white;
+  border-radius: 4px;
+  font-size: 0.9em;
+  font-weight: bold;
+  text-decoration: none;
+  cursor: pointer;
+  overflow: hidden;
+
+  > * {
```
charlieweinberger, line 133: I'm generally not a fan of this kind of CSS, since it's more difficult IMO to understand, as opposed to applying styles to the children directly.

site/src/component/AppHeader/AppHeader.scss
```diff
@@ -107,3 +107,36 @@
...
+    padding: 4px 12px;
+  }
+
+  .shade {
+    background-color: #0005;
```
charlieweinberger, line 138: I get the idea of this like, to make the background 33% opacity. However, my first thought was to try to replace this with a pre-defined color. Turns out, in dark mode, this background color is almost identical to `--blue-secondary`. However, in light mode, `--blue-secondary` is lighter than `--blue-primary`, not darker, so this shade looks off. Because of this, while I'm not a fan of `#0005`, I think it's the best solution for now.

However, this line raises a valuable discussion point - there are several places throughout our codebase that use `--blue-primary` and also want to use a darker shade of blue, *even in light mode*. In dark mode, we can just use `--blue-secondary`, but in light mode, we don't already have a darker shade of blue to use to contrast `--blue-primary`. I think having a darker shade, like what secondary is to primary in dark mode, would be helpful to have in light mode as well, to replace the need for things like `#0005`. One other area I know this would be useful is for primary buttons, which I can explain more another time.

To conclude, I think `#0005` is fine for now, but I'd like to replace it in the future.

site/src/component/AppHeader/AppHeader.tsx
```diff
@@ -49,53 +87,12 @@ const AppHeader: FC = () => {
...
-                      Feedback
-                    </a>
-                  </div>
-                </Grid.Column>
-              </Grid>
-            </Popup>
+            <PPCOverlayTrigger popoverContent={popover} placement="bottom">
+              <div id="beta-tag">
+                <div>beta</div>
+                <div className="shade">v1.2</div>
```
charlieweinberger, line 93: Should this be an id, instead of a class? I know the difference is negligible, but if `shade` is only used once for this element, an id makes more sense than a class. That is, if we even care, which I do but it's understandable if you guys don't.

site/src/component/AppHeader/Profile.tsx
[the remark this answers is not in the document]
charlieweinberger: After looking more into it, the feedback icons in the beta disclaimer also aren't vertically centered. It's possible to vertically center these buttons by creating a class `header-button`, defined below, and applying it to the "report an issue" button, the feedback button, and the log-in button. However, this sort of styling should ideally already be applied, so I'm not sure why this problem is showing up for these buttons but seemlingly not for other buttons, since they all use Bootstrap's `<Button />` class. Maybe this has something to do with `ppc-btn` like I mentioned previously. Also, in my opinion, the chat icon in the feedback button looks slightly worse with this styling. If we want, we can apply this class to these 3 buttons (or just 2, leaving out the feedback button).

```
.header-button {
  display: flex;
  justify-content: center;
  align-items: center;
  gap: 4px;
}
```

site/src/component/AppHeader/AppHeader.scss
```diff
@@ -15,7 +15,7 @@
 
   .beta-popup .button:hover,
   .beta-popup .button:focus {
-    background-color: var(--hover);
+    background-color: #cacbcd;
```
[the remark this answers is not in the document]
Awesome-E: Will be resolved in unifying components and adopting MUI.

site/src/component/AppHeader/AppHeader.tsx
Awesome-E: IDs are generally for things serve a specific functionality/feature of an app rather than visuals. However, similar to the other comment, this is more of a stopgap measure before we fully transfer to MUI and will only be up for a few weeks.

site/src/component/AppHeader/AppHeader.tsx
```diff
@@ -29,6 +33,40 @@ const AppHeader: FC = () => {
...
+        </a>
+        <a
+          href="https://form.asana.com/?k=4h9ZTRkVUT9ZwfJrmvxDDw&d=1208267282546207"
+          target="_blank"
+          rel="noopener noreferrer"
+        >
+          <Button variant={buttonVariant}>
+            <ChatLeftDotsFill /> Feedback
+          </Button>
+        </a>
```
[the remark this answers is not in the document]
Awesome-E: I suppose that works, but I am unfortunately in the position where I'd like to merge this sooner than later

site/src/component/AppHeader/Profile.tsx
```diff
@@ -167,7 +164,7 @@ const Profile = () => {
         <a href={`/api/users/auth/google`}>
           <Button variant={darkMode ? 'dark' : 'light'}>
             <span>
-              <Icon name="sign in" className="login-icon" />
+              <BoxArrowInRight />{' '}
             </span>
             Log In
```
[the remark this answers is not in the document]
Awesome-E: I didn't look that closely as we were just swapping out the icons rather than cleaning up the code elsewhere, but yeah I agree this can be simplified

site/src/component/GradeDist/GradeDist.tsx
```diff
@@ -240,8 +235,7 @@ const GradeDist: FC<GradeDistProps> = (props) => {
     return (
       <div className={`gradedist-module-container ${props.minify ? 'grade-dist-mini' : ''}`}>
         {optionsRow}
-
-        <Grid.Row id="chart">
+        <div className="flex">
```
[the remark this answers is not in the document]
Awesome-E: sure, it's a common style, but it does what it says and is part of a larger selector. Can change this though.

site/src/component/AppHeader/Profile.scss
```diff
@@ -99,7 +99,13 @@
   }
 
   li div {
-    display: inline;
+    display: flex;
+    gap: 4px;
+    align-items: center;
```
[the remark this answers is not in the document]
Awesome-E: Not sure that this is the same thing; maybe to address in a different PR since this was just a patch

Awesome-E (approved):

site/src/component/AppHeader/AppHeader.scss
Awesome-E: Same as other comments – app header is more of a stopgap change

site/src/component/AppHeader/Profile.tsx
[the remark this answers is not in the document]
Awesome-E: Will keep this in mind for review during our next UI library change

PR #662, opened 2025-04-26
feat: otel integration

### LINKED ISSUE #648: OTEL Integration
# Overview

Implement Open Telemetry with the PeterPortal API. We will be using Dash0 to ingress and view our data.

# Scope

- Install and setup OTEL instrumentation.
- Create custom Spans for important actions.
- Ensure Lambda collector is functioning properly

ecxyzzy:
## Description
Integrate backend with OTEL.

## Test Plan
Testing in staging then editing prod workflow.

## Issues

Closes #648.


api/package.json
```diff
@@ -20,6 +21,7 @@
     "drizzle-orm": "^0.37.0",
     "express": "^4.21.1",
     "express-session": "^1.18.1",
+    "install": "^0.13.0",
```
Awesome-E, line 24: I assume this is where it's used?

```
nodejs: {
      install: ['@opentelemetry/auto-instrumentations-node'],
    }
```

sst.config.ts
```diff
@@ -31,6 +31,14 @@ function createLambdaFunction() {
...
     ANTEATER_API_KEY: process.env.ANTEATER_API_KEY!,
     EXTERNAL_USER_READ_SECRET: process.env.EXTERNAL_USER_READ_SECRET!,
+    OTEL_EXPORTER_OTLP_HEADERS: process.env.OTEL_EXPORTER_OTLP_HEADERS!,
+    // hardcoded OTEL options
+    ...($app.stage === 'prod' && {
+      AWS_LAMBDA_EXEC_WRAPPER: '/opt/otel-handler',
+      NODE_OPTIONS: '--require @opentelemetry/auto-instrumentations-node/register',
+      OTEL_SERVICE_NAME: 'peterportal-backend',
+      OTEL_EXPORTER_OTLP_ENDPOINT: 'https://ingress.us-west-2.aws.dash0.com',
+    }),
```
Awesome-E, line 41: Not quite sure if I like this as a spread operator as opposed to a separate if statement and `Object.assign`, but otherwise this looks fine.

api/package.json
ecxyzzy: I don't actually think that's necessary and am not sure how that got in - will try removing and seeing if things break

ecxyzzy: yeah it was a copypasta fail. will push a commit removing it. good catch :+1:

image

Awesome-E (approved):
kewl :shipit:

PR #663, opened 2025-04-27
feat: auto-rename ge requirements

### LINKED ISSUE #642: Auto-rename GE requirements more intelligently
Instead of "Select 1 of the following" being the name of the GE Requirement, we should look for a common prefix string (excluding parenthesized text) and use that as the title of the requirement.

Some examples below, though this may require hard-coding certain replacements to change abbreviations to full words

For upper-div writing, we should extract the common prefix of "I. Upper-Division Writing"
Image

For Foreign Language Requirements, we'd extract "VI. Language"
Image

For International/Global Issues, we'd extract "VIII. International/Global Issues"
Image

fiveminus1:
## Description

Automatically renames top-level "Select 1 of the following" labels by extracting a common prefix from its children requirements. 

Some of the words in the children labels may be common prefixes, but abbreviated. For example, "VIII. International/Global Issues" should have a common prefix in "VIII. Int'l/Global Issues" and "VIII. International/Global Issues," but "Int'l" should be expanded to "International." To account for this, a regex is used, and such abbreviations are recorded in `src/helpers/substitutions/labelSubstitutions`.  See #642.

## Screenshots
- Before 
image
image

- After
image
image

## Test Plan

- Compare new `getCoursesForGE()` return array with original response from 
`trpc.programs.getRequiredCoursesUgrad.query({ id: 'GE' });` to confirm the replaced labels

## Issues

Closes #642


laggycomputer:
thamk you ehrb very cool

fiveminus1:
laggy comptuer

package-lock.json
Awesome-E, line 1: Looks like you added this by accident? Probably happened because you did an `npm install` instead of `pnpm install` in the root

site/src/pages/RoadmapPage/sidebar/GERequiredCourseList.tsx
```diff
@@ -4,9 +4,25 @@ import trpc from '../../../trpc';
...
+import { findCommonLabelPrefix } from '../../../helpers/substitutions';
 
-function getCoursesForGE() {
-  return trpc.programs.getRequiredCoursesUgrad.query({ id: 'GE' });
+async function getCoursesForGE() {
+  const fetchedCourses = await trpc.programs.getRequiredCoursesUgrad.query({ id: 'GE' });
+
+  /* if a top level "Select 1 of the following" group of requirements is seen, attempt to replace the label */
+  fetchedCourses.forEach((geRequirement) => {
+    if (geRequirement.requirementType === 'Group' && geRequirement.label === 'Select 1 of the following') {
```
Awesome-E, line 14: You can reduce a level of curly braces using a guard clause

```ts
if (req.requirementType !== 'Group' || req.label !== 'Select 1 of the following') return;
```

site/src/pages/RoadmapPage/sidebar/GERequiredCourseList.tsx
```diff
@@ -4,9 +4,25 @@ import trpc from '../../../trpc';
...
+      const requirementSubLabels = geRequirement.requirements.map((req) => req.label);
```
Awesome-E, line 15: nitpick on naming – can be something more like `nestedLabels` since 'sub' can be ambiguous and 'requirement' is redundant

site/src/helpers/substitutions.ts
```diff
@@ -0,0 +1,32 @@
...
+    expandedLabel = expandedLabel.replace(regex, substitution);
+  }
+
+  return expandedLabel;
+}
+
+export function findCommonLabelPrefix(labels: string[]): string {
+  if (labels.length === 0) return '';
+
+  const expandedLabels = labels.map(expandAbbreviation); // expand any words in labels that may be abbreviated
```
Awesome-E, line 21: comment is unnecessary
```suggestion
  const expandedLabels = labels.map(expandAbbreviation);
```

site/src/helpers/substitutions.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  expandedLabels.sort();
+  const first = expandedLabels[0].split(' ');
+  const last = expandedLabels[expandedLabels.length - 1].split(' ').filter((word) => !word.startsWith('(')); // ignore any words within parentheses
```
Awesome-E, line 25: I guess this is one way of terminating match since "I. Upper-Division Writing (min grade C-) (subject)" will be different from "I. Upper-Division Writing grade C-) (subject 2)", but I think it's probably more understandable if we just remove all text after the first open paren (`.replace(/\s\(.*/g, '')`) before filtering like this.

Also you can get last item with `arr.at(-1)` instead of `arr[arr.length - 1]`

site/src/pages/RoadmapPage/sidebar/GERequiredCourseList.tsx
```diff
@@ -4,9 +4,25 @@ import trpc from '../../../trpc';
 import RequirementsLoadingIcon from './RequirementsLoadingIcon';
 import { setGERequirements } from '../../../store/slices/courseRequirementsSlice';
 import { useAppDispatch, useAppSelector } from '../../../store/hooks';
+import { findCommonLabelPrefix } from '../../../helpers/substitutions';
 
-function getCoursesForGE() {
-  return trpc.programs.getRequiredCoursesUgrad.query({ id: 'GE' });
+async function getCoursesForGE() {
+  const fetchedCourses = await trpc.programs.getRequiredCoursesUgrad.query({ id: 'GE' });
```
Awesome-E, line 10: variable name is inaccurate, as this is a requirements list.

Additionally, there is already a function to normalize responses, so the logic written here might be better suited to go alongside other logic that does similar things. Even if you do only apply it to GEs, it still might be worth defining a separate function to normalize its data there

Awesome-E (changes requested):
hi!

site/src/helpers/substitutions.ts
```diff
@@ -0,0 +1,52 @@
...
+  return expandedLabel;
+}
+
+export function findCommonLabelPrefix(labels: string[]): string {
+  if (labels.length === 0) return '';
+
+  const expandedLabels = labels.map(expandAbbreviation);
+
+  expandedLabels.sort();
+  /* remove all text after parentheses */
```
Awesome-E, line 26: I would do the replace in the `expandAbbreviation` function and potentially rename that function to "formatLabel"

site/src/helpers/substitutions.ts
```diff
@@ -0,0 +1,52 @@
...
+    .split(' ');
+
+  let i = 0;
+  while (i < first.length && first[i] === last[i]) {
+    i++;
+  }
+  return first.slice(0, i).join(' ');
+}
+
+export function normalizeGERequirements(fetchedRequirements: ProgramRequirement[]): void {
```
Awesome-E, line 40: on the topic of naming functions, "inferLabelTitle" or something similar might be more descriptive, and may later be useful for non-GE requirements as well.

Awesome-E (commented):
Pretty much ready, just a few small changes if you want to make those

Awesome-E (approved):
LGTM but might request a second review

ellisl10 (approved):

PR #665, opened 2025-04-29
Add AP Exam Credits for Redesigned Transfers

### LINKED ISSUE #633: User should be able to add and receive appropriate credit from AP Exams
This is the redesigned AP Exam Import Experience – a user should be able to type in a name and a score, and PeterPortal should be able to automatically clear course and GE prerequisites.

- Exam names should autocomplete
- Once exam name and score are both entered, a tile like the one shown in the design should be added
- The score should be editable via a standard HTML `<select>` element styled as shown
- The units should use the same component as #632
- AP Exam imports should have their respective courses and GEs counted
- Deleting an AP Exam import should function as expected.

Relevant design from the Figma:

Image

ellisl10:
AP Exam transfers

## Description

Created new trpc routes to update, save, and delete exams and updated transferCreditsSlice to store exam info and user's AP exams.

## Screenshots

Screenshot 2025-05-08 at 4 16 16 PM

## Test Plan

Verify that you can save exams, update the score/units, and that it will save to the user's account. The tile should correctly display the courses cleared and the units granted depending on the user's score.

## Issues

Closes #633

### DIFF
--- a/api/src/controllers/transferCredits.ts
+++ b/api/src/controllers/transferCredits.ts
@@ -13,13 +13,9 @@ interface userAPExam {
 }
 
 const zodAPExamSchema = z.object({
-  apExams: z.array(
-    z.object({
-      examName: z.string(),
-      score: z.number(),
-      units: z.number(),
-    }),
-  ),
+  examName: z.string(),
+  score: z.number(),
+  units: z.number(),
 });
 
 /** @todo complete all routes. We will remove comments after all individual PRs are merged to avoid merge conflicts */
@@ -35,7 +31,9 @@ const transferCreditsRouter = router({
       .then((res) => (res.data ? (res.data as APExam[]) : []));
     return response;
   }),
+  // only update one row at a time
   getSavedAPExams: publicProcedure.query(async ({ ctx }): Promise<userAPExam[]> => {
+    /** you should return the data as-is, then handle the null case from the frontend within an individual APExamTile */
     const userId = ctx.session.userId;
     if (!userId) return [];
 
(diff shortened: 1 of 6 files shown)


laggycomputer:
thank you ellis peter portal very cool

api/src/controllers/transferCredits.ts
```diff
@@ -1,9 +1,37 @@
...
+  rewards: {
+    acceptableScores: number[];
+    unitsGranted: number;
+    electiveUnitsGranted: number;
+    geCategories: string[];
+    coursesGranted: {
+      string: string[];
+    };
+  }[];
+}
```
Awesome-E, line 16: When you `pnpm install`, API responses have automatically generated types. You should be able to do this to get the type instead of redefining the interface yourself:

```ts
import { operations } from "./generated/anteater-api-types";
export type APExam = operations['apExams']['responses'][200]['content']['application/json']['data']
```

api/src/controllers/transferCredits.ts
```diff
@@ -1,9 +1,37 @@
...
+
+const getAPExams = async (): Promise<APExam[]> => {
+  const response = await fetch(`${process.env.PUBLIC_API_URL}apExams`, {
+    headers: ANTEATER_API_REQUEST_HEADERS,
+  })
+    .then((res) => res.json())
+    .then((res) => (res.data ? (res.data as APExam[]) : []));
+  return response;
+};
```
Awesome-E, line 25: (subjective thought here)
I personally think this can just go straight in `getAPExamInfo`, as its functionality is single-use and only applies to APs rather than to all of `transferCredits`

site/src/pages/RoadmapPage/transfers/APExamsSection.tsx
```diff
@@ -1,17 +1,42 @@
...
+    acceptableScores: number[];
+    unitsGranted: number;
+    electiveUnitsGranted: number;
+    geCategories: string[];
+    coursesGranted: {
+      string: string[];
+    };
+  }[];
+}
+
```
Awesome-E, line 28: Looks like you've redefined the same interface. I know it's still a draft PR, but I would suggest in that case creating `transferCredits.ts` in `types/src/` and adding those two lines from the previous comment

site/src/pages/RoadmapPage/transfers/APExamsSection.tsx
```diff
@@ -1,17 +1,42 @@
...
   /** @todo style this according to the Figma, add options & optgroup titled 'Score' */
   return (
     <select value={score ?? ''} onInput={(event) => setScore(parseInt((event.target as HTMLInputElement).value))}>
-      <option>1</option>
+      <optgroup label="Score">
+        <option value="1">1</option>
+        <option value="2">2</option>
+        <option value="3">3</option>
+        <option value="4">4</option>
+        <option value="5">5</option>
```
Awesome-E, line 38: If I'm not mistaken, `<option>` tags without values default to their content, so you could do `<option>1</option>`

site/src/pages/RoadmapPage/transfers/APExamsSection.tsx
```diff
@@ -31,13 +56,37 @@ const APCreditMenuTile: FC = () => {
...
+
   return (
     <MenuSection title="AP Exam Credits">
       <SectionDescription>
         Enter the names of AP Exams that you&rsquo;ve taken to clear course prerequisites.
       </SectionDescription>
       {/** @todo map each user AP Exam to one of these tiles */}
       <APCreditMenuTile />
+      <div style={{ display: 'flex', alignItems: 'center', gap: '1rem' }}>
+        <div style={{ flex: '1' }}>
```
Awesome-E, line 83: I assume this is temporary, but don't use inline styles

Awesome-E (changes requested):
Mostly so you have a record of what I said yesterday in the meeting, but I did have a couple other things I noticed

Awesome-E (approved):

PR #666, opened 2025-04-29
Transfers data conversion

### LINKED ISSUE #635: Transferred Credits Data Conversion Pipeline

CadenLee2:
Created a script to organize the data for transferred credits, which is currently stored in the `transferredMisc` table. Valid AP exams will be moved into `transferredApExam`, and valid courses into `transferredCourse`. EDIT: duplicated misc transfers will be combined into single ones. This script will need to be executed only once, as it will immediately convert all of the data.

⚠️ **When it is finished, this script will only need to be executed once. Do not arbitrarily run it just to test it unless you have ensured you understand it.**

## Description

Created a new `api/scripts` folder containing this standalone script, which can be executed using `npx tsx` (may need to install `tsx`).

As mentioned in issue #635, the high-level process looks like:
- Get all AP exams from the API
- Get all misc transfers from the db
- For each transfer row:
  - If it is an AP exam (starts with "ap "), try to match it with an AP exam we know about; if matched, it will be moved to the transferred AP table
  - If it is not an AP exam, try to match it with a course from the API; if matched, it will be moved to the transferred courses table
  - If it does not match with a course or AP exam, it will remain in the misc transfers table; if it is a duplicate (the same user has the same course more than once), all the identical rows will be combined into one with their units added together
- During this process, we build up several large SQL queries with Drizzle, which will be executed at the end:
  - Insert all the moved AP exams into the AP exams table
  - Insert all the moved courses into the courses table
  - Delete all the originals of the moved AP exams and courses (as well as the duplicated misc transfers) from the misc table
  - EDIT: additionally, reinsert the combined versions of the duplicated misc transfers

EDIT: output when running the script includes:
- All AP exams (useful for debugging)
- Which rows matched and which ones failed during the categorization process
- The resulting database queries, which are placed into individual log files (so as not to flood the terminal because they might be very large); these files can be deleted when certain everything works

We should be careful to ensure there are no issues with these large queries because they involve a lot of data manipulation at once.

## Screenshots

A preview of what some of the output might look like for a variety of cases:
```
x FAILED:  x2    'miscTest'      could not be matched with any course
- COMBINED:      'miscTest' (above): duplicates will be combined into one entry
  MATCHED: x1    'AP STATISTICS' with: 'AP Statistics' ('AP STATISTICS')
x FAILED:  x1    ''      could not be matched with any course
  MATCHED: x1    'AP US HISTORY' with: 'AP United States History' ('null')
  MATCHED: x1    'I&C SCI 32' with: 'I&C SCI 32'
  MATCHED: x1    'AP MICROECONOMI' with: 'AP Microeconomics' ('AP ECONOMICS:MICRO')
+ MATCHED:       'AP CALCULUS AB' with the AP Calc AB Subscore on the BC test AS WELL AS:
  MATCHED: x2    'AP CALCULUS AB' with: 'AP Calculus AB' ('AP CALCULUS AB')
x FAILED:  x1    'AP CALCULUS AB'      has ambiguity: AP Calculus AB vs the BC Subscore
```

## Test Plan

- [ ] Confirm code looks good and will function properly in all circumstances (for instance, consider various possible variations of AP exam names)
- [ ] When we are ready to run the script:
  - [ ] Test the script *without* the final delete/insert queries to ensure the right query would be executed (print it out instead)

## Issues

Closes #635


api/scripts/transfersDataConversion.ts
```diff
@@ -0,0 +1,406 @@
...
+  } else {
+    // The user has AP Calculus AB 2+ times and no AB subscore,
+    // so we assume one of them is the BC subscore and the other isn't
+    // Insert the BC subscore here, then proceed normally so that AB is also inserted
+    console.log(`+ MATCHED:       '${transferName}' with the AP Calc AB Subscore on the BC test AS WELL AS:`);
+    toInsertAp.push({
+      userId: transfer.userId,
+      examName: AP_CALC_AB_SUBSCORE,
+      score: null,
+      units: 0, // The units will be counted for the actual AB later, and the subscore will remain 0
```
Awesome-E, line 207: Nice, this is a good approach!

api/scripts/transfersDataConversion.ts
```diff
@@ -0,0 +1,406 @@
...
+/** Normalize a transfer name by converting it to lowercase
+  then removing punctuation and trailing/leading whitespace;
+  if undefined, return an empty string
+  "&" will become "and" */
+const normalizeTransferName = (transferName: string | undefined) => {
+  if (!transferName) return '';
+  return transferName
+    .toLowerCase()
+    .replace(/[.,/#!$%^*;:{}=\-_`~()]/g, '')
+    .replace('&', 'and')
```
Awesome-E, line 81: Are there any cases where there is more than one `&`? It might be good to make this match globally.

api/scripts/transfersDataConversion.ts
```diff
@@ -0,0 +1,406 @@
...
+      continue;
+    }
+    const bestMatch = tryMatchAp(transfer.courseName, allAps);
+    if (bestMatch && bestMatch.fullName == AP_CALC_AB_SUBSCORE) {
+      usersWithABSubscore[transfer.userId] = true;
+    }
+  }
+
+  // Build several large queries
+  const toDelete: (SQL<unknown> | undefined)[] = []; // `or()` should still be valid if we pass nothing into it
```
Awesome-E, line 325: Do you have reference supports this assumption? If there are zero, what does it return?

Alternatively, we could be safe and default it to `[false]` (or however you define a false constant condition) since that will default an empty OR to false and be ignored in non-empty ORs.

Awesome-E (commented):
Great work. Will have @ellisl10 take a second look at this for logic errors but otherwise I would like to set up a time to test and run this with prod data.

api/scripts/transfersDataConversion.ts
CadenLee2: I don't think there are any cases of that, but I agree that matching globally is better here

api/scripts/transfersDataConversion.ts
CadenLee2: Since I couldn't find info in the docs, to be safe I defaulted it to ``sql`FALSE` ``. Tested with an empty transfers array and it produced a query ending with `where FALSE` (so it won't delete anything, which is what we want)

api/scripts/transfersDataConversion.ts
```diff
@@ -0,0 +1,406 @@
...
+*/
+const apMatchQuality = (normalizedName: string, substitutedName: string, ap: ApExamBasicInfo): number => {
+  const normalizedApFull = normalizeTransferName(ap.fullName);
+  const normalizedApCat = normalizeTransferName(ap.catalogueName);
+  if (normalizedApFull == normalizedName) return 4;
+  if (normalizedApCat == normalizedName) return 4;
+  if (normalizedApFull.startsWith(normalizedName)) return 3;
+  if (normalizedApCat.startsWith(normalizedName)) return 3;
+  if (normalizedApFull == substitutedName) return 2;
+  if (normalizedApCat == substitutedName) return 2;
```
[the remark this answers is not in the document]
CadenLee2: That's a good point, I don't think this will matter too much but I changed it because it does make more sense that the substitutions exact match is better

ellisl10 (approved):

api/scripts/transfersDataConversion.ts
```diff
@@ -0,0 +1,406 @@
...
+  return bestMatch;
+};
+
+const removeAllSpaces = (transferName: string) => {
+  return transferName.replace(/\s/g, '');
+};
+
+/** Try to match a transfer name with an existing UCI course; undefined if no match */
+const tryMatchCourse = async (transferName: string): Promise<string | undefined> => {
+  const res = await getAPICourseById(removeAllSpaces(transferName));
```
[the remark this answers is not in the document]
CadenLee2: Done, requesting review again

PR #667, opened 2025-04-30
Merge `TypedProgramRequirement` into `ProgramRequirement`

Awesome-E:
## Description

This is a small PR to merge the typescript definitions of `ProgramRequirement` and `TypedProgramRequirement`. The PR simply uses optional typescript arguments to make it so that `ProgramRequirement` can optionally take an argument.

Before, you had to do `let req: ProgramRequirement` for generic and `let req: TypedProgramRequirement<'Course'>` for specific. Now, generic is the same, but you can just do `ProgramRequirement<'Course'>` for specific, and it should work with Intellisense

## Screenshots
This is not frontend, but here:

image

## Test Plan
- [ ] Intellisense works
- [ ] nothing breaks

## Issues

N/A

### DIFF
--- a/site/src/helpers/courseRequirements.ts
+++ b/site/src/helpers/courseRequirements.ts
@@ -1,10 +1,4 @@
-import {
-  MajorProgram,
-  MajorSpecialization,
-  MinorProgram,
-  ProgramRequirement,
-  TypedProgramRequirement,
-} from '@peterportal/types';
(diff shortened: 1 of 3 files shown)


CadenLee2 (approved):
- Tested a few requirements on the staging instance and nothing seems to be broken, no new console errors or anything
- Code completion/intellisense works
- Code changes look good

image

PR #668, opened 2025-04-30
 Add Unconverted Transfer Credits Section for Redesigned Transfers 

### LINKED ISSUE #636: Unconverted Transfer Credits should be shown to the user
Credits that were not able to be automatically converted by #635 should be shown at the bottom of the Transfer Credits UI. These will be items from the current `transferred_course` table (this table may, however, be renamed to something like `legacy_transfer` since `transferred_course` is desirable for new converted data).

Example Design:

zcharz:
## Description

2 tRPC routers: 
1. retrieve all unconverted transfer courses from database
2. remove singular unconverted transfer course

display a list of tiles, each mapped to an unconverted transfer course
each tile should be able to be deleted using the 

## Screenshots

before: nothing

after:
image

## Test Plan
- [ ] retrieves and displays unconverted transfers (correct name/info, number of courses, etc)
- [ ] delete removes correct component from view
- [ ] delete removes correct row from database
- [ ] no uncategorized transfers results in empty section

modify a user's transfer courses using the main website to ensure that the data is correct

## Issues

Closes #636

### DIFF
--- a/site/src/pages/RoadmapPage/transfers/UncategorizedCreditsSection.tsx
+++ b/site/src/pages/RoadmapPage/transfers/UncategorizedCreditsSection.tsx
@@ -29,7 +29,7 @@ const UncategorizedCreditsSection: FC = () => {
     trpc.transferCredits.getUncategorizedTransfers.query().then((response) => {
       dispatch(setUncategorizedCourses(response));
     });
-  }, [dispatch, trpc.transferCredits.getUncategorizedTransfers]);
+  }, [dispatch]);
 
   if (courses.length === 0) {
     return null;
--- a/site/src/pages/RoadmapPage/transfers/UncategorizedCreditsSection.tsx
+++ b/site/src/pages/RoadmapPage/transfers/UncategorizedCreditsSection.tsx
@@ -29,7 +29,7 @@ const UncategorizedCreditsSection: FC = () => {
     trpc.transferCredits.getUncategorizedTransfers.query().then((response) => {
       dispatch(setUncategorizedCourses(response));
     });
-  }, []);
+  }, [dispatch, trpc.transferCredits.getUncategorizedTransfers]);
 
   if (courses.length === 0) {
     return null;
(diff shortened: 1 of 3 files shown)


CadenLee2 (commented):
My uncategorized transfers did show up with the right amount of units (including 0 for null).

However, when trying this out on staging, after clicking the trash can, items were not immediately shown to be deleted--nothing seemed to happen until I reopened the menu/reloaded the page. I assume this is because the code in the `useEffect` to update from the db was not called again after deleting the item.

I'd also like to propose the idea of the uncategorized credits section only showing up if it is not empty, because when there are no items it's a bit confusing to look at. At least there could be a message that says "No items here" or something (will mention this in the meeting)

zcharz:
> However, when trying this out on staging, after clicking the trash can, items were not immediately shown to be deleted--nothing seemed to happen until I reopened the menu/reloaded the page. I assume this is because the code in the `useEffect` to update from the db was not called again after deleting the item.

totally, i'm still working on that part and should have it figured out today/tmw, thanks! you're definitely right that it's cuz its not updating the course array and not re-rendering the components

site/src/pages/RoadmapPage/transfers/UncategorizedCreditsSection.tsx
```diff
@@ -1,14 +1,50 @@
...
 
 const UncategorizedCreditsSection: FC = () => {
+  const courses = useAppSelector((state) => state.transferCredits.uncategorizedCourses);
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
+    trpc.transferCredits.getUncategorizedTransfers.query().then((response) => {
+      dispatch(setUncategorizedCourses(response));
+    });
+  }, []);
```
Awesome-E, line 32: Lint warning here

Awesome-E (commented):
Address the lint warning, otherwise looks good to me! :octocat:

Awesome-E (approved):
:basecampy:

PR #669, opened 2025-04-30
Redesign the Course/Professor Info Page

Awesome-E:
Currently, it's not great. So, this PR makes it better. 🙃

This PR redesigns the layout and make the styles more consistent throughout the Individual Course and Professor page.

## Screenshots

All screenshots are taken at the same Zoom level (slightly zoomed out)

Course Page Before|Course Page After
:-:|:-:
image|image
image|image
image|image
image|image

Mobile Before|Mobile After
:-:|:-:
image|image

Professor Page Before|Professor Page After
:-:|:-:
image|image
image|image
image|image
image|image

Mobile Before|Mobile After
:-:|:-:
image|image

Note: the "Add Review" form is being redesigned separately.

## Test Plan

- [ ] make sure everything looks as expected on staging

## Issues
N/A


CadenLee2 (commented):
Checked in staging (professor page, course page, including on mobile) and looks as expected

Code:
- I support the renaming of the red color
- Is `App.scss` really the best place for `.course-page` and `.professor-page`? I think it could be justified because we want to share the same css for these without copying it, but it seems a bit weird because `App.scss` in general has components used around the entire app, rather than for specific pages alone. Could it make sense to create a new css file that the course and professor pages can share?
- Otherwise no glaring issues in the big CSS changes

Awesome-E:
Was hoping to do less refactoring but yeah it does seem pretty beneficial to have a separate component to share those styles. Made those changes + re-requesting review.

Might also be worth double checking that no other pages have drastically changed as a result.

CadenLee2 (approved):
The refactor looks good codewise and on staging, nothing seems to have broken

PR #672, opened 2025-05-06
Fix roadmap side panel position

CadenLee2:
When zoomed out, the position of the roadmap side panel currently has empty space to the right because the main wrapper does not take up the whole page. This is fixed.

## Description

Changed the main wrapper width from 80% (likely an old value from when the sidebar width could change) to 100% (consistent with the other pages)

This fix is for main because the issue is from main, but it is the same issue that causes the transfer credits redesign to look weird on big screens, so this will fix that problem as well (confirmed via inspect element).

## Screenshots

Taken at 30% zoom, but tested at all zoom levels, and identical to before for most normal zoom levels.

| Comparison | |
| --- | --- |
| Before: | image |
| After: | image |

## Test Plan

- [x] Looks good on staging

## Issues

The issue does not seem to exist

Closes #


zcharz (approved):
lgtm 👍

Awesome-E (approved):
The reason for originally restricting the width had something to do with the tiles in the roadmap (and the overall roadmap width or something) like this:
image

However, I think that logic is still broken in its own way, and it's way less obviously broken than the sidebar not sticking to the side. So, I will approve this PR and have the other one be fixed separately.

PR #673, opened 2025-05-07
Add Courses Section for Redesigned Transfers

### LINKED ISSUE #632: User should be able to add UCI courses that they received credit for
A user should be able to transfer equivalent course credit.

- There should be a header/text describing this section
- There should be a `react-select` input box that lets you type in course names. Search should occur on these – consider reading docs on React-Select's [asynchronous setup](https://react-select.com/async)
- Selecting a course from the input box should add a tile similar to the one labeled "COMPSCI 122A" in the design below and have a default unit count equivalent to if taken at UCI (i.e. 4 for ICS 33 and 6 for ICS 53)
- The edit button should allow you to change the number of units. **This "Editable Units" field should be its own component so that it can be reused for AP Exams.**

Relevant section from the Figma:

Image

ethank2222:
## Description

- Added TRPC routes to load, add, delete, and modify transfer courses
- Modified frontend to allow adding of valid courses
- Modified global state to reflect changes

## Screenshots

image

## Test Plan

- Go to Staging instance
- Add a course
- Reload the page to ensure the class has been saved
- Update the units, and reload again
- Remove the course, and reload again

## Issues

Closes #632

### DIFF
--- a/site/src/store/slices/transferCreditsSlice.ts
+++ b/site/src/store/slices/transferCreditsSlice.ts
@@ -1,17 +1,44 @@
 import { createSlice, PayloadAction } from '@reduxjs/toolkit';
 
+export interface TransferredCourse {
+  courseName: string;
+  units: number;
+}
+
 export const transferCreditsSlice = createSlice({
   name: 'transferCredits',
   initialState: {
     showTransfersMenu: false,
+    transferredCourses: [] as TransferredCourse[],
   },
   reducers: {
     setShowTransfersMenu: (state, action: PayloadAction<boolean>) => {
       state.showTransfersMenu = action.payload;
     },
+    addTransferredCourse: (state, action: PayloadAction<TransferredCourse>) => {
+      state.transferredCourses.push(action.payload);
+    },
+    removeTransferredCourse: (state, action: PayloadAction<string>) => {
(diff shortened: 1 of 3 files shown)


site/src/store/slices/transferCreditsSlice.ts
```diff
@@ -1,17 +1,44 @@
 import { createSlice, PayloadAction } from '@reduxjs/toolkit';
 
+export interface TransferredCourse {
+  courseName: string;
+  units: number;
+}
```
Awesome-E, line 6: We might want to consider inferring types directly from Drizzle schema to avoid defining these interfaces, but in the scope of this PR it's not a big deal

Awesome-E (approved):
only cause I worked with you on a lot of it. @charlieweinberger for second review once your pr is ready for review

charlieweinberger (approved):
Code looks good to me, and the website seems to display and save data as expected.

site/src/store/slices/transferCreditsSlice.ts
charlieweinberger: Ellis and I both defined similar types in `types/src/transferCredits.ts` for APs/GEs, so this could potentially go in that file, although we both also needed AAPI types whereas Ethan doesn't here.

PR #674, opened 2025-05-07
Stored Major Specializations In Redux to prevent reloading

### LINKED ISSUE #658: Spec choices reload each time the user switches away and back to "Major" tab
Steps to reproduce:
1. Select a major with spec, i.e. Computer Science, in the Roadmap page.
2. Click on "Minor"
3. Click on "Major"

Expected: the specialization stays, and it doesn't appear to reload
Actual: the spec does stay, but it appears to reload the list.

DerrickThrower:
Stored Major Specializations In Redux to prevent reloading

## Description
when switching in between "major" and" "Minor" tab the list of specializations reloads which is not ideal.

1. Identified that the specializations were stored each time in an array with useState
2. instead of repopulating the useState array, I stored specializations in a redux array within the majorWithSpecializations interface to avoid repopulation
3. created a setMajorSpecs function that stored the majors specializations in the interface
4. refactored the interfaces references accordingly

## recording of functionality
https://github.com/user-attachments/assets/06c65fa2-e040-40c8-b141-cc919c4c6ac9

## Test Plan
1. Select a major with spec, i.e. Computer Science, in the Roadmap page.
2. Click on "Minor"
3. Click on "Major" 

Important: Make sure that if a spec is not chosen the list is not reloaded and specs save to proper major

Closes #658


site/src/pages/RoadmapPage/sidebar/MajorCourseList.tsx
```diff
@@ -37,12 +38,12 @@ interface MajorCourseListProps {
 const MajorCourseList: FC<MajorCourseListProps> = ({ majorWithSpec, onSpecializationChange, selectedSpecId }) => {
   const isDark = useContext(ThemeContext).darkMode;
   const [specsLoading, setSpecsLoading] = useState(false);
-  const [specOptions, setSpecOptions] = useState<{ value: MajorSpecialization; label: string }[]>([]);
   const [resultsLoading, setResultsLoading] = useState(false);
   const [open, setOpen] = useState(true);
 
-  const { major, specialization } = majorWithSpec;
+  const { major, selectedSpec: specialization, specializations } = majorWithSpec;
```
Awesome-E, line 44: Would also suggest renaming the variable here `specialization` to `selectedSpec` since the auto-rename you did in the other file doesn't change the names here.

```suggestion
  const { major, selectedSpec, specializations } = majorWithSpec;
```

fiveminus1:
.

site/src/pages/RoadmapPage/sidebar/MajorCourseList.tsx
DerrickThrower: latest commit should have done that now.  included changes to some instances of specialization to selectedSpec :3

DerrickThrower:
> .

hi 4

Awesome-E (approved):
tentatively approved, would like to get a second set of eyes to review though

site/src/pages/RoadmapPage/sidebar/MajorCourseList.tsx
```diff
@@ -95,7 +95,7 @@ const MajorCourseList: FC<MajorCourseListProps> = ({ majorWithSpec, onSpecializa
     major.id,
     majorWithSpec.requirements.length,
     selectedSpecId,
```
fiveminus1, line 97: I think the naming here gets a bit confusing with the new "selectedSpec" and subsequent "selectedSpec?.id". Thoughts on this @Awesome-E?

fiveminus1 (changes requested):
Overall looks good to me, great job @DerrickThrower !

fiveminus1 (commented):
Overall looks good to me, great job @DerrickThrower !

site/src/pages/RoadmapPage/sidebar/MajorCourseList.tsx
Awesome-E: ooh I think the logic here is just getting a bit hairy. I think that's better for another PR to fix though

fiveminus1 (approved):

Issue #353, opened 2023-09-20
Support for courses with variable units


ecxyzzy:
Before #351, courses with variable units are displayed as having only the minimum number of units; for instance, COMPSCI 199 can be counted for 2-5 units, but the courses view and Peter's Roadmap show that it fulfills only 2 units. #351 shows the unit range if a course has `minUnits != maxUnits`, but it will still fulfill only the minimum number of units in Peter's Roadmap since I'm not sure how best to proceed with such a change.

PR #678, opened 2025-05-08
Allow the user to manually complete `Marker` requirements

### LINKED ISSUE #637: Markers should be user-completable in Degree Requirements
The user should be able to click a checkbox next to "markers" if they have completed, and they should be stored in the database. Additionally, markers should be shown for GEs at the top-level.

Example: "VI. Language" should show `[ ] ... 3 years of H.S. Study` as an option (it's in the API response for GE Requirements), and if the user checks it, it should complete the language requirement and be saved to the current user.

Awesome-E:
## Description

Previously, there were two issues with degree requirements:
1. Marker requirements (i.e. Language via 4 years of high school study)  were completely hidden and could not be completed
2. Some course requirements did not have an explicit course list (i.e. upper div writing), so they could not ever be completed.

This PR addresses these issues by making both (1) Markers and (2) course requirements with zero courses into items that can be manually marked as complete by the user.

Markers are saved both to localStorage (when logged out) and to the user's account (when logged in), and they are shared between all individual 4-year plans.

## Screenshots

image image

## Test Plan

1. Go to the GEs tab while signed out
2. Open a group, i.e. Lower Div Writing, and click on a marker.
3. Reload the page. The marker should persist
4. Uncheck the marker you just ticked, and check a marker from a different GE group, i.e. upper div writing.
5. Reload and observe that the first marker is gone and that the second marker is present still.
6. Repeat steps 1-5 but while signed in
7. Explore majors, minors, and other GE groups to make sure markers are displayed properly.

## Issues

Closes #637 
(it might be worth generating a change log for this)


ecxyzzy (changes requested):
Code looks clean as per usual, not much to complain about there.

I definitely agree that we should generate a changelog for this and other features, because although it's a small (visual) feature, it is clearly a large (code-wise) feature and has a rather large impact on user experience as well. Hence why I'm requesting changes on this one, but I also have nits.

- I don't know if this is possible or even desirable but maybe 4 years HS lang should clear both GE-6 and -8 when checked (which is what it does in reality).
- I have a feeling upper div writing may be able to be cleared automatically but that would correspond with an API change. Not sure if this is something you want to create an issue for over there.
- Not relevant to this PR, but I am once again asking for some love on #353 🥲

CadenLee2 (commented):
## Testing
- Tested with several GEs and also found a marker that behaves as expected in a minor (Chinese Studies) (signed in and signed out)

## Opinions
Code looks good


I think at first glance, markers can look a bit confusing because of their indentation.

In a case like this it's fine because visually everything looks like it's nested "beneath" the "Complete 1 of the following series":
image

But in a case like this, where the indentation levels differ, it feels a bit unclear, as though the "complete 1" applies only to the two markers and the rest are an additional requirement.
image
 Of course when actually thinking about it everything makes sense, and once a user tries adding courses or checking the boxes I think they'll understand, but it did confuse me a bit.

I'm wondering if it would make sense to add an "or" in this case, to make things very clear from the start. Based on my initial reading this could imply that only the markers are "series" and the rest are "courses", but regardless of how someone sees it, it still leads to the same result.
image
This is very subjective though, so I'm not sure.

Awesome-E:
> I don't know if this is possible or even desirable but maybe 4 years HS lang should clear both GE-6 and -8 when checked (which is what it does in reality).

Yes if you can provide a common ID from API about markers. Otherwise, it requires a jank implementation because the two marker strings are structured inconsistently (because of course they are)

> I have a feeling upper div writing may be able to be cleared automatically but that would correspond with an API change. Not sure if this is something you want to create an issue for over there.

Yes that's absolutely correct. You will need to return some sort of RegEx or method to match the courses, and we can automatically detect that. But this PR deliberately converts 0-course-list course requirements into markers so that the user can at least complete them for now.

> but I am once again asking for some love on #353

Appreciate the reminder because I forgot it after spare cycles again. This should be a lot more doable as an unintentional extra benefit of creating modular code (wow, shocking that components created to be reusable become reusable, didn't know that could happen in software development 🤯). However, it may require changing how the planner is stored 😭

---

- Change log will be added in this PR
- Indentation will be looked at again because the hierarchy of "series" is in fact not fully developed.

Awesome-E:
Re-requesting review – To re-trigger the change log, run `delete localStorage.changelogSeen` in your console.

CadenLee2 (approved):
The redesign of the nested markers and groups is clear now, I can tell how it is meant to be structured. I checked some other requirements and they still look good as well.

I assume that the changes in padding/margins even when there are no markers are intentional, because it does look better?
| Before | After |
| --- | --- |
| image | image |

Awesome-E:
> I assume that the changes in padding/margins even when there are no markers are intentional

I don't think they were. Can check this though

Awesome-E:
I see. The reason for less space below is so it's more naturally grouped with the courses. It should be more consistent with how it is in other places:
image image

ecxyzzy (approved):
🚀

PR #679, opened 2025-05-08
Review Form Redesign

### LINKED ISSUE #623: Redesign the User Review Form
Redesign the review form to make it more consistent with other modals and to reduce confusion overall. It may be better to completely recreate the component from scratch, borrowing code from the existing one as needed. (i.e. consider doing that instead of starting with the existing one and tweaking until it's good)

Example Designs:
Image

This would also fix the whole review form submission state issue, ideally. Submitting a review should instead temporarily disable the form, wait for it to submit, then show a toast (instead of changing the content of the form popup) that tells the user the review was submitted successfully

fiveminus1:
## Description

The current review form is quite dated and doesn't match our current modals. I've redesigned the Review Form to use the ``ppc-modal`` style, following the mockup in #623. 

In addition, terms are now checked when reviewing a course - available year/quarter options are given based on the course/professor, and quarters should be matched to the selected year.

Finally, the new form removes the ``<ThankYouMessage />`` popup and displays a toast upon completion now, which should fix the submission state issue. 

(Note: styling changes to the modal container are TBD from @Awesome-E )

### Before
Course Review 
image
Professor Review 
image

### After
Course Review Page
image
Professor Review Page
image

## Test Plan
This form is displayed on the Course Page, Professor Page, and Your Reviews Page. 

- [ ] Ensure posting reviews on Course/Professor pages work as intended
- [ ] Ensure editing reviews on Course/Professor/Your Reviews pages work as intended
- [ ] Confirm valid terms are listed when editing a review from the Your Reviews page.

Note that when editing via Your Reviews, the terms available will be listed based on all terms taught by the professor, not necessarily the course. This is because "Your Reviews" are implied to be towards a professor rather than a course. In general, the nuances on available terms per page should probably be updated, but that might be best for a separate issue/PR. 

## Issues

Closes #623


fiveminus1:
Thanks for all the feedback @charlieweinberger! I've applied some of your suggested changes:
- Stars now use blue-primary.
- Fixed ReCaptcha border bug.
- Removed "(optional)" on Additional Details.
- Fixed checkbox bug for course details - let me know if you're still able to reproduce it in some other way. 
- Tags dropdown not showing should be fixed - previously selected tags just weren't being rendered.
- Extended Grade/Difficulty to use the full width - I'm a bit iffy on this one, since the values are only 1 char long and 2:1:2 for the top 3 rows looks off to me, but I agree that it doesn't look great, especially when a course/professor dropdown is not provided. Alternatively, could move course/professor up but I'm not sure if 1:2:2 looks good either..

To address some of your other points:
> the contrast between the modal's height and its width is jarring; similarly, the fact that you have to scroll the modal isn't great. 
- I'll leave that up to @Awesome-E 😅, but I believe we'll just cut the max-height of the modal, make it not stick to the top/bottom, and keep the scrolling.

> why are the dropdown menus for year, quarter, professor (or course), grade, and difficulty all the default dropdown, instead of the react-select dropdowns used throughout the app, including for the tags below?
- This was admittedly just an oversight when trying to reuse components from the original form. While I agree it looks significantly better, the switch would introduce some problems. From what I understand, react-select does not have native integration with Bootstrap forms, so the current feedback on missing values would not be available. I'm not sure if there's a clean solution to keeping the feedback, other than having to introduce some messier states/styling. In addition, the react-select dropdown styling would have to be tweaked (too little items are displayed at once, making common actions like selecting a professor or grade less convenient). 
> I'd recommend replacing the "Summer10wk" option in the dropdown with just "Summer"
- Would you recommend only replacing "Summer10wk" and leaving "Summer1/2"? 


--
I believe the following changes might be best fitted for separate PRs - this one's mostly to clean up styling and make the modal more uniform with other ones.
> Is it too much to, in the professor/course taken dropdown, only show the professors/courses who were offered in the specific year & quarter that are selected above? That's probably unnecessary, but slightly more applicable than showing all courses.
- I think this would make sense. Unfortunately, it might get a bit hairy with how professor/course/terms props are passed into ``ReviewForm`` on different pages..

> After editing a review, should the edit label say "(edited on 5/9/2025)", instead of "(edited 5/9/2025)", to be more consistent with "Posted on: January 14, 2025"? Note the word "on" in both "Posted on" and my requested change "edited on".
- I think "on" is redundant here. I feel like I commonly see "edited" rather than "edited on" on other platforms. If this is to be changed though, might fit best in an bigger change/redesign for the Review/SubReview component(s)..it's also a bit odd that the title displays the ucinetid of the professor rather than the name. "Would Take Again" is also displayed separately from "Requires Textbook" and "Mandatory Attendance" on those components.

> Is it possible to merge reviews from cross-listed courses into one label? I.e. for reviewing shindler, instead of allowing me to select either "COMPSCI 162" or "LSCI 102" in the course taken dropdown, can it say "COMPSCI 162 / LSCI 102", since they're the same class and therefore it seems to me like they should have the same reviews?

> Similarly, are we sure that the tags should remain constant between courses and professors?

charlieweinberger:
@fiveminus1 Other than the 1 response below and a new bug below that, I agree that what you have right now (aka what you had yesterday + the few changes of mine you added) is good enough for this PR. A lot of what I suggested can either be a separate PR or is fine the way it is now. This is a huge improvement over the old review form, great work!

--------

> This was admittedly just an oversight when trying to reuse components from the original form. While I agree it looks significantly better, the switch would introduce some problems. From what I understand, react-select does not have native integration with Bootstrap forms, so the current feedback on missing values would not be available. I'm not sure if there's a clean solution to keeping the feedback, other than having to introduce some messier states/styling. In addition, the react-select dropdown styling would have to be tweaked (too little items are displayed at once, making common actions like selecting a professor or grade less convenient).

Hmm... if that's the case, if it's too complex to switch those dropdowns to use react-select, then should we change the tag dropdown to also not use react-select, just so every dropdown in the review form is consistent?

--------

New bug: the textarea for additional details has no height limit, and can be increased in size behind/below the elements below it.

Screenshot 2025-05-09 at 10 18 23 PM

charlieweinberger (changes requested):
See comment above

Awesome-E:
> the contrast between the modal's height and its width is jarring; similarly, the fact that you have to scroll the modal isn't great. Unfortunately, I can't easily come up with a great solution to this problem.

You will have to scroll the review form, but regardless of solution, you cannot prevent this from happening. The width and height do need some polishing, but as Jeremiah mentioned, I will address that.

> why are the dropdown menus for year, quarter, professor (or course), grade, and difficulty all the default dropdown, instead of the react-select dropdowns used throughout the app, including for the tags below?

React-select was only applied to the tags since tags are the only multiselect element. We discussed this as the solution for both dev simplicity and the fact that the react select menu has more scroll bars than native select, which is annoying to deal with when the form already scrolls.

> the chevrons to the default dropdown menus are too far to the right, almost touching the right border of the dropdown, looks a bit strange

Bootstrap moment – one of many things that would be improved with unifying components to MUI

> I'd recommend replacing the "Summer10wk" option in the dropdown with just "Summer"

Out of scope since this PR is to rewrite the component logic rather than the data/functional logic.

> Is it too much to, in the professor/course taken dropdown, only show the professors/courses who were offered in the specific year & quarter that are selected above? That's probably unnecessary, but slightly more applicable than showing all courses.

Yes, it is out of scope for this PR as per my discussion with Jeremiah. Would be a nice feature to see in the future though!

> Is it possible to merge reviews from cross-listed courses into one label? I.e. for reviewing shindler, instead of allowing me to select either "COMPSCI 162" or "LSCI 102" in the course taken dropdown, can it say "COMPSCI 162 / LSCI 102", since they're the same class and therefore it seems to me like they should have the same reviews?

We don't have equivalent course data yet

> I'd recommend extending the grade & difficulty dropdowns to be the full width of the modal, similar to year & quarter.

This could work, but a square actually might look awakward. A single digit/letter usually indicates a different type of input, which is what I saw in the screenshot

> it'd be cool for the rating stars to be our primary blue color, instead of white

Not sure I agree with this, at least not for hover and non-hover. I don't completely disagree either though

> "course details" is potentially out-of-place when reviewing a professor, or at least should be renamed to "professor details". Similarly, are we sure that the tags should remain constant between courses and professors?

Yeah, I guess that makes sense for the header. However, tags should not – recall that reviews require both prof and course.

> In the header for additional details, the "(optional)" text looks strange IMO, do we really need that optional warning?

(optional) is typically more user-friendly than red stars, and it does signify that the user doesn't have to write an extended description

> For some reason the outline of the the captcha is missing on it's right side, makes the captcha look a bit off.

I can handle that

> When editing a review in "Your Reviews", I'm unable to add any tags - there are no elements that dropdown, and typing in the dropdown text input does nothing (see image below). This seems to be because the review I'm trying to edit currently has 4 tags - 3 selected, plus "requires textbook", which is one of the course details but is for some reason showing up as a tag. The same thing applies to the "mandatory attendance" course detail, although the third course detail "would take again" is notably not a tag, instead being listed above/higher up in the review. I'm perfectly able to update the tags for a review that currently only has 3 tags.

This might be on me as part of redesigning the course page. I would have to take a look at that.

> After editing a review, should the edit label say "(edited on 5/9/2025)", instead of "(edited 5/9/2025)", to be more consistent with "Posted on: January 14, 2025"? Note the word "on" in both "Posted on" and my requested change "edited on".

Also related to my PR, but I tried that, and having consistent formatting actually looked less readable to me. Quite strange, I know, but that's not for this PR

fiveminus1:
@charlieweinberger Text area bug should be fixed now, thanks for catching that - and no worries at all, it's always nice to have another set of eyes look at the design and I appreciate the time you took to thoroughly test everything!

Awesome-E (changes requested):
Captcha should not show when editing the review

site/src/component/Review/SubReview.tsx
```diff
@@ -256,6 +258,7 @@ const SubReview: FC<SubReviewProps> = ({ review, course, professor }) => {
           closeForm={closeReviewForm}
           show={showReviewForm}
           editing
+          terms={sortTerms(course?.terms || (professor ? getProfessorTerms(professor) : []))}
```
Awesome-E, line 261: this line is quite long, might be worth it to define this as a variable earlier and just pass the variable in

site/src/component/ReviewForm/ReviewForm.scss
```diff
@@ -1,139 +1,64 @@
...
   }
 }
+
+// monkey patch for recaptcha border bug
+.g-recaptcha {
+  overflow: hidden;
+  height: 76px;
+  border-radius: 4px;
+  width: 302px;
+}
```
Awesome-E, line 64: Originally, this was just for dark mode and for captchas in the review form. Now, it's global for all captchas. If that's the case, it should probably be moved to `App.scss`, otherwise, it can stay in this file but as a nested rule inside of a review form

site/src/component/ReviewForm/ReviewForm.tsx
Awesome-E, line 1: still quite involved but doesn't look too bad

site/src/component/Review/SubReview.tsx
fiveminus1: Defined as variable in latest push

site/src/component/ReviewForm/ReviewForm.scss
fiveminus1: From what I understand, the border bug does apply to all ReCAPTCHAs (though this is currently the only captcha in the app), and it does seem to add a slight shadow if the patch is not applied in light mode, so I've opted to include this in ``App.scss``. See latest push

site/src/component/ReviewForm/ReviewForm.tsx
fiveminus1: Yeah, unfortunately I think adding the terms checking just made it worse, and it's already a pretty long component. Would welcome any suggestions though

Awesome-E (approved):
HELL YEAH

Awesome-E:
@charlieweinberger need you to approve so we can merge this!

charlieweinberger:
1. Should "Post as Anonymous" be shown when editing a review? Currently, if you check this box after posting a non-anonymous review, nothing happens.

2. There's a slight bug in selecting the year/quarter when editing a review. When creating a review, if you have a year and a quarter selected, then try to change the year to a year that the course wasn't offered in that quarter, the quarter will reset to "select" and the valid quarter options will be shown in the quarter dropdown. However, when editing a review, if you do the same action, instead of the quarter dropdown resetting to "select", the year dropdown will stay at its current value. This means that it's impossible to edit the review date to be certain dates. The fix should be to change the edit review action to mirror the create review action. However, this bug is relatively minor, given that most people wont be editing review dates, so if it's difficult to fix, then this can just be a separate PR, it shouldn't hold back the majority of changes.

Other than those two suggestions, looks good to me!

fiveminus1:
@charlieweinberger 

1. Yes, it should be displayed when editing. You can check this to confirm, but it appears to function the same with the current form - the review is updated, but the frontend component itself is not. If you refresh, it should be updated with the Anonymous change. A bit odd that other details on the review are immediately refreshed, but that's outside of the scope of this PR anyways. 

2. Again, let me know if this doesn't address your bug, but this is because editing a review validates terms based on the professor rather than the course. So, if you write a review for say, ICS 46 by Thornton in 2025 Fall, you won't be able to edit the review because it shouldn't be a valid review in the first place. If you try editing a review that does have a valid term w/ valid professor, editing the dropdown should work.

Of course, this isn't great because you also can't edit the professor via Your Reviews, but I believe this would be best fitted for a separate PR as it requires more validation and as you said, it's not a very common action.

charlieweinberger:
That all makes sense, I see what you mean for both points. I agree those are separate PRs. All good to merge this PR then!

charlieweinberger (approved):

PR #682, opened 2025-05-13
removed leftover semantic-ui code

charlieweinberger:
## Description

This PR removes code that I believe was left over from Semantic UI. Reasons I think it's Semantic UI:

- There's literally a comment that explicitly says "semantic-ui", heavily implying this code is from semantic ui.
- The CSS classes deleted - `ui`, `dropdown`, `menu`, and `item` - aren't used anywhere else in our code.
- The method of adding custom styling to default components (which is what I believe this code is trying to do) used in `SearchPopus.scss` is different from how bootstrap does it; bootstrap does `dropdown-item`, not `.dropdown.menu.item`, as I understand it.
- As far as I can tell, the styles (background/border colors) in `SearchPopus.scss` aren't applied anywhere. I tried removing other styling that be overwriting this code, but I was completely unable to get these styles to appear at all. I'm 99.9% confident that this is because the styles are attempting to overwrite semantic ui's default components, as mentioned above, but since we don't have any semantic ui components, the code isn't doing anything.

Because this code is leftover from Semantic UI and isn't being used, I propose we remove it.

## Screenshots

Removing this code results in no visual change.

## Test Plan

- [ ] Verify that these styles are not being applied anywhere.
- [ ] Verify that this code is not being used anywhere. 

## Issues

This PR doesn't directly close any issue, but is related to issue #14 and pull request #660.


charlieweinberger:
@ethank2222 Is there anything that I'm missing/misunderstanding here, related to this PR?

charlieweinberger:
@Awesome-E Not sure if you did anything (thanks if you did!), but all github checks are passing now.

Awesome-E (approved):
ok yeah it does look like it's unused now

PR #690, opened 2025-05-14
Save new transfers to `localStorage`

Awesome-E:
omg finally.

## Description

This PR makes it so that the new redesigned transfers experience works both when logged in and logged out.
- Local Roadmap gets automatically upgraded when the page is upgraded, so transfers are always in the new format
- Adding, removing, updating a transfer in the new menu will save locally if logged out (and still saves to DB if logged in).
- Added basic backend logic to `api/scripts/transfersDataConversion.ts`, though this is **planned to be refactored** in Caden's next PR to implement auto-categorize transfers from transcripts
- Refactors the `<Planner/>` component to not be in one giant `useEffect()`. Instead offloads data loading and processing to a helper component (which shows the out of sync modal when action related to data is needed)

## Screenshots

N/A

## Test Plan

The below tests most cases, but may not be _fully_ comprehensive.

Provided Script:
```js
transfers = [{"name":"AP Calculus BC","units":5},{"name":"MATH 2B","units":4},{"name":"BAD","units":0}]
localStorage.clear()
localStorage.changelogSeen = '5/11/2025'
localStorage.roadmap = `{"timestamp":"2025-05-30T06:04:55.895Z","planners":[{"name":"Peter's Roadmap","content":[{"startYear":2025,"name":"Year 1","quarters":[{"name":"Fall","courses":[]},{"name":"Winter","courses":[]},{"name":"Spring","courses":[]}]}]}],"transfers":${JSON.stringify(transfers)}}`
```

1. Run the provided script while signed out, then reload the page. This simulates having legacy transfers before and loading the new experience for the first time. The transfers should show up in the new menu.

2. Reload the page again. The new transfers should persist.

3. Delete all of your roadmaps and new transfers from the dev databases (you can use a separate account), `localStorage.clear()`, then log in.

4. Run the provided script to add basic old transfers to `localStorage`.

5. Reload the page while logged in. It should auto-populate new transfers

6. Reload again and make sure data is still in the account.

7. Change the number of units for one of the transfers.

8. Repeat step 2, and reload the page. This time, there should be a popup to use this device vs my account roadmap. Choose "My account".

9. The changes from step 7 should still be there.

10. Reload again, but choose "This Device". Now, the changes from step 7 should be gone.

11. Reload after (this reload should not prompt you to choose device vs account). The changes from step 7 should still be gone.


site/src/helpers/planner.ts
```diff
@@ -196,6 +186,64 @@ export function convertLegacyLocalRoadmap(roadmap: SavedRoadmap | LegacyRoadmap)
   }
 }
 
+// Upgrading Transfers
+async function saveUpgradedTransfers(roadmapToSave: SavedRoadmap, transfers: TransferData[]) {
+  // Avoid issues with double first render
+  if (!transfers.length || localStorage.roadmap__versionKey === '3') return false; // nothing to convert
+
+  localStorage.roadmap__versionKey = '3';
```
CadenLee2, line 194: Probably a good idea to put the `'3'` in a variable so the purpose is more clear and it would only have to be modified in one place in the future

site/src/helpers/planner.ts
```diff
@@ -196,6 +186,64 @@ export function convertLegacyLocalRoadmap(roadmap: SavedRoadmap | LegacyRoadmap)
...
+  return roadmapWithoutLegacyTransfers;
+}
+
+export const saveRoadmap = async (isLoggedIn: boolean, planners: SavedPlannerData[], showToasts: boolean = true) => {
+  const roadmap: SavedRoadmap = { timestamp: new Date().toISOString(), planners, transfers: [] };
+  localStorage.setItem('roadmap', JSON.stringify(roadmap));
+
+  const showMessage = showToasts ? spawnToast : (str: string) => str;
+
+  const SAVED_LOCALLY_MESSAGE = 'Roadmap saved locally! Login to save it to your account.';
```
CadenLee2, line 238: I think "Log in" is more grammatically correct

CadenLee2 (commented):
## Testing
- Logged in and changed some things
- Logged out and the transfer credits locally matched what had been in the account
- Logged back in and they work fine
- Followed the steps through to step 7.
- At step 8, not fully sure what was intended; I just reloaded the page twice, which (correctly) didn't show the popup. Did you mean "repeat step 1" or "repeat step 3"?
- Note: I think when I changed the units on one of the transfers and logged out, the localStorage wasn't updated the first time (it retained the old value)? I can't replicate this, though, so I may not have seen it properly
- Issue: I added an AP exam (AP 2D Art) while logged out, then immediately logged in. The popup didn't show and the added AP exam didn't appear. The same happened if I reloaded in between these steps. Interestingly, `localStorage` still shows AP 2D Art, even though it's not appearing under AP Exam Credits while logged in. However, once I log out (even without reloading the page), localStorage loses AP 2D art
  - Cause: I think this is because my roadmap timestamp didn't change after adding the transfer (since I didn't actually modify the roadmap itself). When changing transfers while logged out AND changing/saving the roadmap, only then does the popup appear when I sign in
  - To handle this, should we update the roadmap timestamp (or add a new separate timestamp for transfers, but that'd be complicated) when transfers are changed?
- When saving roadmap as mentioned above, popup "this device" seems to work as expected, it brings in the changes
- Popup "my account" seems to work as expected, it ignores the local changes

site/src/helpers/planner.ts
Awesome-E: Yep! this was copied from a different section but changing that!

site/src/helpers/planner.ts
Awesome-E: put this in an enum so it's more obvious as to what it is

CadenLee2 (approved):
Since we re-thought the idea of roadmap saving and transfers when logged out (the issue mentioned above), the current logic should be fine for now. In the future we should probably deal with the fact that the user's roadmap data persists after they log out since that was the source of some of the confusion.

Otherwise, looks good

api/src/controllers/transferCredits.ts
```diff
@@ -111,54 +120,104 @@ const transferCreditsRouter = router({
...
+          courseName,
+          units,
+          userId: ctx.session.userId!,
+        }));
+        const addOtherQuery = db.insert(transferredMisc).values(rows);
+        dbQueries.push(addOtherQuery);
       }
 
-      await db.delete(transferredMisc).where(and(...conditions));
+      await Promise.all([dbQueries.map((q) => q.execute())]);
```
ecxyzzy, line 220: Might be more appropriate to wrap this in a transaction since this mutation should be one logical step?

site/src/helpers/courseRequirements.ts
```diff
@@ -160,7 +160,6 @@ export function flattenSingletonGroups(requirements: ProgramRequirement[]): Prog
  * @param requirements The program requirements list
  */
 export function sortGroupRequirementsByType(requirements: ProgramRequirement[]): ProgramRequirement[] {
-  // console.log(requirements)
```
ecxyzzy, line 163: assuming this will be nuked when merged into main

ecxyzzy (commented):
Vibes seem fine, good work.

One non-blocking question re: db so didn't request changes, feel free to merge if that is a non-issue

site/src/helpers/courseRequirements.ts
Awesome-E: not sure what you mean since this is a line removal, but the commented console log will not be there

ecxyzzy: oh i'm fukcign stupid

don't do code review on 4 hours of sleep kids

api/src/controllers/transferCredits.ts
Awesome-E: Ideally yes, but I think this should not be an issue since we always update when there's conflict (and the schema prevents invalid values from entering in)

PR #691, opened 2025-05-15
Add Selected rewards table to the database

ellisl10:
## Description

Added selectedApReward to database schema and migrated

## Test Plan

Check that the table exists, be able to add a row, try when user_id and exam_name should fail, deleting user should also delete the row


Awesome-E (approved):
Looks good to me. @fiveminus1 if you want to take a second look, the general idea is that files inside `api/drizzle` are auto-generated, and the `schema.ts` file is changed manually to update how a table is structured (from within our code).

Feel free to look through these and offer any potential thoughts on the schema changes.

fiveminus1 (approved):

fiveminus1:
Looks good to me!

Issue #688, opened 2025-05-13
There should be an indication of how a credit was completed in Degree Requirements


Awesome-E:
(once #681 is merged...)

Currently, a course is struck out if it's counted in Degree Requirements.

However, there are a few issues:

1. Because courses are struck out for transfer courses and APs, it may sometimes be confusing where a course is completed (since it's not immediately visible in the roadmap. There should be a badge on the course to indicate that it's completed outside of the roadmap (i.e. a transfers icon or "AP" text)

2. You can add GE course credit, and completion for GEs will work. However, this means that (for example) "3 courses Category II" can be completed by one course in roadmap and 2 via "Transfer GEs". the Label (3 courses Category II) should have an indicator of how many GE courses were applied from transfers rather than from a course within that list being counted (counted can mean roadmap, course transfer, and AP exam)

Afterwards, it should be clear in that struck-out courses will either be in the roadmap or have a badge, and GEs completed with less than the required amount of courses being struck out (i.e. GE II requires 3 courses but it can have 2 struck out courses and 1 transferred GE-II course credit) should have an indication that the non-course credit exists and is from transfer GEs.

PR #692, opened 2025-05-16
Show the source of a transferred credit in degree requirements

charlieweinberger:
## Description

If a transferred course/AP exam clears a course in degree requirements, then the badge for that course will now have a marker in the top right corner that shows it was sourced from a transferred course/AP exam.

~~There are many ways to implement this change in the code. In this PR, I chose to create duplicate types of existing types (i.e. `TransferredCourseWithType` instead of `TransferredCourse`, and `CompletedCourseSetWithType` instead of `CompletedCourseSet`), so that I could use those duplicate types in the necessary files (`ProgramRequirementsList.tsx` and `transferCredits.ts`), but not have to change any other files that used the existing types. I'm open to doing this PR another way, since there are a couple options for where exactly to start tracking the type of each transferred course.~~ Instead of creating new types, this PR now modifies existing code to have the optional `transferType` property.

~~The actual design of the badge is also possible to change. Should it have a background at all, or should it just be text? What should we do for the size/color of the text/background?~~ See screenshot.

Also, note that in the final version, transferred courses will use an arrow left/right icon, but until @ecxyzzy finishes replacing bootstrap icons with MUI icons, this PR will just stick with the word "course".

## Screenshots

Before vs. after:
Screenshot 2025-05-17 at 7 27 12 PM    Screenshot 2025-05-27 at 10 51 09 PM

## Test Plan

- [ ] Verify the overlay shows for the correct course ("AP" for AP exams, the double arrow icon for transferred courses, and nothing for other courses).
- [ ] Verify that the overlay help tooltip shows on hover.
- [ ] Verify that no other code is negatively affected by this PR's changes.

## Issues

Closes part 1 of issue #688

### DIFF
--- a/site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.scss
+++ b/site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.scss
@@ -149,6 +149,7 @@
   font-size: 13px;
   border-radius: 4px;
   cursor: grab;
+  position: relative;
 
   a {
     color: inherit;
@@ -163,6 +164,19 @@
     text-decoration-thickness: 2px;
     opacity: 0.75;
   }
+}
+
+.source-overlay {
+  position: absolute;
+  top: 0;
+  right: 0;
+  transform: translate(25%, -25%);
+  color: white;
+  font-size: 8px;
+
+  background-color: var(--blue-primary);
+  padding: 1px 4px;
+  border-radius: 4px;
 }
 
 body[data-theme='dark'] {
--- a/site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.scss
+++ b/site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.scss
@@ -173,7 +173,7 @@
   transform: translate(25%, -25%);
   width: 20px;
   height: 20px;
-  border-radius: 50%;
+  border-radius: 20px;
   background-color: var(--text-secondary);
   color: var(--gray-blue);
   display: flex;
--- a/site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.scss
+++ b/site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.scss
@@ -181,6 +181,10 @@
   align-items: center;
   font-size: 10px;
   cursor: help;
+
+  .MuiSvgIcon-root {
+    font-size: 16px;
+  }
 }
 
 body[data-theme='dark'] {
--- a/site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.scss
+++ b/site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.scss
@@ -171,12 +171,16 @@
   top: 0;
   right: 0;
   transform: translate(25%, -25%);
-  color: white;
-  font-size: 8px;
-
-  background-color: var(--blue-primary);
-  padding: 1px 4px;
-  border-radius: 4px;
+  width: 20px;
+  height: 20px;
+  border-radius: 50%;
+  background-color: var(--text-secondary);
+  color: var(--gray-blue);
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  font-size: 10px;
+  cursor: help;
 }
 
 body[data-theme='dark'] {
--- a/site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.tsx
+++ b/site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.tsx
@@ -25,15 +25,14 @@ import { useAppDispatch, useAppSelector } from '../../../store/hooks';
 import { Spinner } from 'react-bootstrap';
 import { ProgramRequirement } from '@peterportal/types';
 import { setGroupExpanded, setMarkerComplete } from '../../../store/slices/courseRequirementsSlice';
-import { TransferredCourse } from '../../../store/slices/transferCreditsSlice';
 import { getMissingPrerequisites } from '../../../helpers/planner';
 import { useClearedCourses } from '../../../hooks/planner';
-import { useTransferredCredits } from '../../../hooks/transferCredits';
+import { useTransferredCredits, TransferredCourseWithType } from '../../../hooks/transferCredits';
 import { useIsLoggedIn } from '../../../hooks/isLoggedIn';
 
 interface CourseTileProps {
   courseID: string;
-  completedBy: TransferredCourse['transferType'] | 'roadmap' | null;
+  completedBy: TransferredCourseWithType['transferType'] | 'roadmap' | null;
   /** The timestamp at which the course data is requested to load */
   dragTimestamp?: number;
 }
--- a/site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.tsx
+++ b/site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.tsx
@@ -30,13 +30,28 @@ import { useClearedCourses } from '../../../hooks/planner';
 import { useTransferredCredits } from '../../../hooks/transferCredits';
 import { useIsLoggedIn } from '../../../hooks/isLoggedIn';
 
+interface CompletedCourseSetWithType {
+  [k: string]: {
+    units: number;
+    transferType: 'AP' | 'Course';
+  };
+}
+
 interface CourseTileProps {
   courseID: string;
   /** The timestamp at which the course data is requested to load */
   dragTimestamp?: number;
   taken?: boolean;
+  transferredByAP?: boolean;
+  transferredByCourse?: boolean;
 }
-const CourseTile: FC<CourseTileProps> = ({ courseID, dragTimestamp = 0, taken }) => {
+const CourseTile: FC<CourseTileProps> = ({
+  courseID,
+  dragTimestamp = 0,
+  taken,
+  transferredByAP,
+  transferredByCourse,
+}) => {
   const [courseData, setCourseData] = useState<string | CourseGQLData>(courseID);
   const [loading, setLoading] = useState(false);
   const isMobile = useIsMobile();
@@ -90,13 +105,20 @@ const CourseTile: FC<CourseTileProps> = ({ courseID, dragTimestamp = 0, taken })
 
   return (
     <div className={className} {...tappableCourseProps} style={{ fontSize }}>
+      {transferredByAP && <div className="source-overlay">AP</div>}
+      {transferredByCourse && <div className="source-overlay">Course</div>}
       <CourseNameAndInfo data={courseData} openPopoverLeft popupListener={handlePopoverStateChange} alwaysCollapse />
       {isMobile && loading && <Spinner animation="border" />}
     </div>
   );
 };
 
-const CourseList: FC<{ courses: string[]; takenCourseIDs: CompletedCourseSet }> = ({ courses, takenCourseIDs }) => {
+interface CourseListProps {
+  courses: string[];
+  takenCourseIDs: CompletedCourseSet;
+  transferredCourseIDs: CompletedCourseSetWithType;
+}
+const CourseList: FC<CourseListProps> = ({ courses, takenCourseIDs, transferredCourseIDs }) => {
   const isMobile = useIsMobile();
   const [timestamps, setTimestamps] = useState<number[]>(new Array(courses.length).fill(0));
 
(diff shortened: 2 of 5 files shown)


site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.tsx
```diff
@@ -30,13 +30,28 @@ import { useClearedCourses } from '../../../hooks/planner';
 import { useTransferredCredits } from '../../../hooks/transferCredits';
 import { useIsLoggedIn } from '../../../hooks/isLoggedIn';
 
+interface CompletedCourseSetWithType {
+  [k: string]: {
+    units: number;
+    transferType: 'AP' | 'Course';
```
Awesome-E, line 36: Make this property optional instead of drilling two props all the way down. Then, for non-transferred courses, simply leave it as undefined

site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.tsx
```diff
@@ -307,12 +372,17 @@ const ProgramRequirementsList: FC<RequireCourseListProps> = ({ requirements, sto
     .flatMap((quarter) => quarter.courses)
     .map((course) => [course.id, course.minUnits]);
   const transferCourseMap = transferredCourses.map((t) => [t.courseName.replace(/\s/g, ''), t.units ?? 0]);
+  const transferCourseMapWithType = transferredCourses.map((t) => [
+    t.courseName.replace(/\s/g, ''),
+    { units: t.units ?? 0, transferType: t.transferType },
+  ]);
```
Awesome-E, line 378: as per previous comment:

```suggestion
    .map((course) => [course.id, { units: course.minUnits }]);
  const transferCourseMap = transferredCourses.map((t) => [
    t.courseName.replace(/\s/g, ''),
    { units: t.units ?? 0, transferType: t.transferType },
  ]);
```

site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.tsx
```diff
@@ -307,12 +372,17 @@ const ProgramRequirementsList: FC<RequireCourseListProps> = ({ requirements, sto
...
 
   const takenCourseSet: CompletedCourseSet = Object.assign(
     {},
     Object.fromEntries(roadmapCourseMap),
     Object.fromEntries(transferCourseMap),
   );
+  const transferredCourseSet: CompletedCourseSetWithType = Object.fromEntries(transferCourseMapWithType);
```
Awesome-E, line 385: As per previous comment, just use the same variable

```suggestion
  const takenCourseSet: CompletedCourseSetWithType = Object.assign(
    {},
    Object.fromEntries(roadmapCourseMap),
    Object.fromEntries(transferCourseMap),
  );
```

site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.tsx
```diff
@@ -323,6 +393,7 @@ const ProgramRequirementsList: FC<RequireCourseListProps> = ({ requirements, sto
           key={i}
           storeKey={`${storeKeyPrefix}-${i}`}
           takenCourseIDs={takenCourseSet}
+          transferredCourseIDs={transferredCourseSet}
```
Awesome-E, line 396: Do not pass an extra prop through a several-layer nested component

```suggestion
```

site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.tsx
```diff
@@ -115,7 +137,14 @@ const CourseList: FC<{ courses: string[]; takenCourseIDs: CompletedCourseSet }>
...
     >
       {courses.map((c, i) => (
-        <CourseTile courseID={c} key={c} dragTimestamp={timestamps[i]} taken={c in takenCourseIDs} />
+        <CourseTile
+          courseID={c}
+          key={c}
+          dragTimestamp={timestamps[i]}
+          taken={c in takenCourseIDs}
+          transferredByAP={c in transferredCourseIDs && transferredCourseIDs[c].transferType === 'AP'}
+          transferredByCourse={c in transferredCourseIDs && transferredCourseIDs[c].transferType === 'Course'}
```
Awesome-E, line 146: Use one prop instead of 3
```suggestion
          completedBy={c in takenCourseIDs ? (takenCourseIDs[c].transferType ?? 'roadmap') : null}
```

Then change the prop type to be `TransferredCourseWithType['transferType'] | 'roadmap' | null`

site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.tsx
Awesome-E: Additionally, instead of repeating `'AP' | 'Course'`, you should read the type directly from `TransferredCourseWithType` after exporting:
```suggestion
    transferType?: TransferredCourseWithType['transferType'];
```

site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.tsx
```diff
@@ -90,13 +105,20 @@ const CourseTile: FC<CourseTileProps> = ({ courseID, dragTimestamp = 0, taken })
 
   return (
     <div className={className} {...tappableCourseProps} style={{ fontSize }}>
+      {transferredByAP && <div className="source-overlay">AP</div>}
+      {transferredByCourse && <div className="source-overlay">Course</div>}
```
Awesome-E, line 109: Source overlay should have a `title` attribute and ideally have `cursor: help;`

Screenshot 2025-05-16

Awesome-E (changes requested):

Awesome-E:
Alternative of what it could look like (this one also uses MUI icons)

image

ellisl10 (approved):

site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.tsx
charlieweinberger: Instead of making `transferType` within `TransferredCourseWithType` optional and then exporting it, can I just add `transferType` as an optional property of `TransferredCourse`, and then just use `TransferredCourse` instead of `TransferredCourseWithType`?

charlieweinberger:
Instead of creating any new variables/types/props/etc. for cleared courses, this PR now modifies the original variables/types/props/etc. to allow for the optional `transferType` property. I also updated the overlay as requested. Let me know what you think

site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.tsx
Awesome-E: I don't necessarily think this is the correct approach for one main reason: `TransferredCourse` is the interface that's used to store items in the slice, and the slice is ONLY for explicitly transferred courses, not rewarded courses from AP. We should not allow the ambiguity that maybe AP rewards should be allowed to be stored in the `transferCreditsSlice.transferredCourses: TransferredCourse[]` array, meaning that `TransferredCourse` should not allow that property.

Awesome-E: This is something we should do when consolidating types though when we refactor

Awesome-E:
The diff on this commit indicates that something is out of sync right now. Please update your branch from `main`, and then re-request review again.

charlieweinberger:
I'm running into a bunch of merge conflicts, something related to me starting this PR based on a partially-completed `redesigned-transfers` branch. I'm struggling to merge everything together properly right now. Can you take a look, or help me out at tmrw's meeting?

Awesome-E:
See response to your comment here: https://github.com/icssc/peterportal-client/pull/692#discussion_r2094589142
image

site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.tsx
charlieweinberger: Reverted back to using `TransferredCourseWithType`, although still using the modified `CompletedCourseSet`.

site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.scss
```diff
@@ -165,6 +166,23 @@
...
 }
 
+.source-overlay {
+  position: absolute;
+  top: 0;
+  right: 0;
+  transform: translate(25%, -25%);
+  width: 20px;
+  height: 20px;
+  border-radius: 50%;
```
Awesome-E, line 176: In the case that a similar pill is not the same width or height, this would cause it to be an ellipse rather than a pill. instead, use a fixed amount that is no less than half the height of the element:
```suggestion
  border-radius: 20px;
```

site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.tsx
```diff
@@ -88,15 +88,31 @@ const CourseTile: FC<CourseTileProps> = ({ courseID, dragTimestamp = 0, taken })
...
+    completedBy === 'AP' ? (
+      <p className="source-overlay" title={`${courseID} is cleared by an AP exam`}>
+        AP
+      </p>
+    ) : completedBy === 'Course' ? (
+      <p className="source-overlay" title={`${courseID} is cleared by a transferred course`}>
+        Course {/* to do: replace with MUI icon */}
+      </p>
+    ) : null;
+
```
Awesome-E, line 101: Consider putting this into its own component since the logic is more involved than I'd like for inline.

that component would looks something like:
```tsx
const SourceOverlay: FC<...> = ({ completedBy: ... }) => {
  if (!completedBy || completedBy === 'roadmap') return null;
  const title = 'Cleared by ' + (completedBy === 'AP' ? 'an AP Exam' : 'a transferred course')
  const icon = completedBy === 'AP' ? 'AP' : <Icon/>
  return <p className="source-overlay" title={title}>{icon}</p>
}
```

site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.tsx
```diff
@@ -88,15 +88,31 @@ const CourseTile: FC<CourseTileProps> = ({ courseID, dragTimestamp = 0, taken })
...
   return (
     <div className={className} {...tappableCourseProps} style={{ fontSize }}>
+      {sourceOverlay}
```
Awesome-E, line 104: Then simply call the component here:
```suggestion
      <SourceOverlay completedBy={completedBy} />
```

Awesome-E (changes requested):

charlieweinberger:
Nice suggestion to make `SourceOverlay` its own component, just pushed that change + the border radius thing + rebased w/ main.

Awesome-E:
We only needed to rebase that one time because the base branch changed and the commits got squashed. If you just update your branch later or main updates for a reason other than being the base branch of current PR, prefer to git merge

charlieweinberger:
Got it, I'll keep that in mind for the future. In this case I just did what my terminal prompted me to do, so maybe I have a setting set to prefer rebasing over merging. I'll look into it.

site/src/pages/RoadmapPage/sidebar/ProgramRequirementsList.tsx
```diff
@@ -26,19 +26,33 @@ import { ProgramRequirement } from '@peterportal/types';
...
 import ArrowDropDownIcon from '@mui/icons-material/ArrowDropDown';
+import SwapHorizOutlinedIcon from '@mui/icons-material/SwapHorizOutlined';
+
+interface SourceOverlayProps {
+  completedBy: TransferredCourseWithType['transferType'] | 'roadmap' | null;
+}
+const SourceOverlay: FC<SourceOverlayProps> = ({ completedBy }) => {
+  if (!completedBy || completedBy === 'roadmap') return null;
+  const title = `Cleared by ${completedBy === 'AP' ? 'an AP Exam' : 'a transferred course'}`;
+  const icon = completedBy === 'AP' ? 'AP' : <SwapHorizOutlinedIcon style={{ fontSize: 16 }} />;
```
Awesome-E, line 41: because I'm a hater of inline styles, I believe you should instead write `.MuiSvg-root { font-size: 16px; }` inside of the `.source-overlay` rule.

Awesome-E (commented):
pretty much good, just one quick thing

Awesome-E (approved):
.

Issue #689, opened 2025-05-13
User should be able to pick which reward they want from AP exams


Awesome-E:
Right now, if there's an OR reward like the AP Exam shown:

<img src="https://github.com/user-attachments/assets/c13c2e92-703b-4445-8735-b9aca5f395bd" width=400>

It will always pick the first one to count in your roadmap (i.e. `ECON 20B`)

We should make it so the user can choose which one to apply. This requires a DB change to store preferred rewards for an AP Exam transfer (new table with on delete cascade for the user id + AP Exam Name foreign key), and then will require additional frontend changes to accommodate these preferences.

PR #693, opened 2025-05-18
Update Zot4Plan AP Imports 

### LINKED ISSUE #686: Update Zot4Plan Imports to automatically categorize non-UCI credits
When a user goes to import items from Zot4Plan, it will currently put courses in "Other Transfers". A PR addressing this issue should do two things:

1. Current data should perform a batch course lookup to be sorted into courses.

2. Zot4Plan also provides an AP exams array in the response, which is currently ignored. We should do direct matches on these when we import from Zot4Plan.

The way data is returned and updated on the frontend and be saved should be similar to if not exactly the same as #685. This issue, however, does not require the creation of a new backend route, as Zot4Plan imports is already a backend route.

fiveminus1:
## Description

Currently, importing a roadmap from [Zot4Plan](https://zot4plan.com/) does not import AP Exams nor handle them in any way. This PR ensures that AP Exams from Z4P are:
- Matched to an exam name using existing matching logic (#666)
- Categorized into the `transferred_ap_exam` table
- Reflected in the frontend

## Demo

z4p_aps

## Test Plan

- [ ] Import a schedule from Zot4Plan that includes AP exams (you can try `test_ap_migration`, which should have APs Computer Science A, Biology, and Calculus BC).
- [ ] Verify the proper APs, scores, and units are imported under the Transfer Credits menu.
- [ ] Confirm the proper rows were added to the `transferred_ap_exam` table.
- [ ] Confirm existing added exams are not overwritten.

## Issues

Closes #686 

Note: the details in the issue are inaccurate - courses are not placed into "Other Transfers" and batch course lookup is not required.


api/src/controllers/zot4planimport.ts
```diff
@@ -8,6 +8,13 @@ import { publicProcedure, router } from '../helpers/trpc';
 import { zot4PlanImports } from '../db/schema';
 import { TRPCError } from '@trpc/server';
 import { SavedRoadmap, SavedPlannerData, SavedPlannerQuarterData, QuarterName } from '@peterportal/types';
+import { tryMatchAp, getAPIApExams } from '../helpers/transferCredits';
+
+interface userAPExam {
```
fiveminus1, line 13: This should probably be moved at some point since there's a duplicate in `controllers/transferCredits.ts`,

ecxyzzy:
jerb zot4plan

api/src/controllers/zot4planimport.ts
CadenLee2: Looks like it's also defined on the frontend in `transferCreditsSlice`, we should probably move it to the shared `types/src/transferCredits.ts`. I can do that in my transcript PR since I'm doing related refactoring, but we'll need to remember to reference it here once things get merged

api/src/controllers/zot4planimport.ts
```diff
@@ -168,9 +197,14 @@ const zot4PlanImportRouter = router({
...
-      const res = convertIntoSavedRoadmap(originalScheduleRaw, input.scheduleName, getStartYear(input.studentYear));
+      const savedRoadmap = convertIntoSavedRoadmap(
+        originalScheduleRaw,
+        input.scheduleName,
+        getStartYear(input.studentYear),
+      );
+      const apExams = await getApExamsFromZot4Plan(originalScheduleRaw);
       await db.insert(zot4PlanImports).values({ scheduleId: input.scheduleName, userId: ctx.session.userId });
-      return res;
+      return [savedRoadmap, apExams] as [SavedRoadmap, userAPExam[]];
```
Awesome-E, line 207: In React/TS, multiple return items are usually done in an object as opposed to an array.
```suggestion
      return { savedRoadmap, apExams }
```
The type should also be inferred correctly when it's done this way

site/src/pages/RoadmapPage/ImportZot4PlanPopup.tsx
```diff
@@ -9,28 +9,51 @@ import trpc from '../../trpc.ts';
...
   const [studentYear, setStudentYear] = useState('1');
   const [busy, setBusy] = useState(false);
   const allPlanData = useAppSelector(selectAllPlans);
+  const apExams = useTransferredCredits().ap;
 
   const obtainImportedRoadmap = async (schedName: string, currYear: string) => {
     // Get the result
     try {
-      const result = await trpc.zot4PlanImportRouter.getScheduleFormatted.query({
+      const [savedRoadmap, z4pApExams] = await trpc.zot4PlanImport.getScheduleFormatted.query({
```
Awesome-E, line 33: To reflect changes from the previous comment
```suggestion
      const { savedRoadmap, apExams: z4pApExams } = await trpc.zot4PlanImport.getScheduleFormatted.query({
```

site/src/pages/RoadmapPage/ImportZot4PlanPopup.tsx
```diff
@@ -9,28 +9,51 @@ import trpc from '../../trpc.ts';
...
         scheduleName: schedName,
         studentYear: currYear,
       });
+
+      // Combine existing AP exams with AP exams from Zot4Plan; ignore duplicates
+      const combinedExams = [
+        ...apExams,
+        ...z4pApExams.filter((z4pExam) => !apExams.some((existingExam) => existingExam.examName === z4pExam.examName)),
+      ];
```
Awesome-E, line 42: I'd prefer it if the final array construction was more straightforward.

```suggestion
      const newExams = z4pApExams.filter((imported) => !apExams.some((existing) => existing.examName === imported.examName))
      const combinedExams = apExams.concat(newExams);
```

Awesome-E (changes requested):
Looks mostly good, but a few code style things.

CadenLee2 (commented):
## Testing
- Tried out some schedules while logged out and all the exams I tried seem to import correctly
- Current exams (and their scores/units if there is a match) are not overwritten, which is good
- Problem: when I'm logged in and import a schedule (`test_ap_m_2`), the APs show up temporarily but for some reason the new roadmap doesn't get created, the APs don't seem to be saved to the db, and I get a notification that the process failed,:

image

Is anyone able to replicate this?

## Thoughts
- The import menu description is definitely pretty long already, but it might be worth mentioning that AP exams will also be imported to the user's account
- Zot4Plan does store the (user-defined) courses that each AP clears, and since we already handle AP clearing logic ourselves it's probably fine that we don't do anything with those. But once #689 is done, it might be possible to intelligently auto-choose rewards in certain cases. Mentioning just in case, though it's out of scope and probably too small a feature to be worth the time

fiveminus1:
> The import menu description is definitely pretty long already, but it might be worth mentioning that AP exams will also be imported to the user's account

Yeah, I think there isn't a great indication of where APs go. Perhaps an indicator on the "Transfer Credits" button might be nice in the future?

--
I'm still unable to reproduce your issue with `test_ap_m_2` mentioned here, but I did fix the issue with importing a schedule with duplicate exams like `test_ap_m_3`. Current exams are still not overwritten when importing, but the backend should filter out any duplicates from Z4P before returning the array back to the frontend. Re-requesting your review @CadenLee2

CadenLee2 (approved):
Tested some more and things seem to work now. Since I can't seem to replicate that original issue beyond the first time, I'll assume it was something to do with how I was testing it.

Everything else looks good to me

Awesome-E (approved):
:basecampy:

PR #695, opened 2025-05-19
Categorize imported transcript transfers

### LINKED ISSUE #685: Update Transcript Imports to automatically categorize non-UCI credits
When the user goes to import a transcript, it will currently throw everything into "Other Transfers", which is not ideal. We should make it so that the user's data automatically gets sorted into Courses & AP Exams whenever possible.

This will require logic used in #666 to automatically detect AP Exams, and should use a single batch course request for all "non-'AP '-prefixed items" to automatically detect courses transferred

Transcript parsing logic is currently on the frontend and does not need to be changed.

Instead, the transfers array you receive after parsing should simply be input to a new backend route that does AP Exam and Course detection. That backend route should return an array of UCI courses and an array of AP Exams cleared (should be the same JSON structure as used in Redesigned Transfers).

After receiving the response, it should be joined with what's currently on the frontend. Unique new items should be added to the frontend arrays in the `transferCreditsSlice` AND should be saved immediately (both to localStorage and to account – wait for localStorage PR to be merged). The backend route may need to be modified to do nothing on row insert conflict in case the user runs this in two different tabs.

CadenLee2:
Transfers are now properly categorized and saved when importing a transcript

## Description

- Updated the transfer conversion helper code to support optional scores as well
- Refactored the transfer conversion helper code somewhat (could use some advice on other steps to make it more readable)
- Refactored `userAPExam` to be a shared type across backend and frontend to reduce duplicate code (update  #693 with this as well, depending on which gets merged first)
- Transfers are now categorized (and saved to the db, if logged in) in the transfer import
  - Specifically, they are upserted (on the frontend, so that we are free to change `overrideAllTransfers` to be more accurate to its name) before being saved
- Note: we may want to do more testing with multiple tabs/handling collisions on the backend as mentioned in #685

## Screenshots

Since @fiveminus1 included a video I had to as well

ezgif-75fb02f721921e

## Test Plan

- [ ] Test with various transcripts (such as your own)
- [ ] Test while logged in and logged out, and saving to the account as well as locally (try reloading)
- [ ] Test if you currently have various exams, courses, etc. to ensure things don't get duplicated or erased
- [ ] Ensure scores are accurate, edge cases are handled (like multiple Calc ABs), etc.

## Issues

Closes #685

### DIFF
--- a/api/src/controllers/transferCredits.ts
+++ b/api/src/controllers/transferCredits.ts
@@ -1,5 +1,5 @@
 import { z } from 'zod';
-import { transferData, TransferredGE } from '@peterportal/types';
+import { extendedTransferData, TransferredGE } from '@peterportal/types';
 import { publicProcedure, router, userProcedure } from '../helpers/trpc';
 import { ANTEATER_API_REQUEST_HEADERS } from '../helpers/headers';
 import { db } from '../db';
@@ -155,7 +155,7 @@ const transferCreditsRouter = router({
 
     await db.delete(transferredMisc).where(and(...conditions));
   }),
-  convertUserLegacyTransfers: publicProcedure.input(z.array(transferData)).query(async ({ input }) => {
+  convertUserLegacyTransfers: publicProcedure.input(z.array(extendedTransferData)).query(async ({ input }) => {
     return await organizeLegacyTransfers(input);
   }),
   overrideAllTransfers: userProcedure
--- a/api/src/helpers/transferCredits.ts
+++ b/api/src/helpers/transferCredits.ts
@@ -3,18 +3,11 @@ Server helpers for Transfer Credits - based on the transfer data migration scrip
 The script itself is not modified so that it can be self-contained
 */
 
-import { CourseAAPIResponse } from '@peterportal/types';
+import { CourseAAPIResponse, ExtendedTransferData } from '@peterportal/types';
 import { ANTEATER_API_REQUEST_HEADERS } from './headers';
 
 const AP_CALC_AB_SUBSCORE = 'AP Calculus BC, Calculus AB subscore';
 const AP_CALC_AB = 'AP Calculus AB';
-
-/** An extended version of TransferData that optionally allows for a score */
-type ExtendedTransferDataRaw = {
-  name: string;
-  units?: number | null | undefined;
-  score?: number;
-};
 
 type ApExamBasicInfo = {
   fullName: string; // E.g. "AP Microeconomics"
@@ -321,7 +314,7 @@ const mergeDuplicateCourseResults = (toInsertCourse: TransferredCourseRow[]) => 
 };
 
 /** Organize the data in the database */
-export const organizeLegacyTransfers = async (rows: ExtendedTransferDataRaw[]) => {
+export const organizeLegacyTransfers = async (rows: ExtendedTransferData[]) => {
   const allAps = await getAPIApExams();
 
   // Determine the unique transfers
@@ -333,7 +326,7 @@ export const organizeLegacyTransfers = async (rows: ExtendedTransferDataRaw[]) =
       existing.count++;
       existing.totalUnits += row.units ?? 0;
     } else {
-      lookup[row.name] = { totalUnits: row.units ?? 0, count: 1, score: row.score };
+      lookup[row.name] = { totalUnits: row.units ?? 0, count: 1, score: row.score ?? undefined };
     }
   });
   const uniqueRows: GroupedUncategorizedTransfer[] = Object.entries(lookup).map(([courseName, data]) => ({
--- a/api/src/helpers/transferCredits.ts
+++ b/api/src/helpers/transferCredits.ts
@@ -3,11 +3,18 @@ Server helpers for Transfer Credits - based on the transfer data migration scrip
 The script itself is not modified so that it can be self-contained
 */
 
-import { CourseAAPIResponse, TransferData } from '@peterportal/types';
+import { CourseAAPIResponse } from '@peterportal/types';
 import { ANTEATER_API_REQUEST_HEADERS } from './headers';
 
 const AP_CALC_AB_SUBSCORE = 'AP Calculus BC, Calculus AB subscore';
 const AP_CALC_AB = 'AP Calculus AB';
+
+/** An extended version of TransferData that optionally allows for a score */
+type ExtendedTransferDataRaw = {
+  name: string;
+  units?: number | null | undefined;
+  score?: number;
+};
 
 type ApExamBasicInfo = {
   fullName: string; // E.g. "AP Microeconomics"
@@ -19,10 +26,11 @@ type TransferredMiscRow = {
   units: number | null;
 };
 
-type TransferredMiscSelectedRow = {
+type GroupedUncategorizedTransfer = {
   courseName: string;
   totalUnits: number;
   count: number;
+  score?: number;
 };
 
 type TransferredApExamRow = {
@@ -179,7 +187,7 @@ const tryMatchCourse = (transferName: string, validCourses: Record<string, strin
 /** Handle some special cases for AP Calc AB, which could have a subscore,
   returning whether it's okay to proceed to add "AP Calculus AB" */
 const handleApCalcAB = (
-  transfer: TransferredMiscSelectedRow,
+  transfer: GroupedUncategorizedTransfer,
   toInsertAp: TransferredApExamRow[],
   hasSubscore: boolean,
 ): boolean => {
@@ -195,7 +203,7 @@ const handleApCalcAB = (
     // Insert the BC subscore here, then proceed normally so that AB is also inserted
     toInsertAp.push({
       examName: AP_CALC_AB_SUBSCORE,
-      score: null,
+      score: transfer.score ?? null,
       units: 0, // The units will be counted for the actual AB later, and the subscore will remain 0 units
     });
     return true;
@@ -204,7 +212,7 @@ const handleApCalcAB = (
 
 /** Handle organizing an AP exam, returning whether it was successfully categorized */
 const organizeApExam = (
-  transfer: TransferredMiscSelectedRow,
+  transfer: GroupedUncategorizedTransfer,
   toInsertAp: TransferredApExamRow[],
   hasSubscore: boolean,
   allAps: ApExamBasicInfo[],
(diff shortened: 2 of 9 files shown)


api/src/helpers/transferCredits.ts
```diff
@@ -186,36 +189,30 @@ const handleApCalcAB = (
...
+const organizeApExam = (
+  transfer: GroupedUncategorizedTransfer,
   toInsertAp: TransferredApExamRow[],
   hasSubscore: boolean,
   allAps: ApExamBasicInfo[],
 ): boolean => {
-  const bestMatch = tryMatchAp(transferName, allAps);
+  const bestMatch = tryMatchAp(transfer.courseName, allAps);
   if (!bestMatch) {
     // Could not match; leave it here
```
Awesome-E, line 215: "Leave it here" is a reference to the misc table, which is not necessary in this case.

```suggestion
  if (!bestMatch) return false;
```

api/src/helpers/transferCredits.ts
```diff
@@ -228,26 +225,21 @@ export const organizeApExam = (
...
-  transfer: TransferredMiscSelectedRow,
-  transferName: string,
+  transfer: GroupedUncategorizedTransfer,
   toInsertCourse: TransferredCourseRow[],
   validCourses: Record<string, string>,
 ): Promise<boolean> => {
-  const bestMatch = tryMatchCourse(transferName, validCourses);
+  const bestMatch = tryMatchCourse(transfer.courseName, validCourses);
   if (!bestMatch) {
     // Could not match; leave it here
```
Awesome-E, line 242: same as previous

api/src/helpers/transferCredits.ts
```diff
@@ -17,10 +19,11 @@ type TransferredMiscRow = {
   units: number | null;
 };
 
-type TransferredMiscSelectedRow = {
+type GroupedUncategorizedTransfer = {
```
Awesome-E, line 22: Consider adding a docstring to this type because "grouped" is not inherently clear on its meaning

api/src/helpers/transferCredits.ts
```diff
@@ -324,33 +314,30 @@ const mergeDuplicateCourseResults = (toInsertCourse: TransferredCourseRow[]) =>
...
       existing.count++;
       existing.totalUnits += row.units ?? 0;
     } else {
-      lookup[row.name] = { totalUnits: row.units ?? 0, count: 1 };
+      lookup[row.name] = { totalUnits: row.units ?? 0, count: 1, score: row.score ?? undefined };
     }
   });
-
-  const uniqueRows: TransferredMiscSelectedRow[] = Object.entries(lookup).map(([courseName, data]) => ({
+  const uniqueRows: GroupedUncategorizedTransfer[] = Object.entries(lookup).map(([courseName, data]) => ({
```
Awesome-E, line 332: If we're refactoring this whole file, I'd argue that these should be pulled out to helper functions so the main function is easier to read

api/src/helpers/transferCredits.ts
```diff
@@ -324,33 +314,30 @@ const mergeDuplicateCourseResults = (toInsertCourse: TransferredCourseRow[]) =>
...
-  const transfers: TransferredMiscSelectedRow[] = uniqueRows
-    //.limit(40) // For testing, only look at a few entries
-    .filter((transfer) => transfer.courseName != null && transfer.count != 0) as TransferredMiscSelectedRow[];
+  // Obtain/clean up all relevant transfers (invalid ones that cannot be categorized are filtered out here)
+  const transfers: GroupedUncategorizedTransfer[] = uniqueRows
+    .filter((transfer) => transfer.courseName != null && transfer.count != 0)
+    .map((transfer) => ({ ...transfer, courseName: transfer.courseName.trim() })) as GroupedUncategorizedTransfer[];
 
   // Get all the users who explicitly have an AB subscore
   let hasSubscore = false;
```
Awesome-E, line 343: Even the for loop here can be a helper function

something like
`const hasABSubscore = checkABSubscore(transfers, allAps)`

api/src/helpers/transferCredits.ts
```diff
@@ -362,44 +349,27 @@ export const organizeLegacyTransfers = async (rows: TransferData[]) => {
...
-    const transferName = transfer.courseName.trim();
-    if (transferName.startsWith('AP ')) {
+    if (transfer.courseName.startsWith('AP ')) {
       continue;
     }
-    coursesToValidate.add(removeAllSpaces(transferName));
+    coursesToValidate.add(removeAllSpaces(transfer.courseName));
   }
-  // console.log('Validating courses from the API (batch)');
   const validCourses = await validateCoursesAPI(coursesToValidate);
```
Awesome-E, line 357: similar idea here
`const validCourses = getValidCourseNames(transfers)`

would call a helper that builds the courses to validate set and loops through all transfers

api/src/helpers/transferCredits.ts
```diff
@@ -362,44 +349,27 @@ export const organizeLegacyTransfers = async (rows: TransferData[]) => {
...
-    if (transferName.startsWith('AP ')) {
-      const reorganized = organizeApExam(
-        transfer as TransferredMiscSelectedRow,
-        transferName,
-        toInsertAp,
-        hasSubscore,
-        allAps,
-      );
+    if (transfer.courseName.startsWith('AP ')) {
+      const reorganized = organizeApExam(transfer as GroupedUncategorizedTransfer, toInsertAp, hasSubscore, allAps);
```
Awesome-E, line 365: this cast seems like code smell, see if it's possible to eliminate?

site/src/pages/RoadmapPage/ImportTranscriptPopup.tsx
```diff
@@ -145,19 +153,70 @@ async function processTranscript(file: Blob) {
...
+        await trpc.transferCredits.overrideAllTransfers.mutate({
+          courses: mergedCourses,
+          ap: mergedAps,
+          ge: [],
+          other: mergedOther,
+        });
+      }
+
+      // Finally, set the actual plan
+      // This does not save automatically (the user has to save it manually)
```
Awesome-E, line 219: For consistency with z4p, we should make it save automatically right?

site/src/store/slices/transferCreditsSlice.ts
```diff
@@ -1,17 +1,11 @@
 import { createSlice, PayloadAction } from '@reduxjs/toolkit';
-import { APExam, TransferredGE } from '@peterportal/types';
+import { APExam, TransferredGE, UserAPExam } from '@peterportal/types';
 
 export interface TransferredCourse {
   courseName: string;
   units: number;
 }
```
Awesome-E, line 7: I'd argue that these other types should also be refactored, and that the 3 should be named consistently

`TransferredCourse`, `TransferredAPExam`, `TransferredGE`, `TransferredUncategorizedItem` (idk last one is kinda weird)

and they should _all_ be in `@peterportal/types` for consistency. Even better, they should be consolidated with the types currently defined in the transfer credits controller by using `z.infer`

Awesome-E (changes requested):

api/src/helpers/transferCredits.ts
CadenLee2: Fixed various comments in this file

api/src/helpers/transferCredits.ts
CadenLee2: Oh this was from the past when the null check was in this area, these casts are all unnecessary now and can be removed. Thanks for pointing it out

api/src/helpers/transferCredits.ts
CadenLee2: All these sections are now refactored into helper functions

site/src/pages/RoadmapPage/ImportTranscriptPopup.tsx
fiveminus1: It could be argued that since this saves to your current roadmap rather than a new one like Z4P does (which I personally think is fine), it shouldn't auto-save, but if that's the case then there should at least be a toast saying this isn't saved.

fiveminus1 (changes requested):
Love the video! #qualitypr #cadenpeterportal

-- 

So, I can't test this with my own transcript because it's broken for some reason - will DM you it if you want to check it out, but it just leads to a blank page when I import it. This may or may not be directly related to this PR, since my transcript has never worked. 

I have tested my roommate's transcript, and the saving behavior locally/logged in seems to be working as expected (other than the auto-saving Ethan & I commented on). Will try to get some other sample transcripts and test some edge cases, then edit this review if I find anything

Edit: Testing with another transcript, and it appears AP Calculus AB isn't correctly categorized - it goes into "Other Transferred Credits."

CadenLee2:
> So, I can't test this with my own transcript because it's broken for some reason

This should now be fixed: any invalid courses on the transcript will not be placed on the roadmap, and instead will also be included as other/uncategorized transfers. Duplicates should also be eliminated if there are any.

> it appears AP Calculus AB isn't correctly categorized

This is intentional, due to the fact that with certain transcript imports it could be ambiguous whether this is AB or AB Subscore on BC when we only have one instance of AP Calculus AB

api/src/controllers/transferCredits.ts
```diff
@@ -120,7 +100,7 @@ const transferCreditsRouter = router({
       .where(eq(transferredGe.userId, ctx.session.userId!));
     return response as TransferredGE[];
   }),
-  setTransferredGE: userProcedure.input(z.object({ GE: zodGETransferSchema })).mutation(async ({ input, ctx }) => {
+  setTransferredGE: userProcedure.input(z.object({ GE: zodTransferredGE })).mutation(async ({ input, ctx }) => {
```
Awesome-E, line 103: if you're updating the types, it'd probably be good to update this and make it more consistent as well.

```suggestion
  setTransferredGE: userProcedure.input(zodTransferredGE)).mutation(async ({ input, ctx }) => {
```

site/src/pages/RoadmapPage/ImportTranscriptPopup.tsx
Awesome-E: In my recent PR to consolidate things in the top navigation, it now adds to a new roadmap instead of overriding the current one

Awesome-E (approved):

fiveminus1 (approved):
Everything LGTM, good work and thanks for fixing my bug!!

api/src/controllers/transferCredits.ts
CadenLee2: Done

Awesome-E (approved):

PR #697, opened 2025-05-20
[DUPLICATE] Update Zot4Plan AP Imports

### LINKED ISSUE #686: Update Zot4Plan Imports to automatically categorize non-UCI credits

fiveminus1:
This is a duplicate rebase PR. See #693 for discussion.

## Description

Currently, importing a roadmap from [Zot4Plan](https://zot4plan.com/) does not import AP Exams nor handle them in any way. This PR ensures that AP Exams from Z4P are:
- Matched to an exam name using existing matching logic (#666)
- Categorized into the `transferred_ap_exam` table
- Reflected in the frontend

## Demo

z4p_aps

## Test Plan

- [ ] Import a schedule from Zot4Plan that includes AP exams (you can try `test_ap_migration`, which should have APs Computer Science A, Biology, and Calculus BC).
- [ ] Verify the proper APs, scores, and units are imported under the Transfer Credits menu.
- [ ] Confirm the proper rows were added to the `transferred_ap_exam` table.
- [ ] Confirm existing added exams are not overwritten.

## Issues

Closes #686 

Note: the details in the issue are inaccurate - courses are not placed into "Other Transfers" and batch course lookup is not required.

### DIFF
--- a/api/src/controllers/index.ts
+++ b/api/src/controllers/index.ts
@@ -28,7 +28,7 @@ export const appRouter = router({
   schedule: scheduleRouter,
   transferCredits: transferCreditsRouter,
   users: usersRouter,
-  zot4PlanImportRouter: zot4PlanImportRouter,
+  zot4PlanImport: zot4PlanImportRouter,
 });
 
 // Export only the type of a router!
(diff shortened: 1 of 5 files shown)


Awesome-E (approved):
Not sure why this is a draft PR but we'll basically just merge this one and close the other one

PR #701, opened 2025-05-21
Consolidate buttons in the header

Awesome-E:
## Description

This PR consolidates the buttons based on what we discussed in the meeting – Zot4Plan and Transcript imports are now alongside creating a new roadmap, and "Add Year" + "Transfer Credits" + "Save" are now in the top.

It also changes it so the buttons simply wrap on mobile.

## Screenshots

Before|After
:-:|:-:
image|image
image|image

## Test Plan

- [x] make sure staging looks good

## Issues

N/A

### DIFF
--- a/site/src/pages/RoadmapPage/AddYearPopup.scss
+++ b/site/src/pages/RoadmapPage/AddYearPopup.scss
@@ -1,5 +1,4 @@
 .add-year-icon {
-  margin-right: 4px;
   transform: scale(1.5);
 }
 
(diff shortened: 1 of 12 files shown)


zcharz:
looks good i like

one possible concern is when the screen is extra extra small, the buttons flow outside of the box. do we care about this? realistically a non-issue in practical use cases
image

Awesome-E:
> one possible concern is when the screen is extra extra small, the buttons flow outside of the box. do we care about this? realistically a non-issue in practical use cases

No, that is below a usable screen width. Generally if it works around 350-400px, it is fine.

zcharz:
in that case looks good big bossman

PR #703, opened 2025-05-21
Created roadmap loading icon

### LINKED ISSUE #696: Loading icon for roadmap
When you click on the roadmap, it takes a few seconds to load, but it just looks like the data is gone because there's no indication of work. A loading screen/suspense would be nice.

charlieweinberger:
## Description

Shows a spinning icon in that short time period between opening the roadmap page and having your data loaded.

I'm not sure about where/which elements should be shown always, vs. with a spinner inside it, vs. only after loading. That's half of this PR - the other half is the logic for determining when the data is loading, which takes into account both searching for a user and then setting the user's data in the roadmap. I'm also not sure if I did this logic the best, so I'm open to feedback.

Note: ~~this PR does slightly overlap with #701 , since this PR also modifies the transcript/z4p import buttons. However, it's a simple thing to merge (just deleting those buttons and their imports in `Planner.tsx`), so it shouldn't be a problem.~~ This is no longer a problem, #701 has since been merged into main and then pulled into this branch.

## Screenshots

Before:
Screenshot 2025-05-21 at 2 04 15 AM

After:
Screenshot 2025-05-21 at 3 18 36 AM

## Test Plan

- [ ] Verify that the spinner shows for the correct amount of time, no more and no less
- [ ] Verify that the spinner works for both accounts and local data
- [ ] Verify that the spinner looks good
- [ ] Verify that this PR doesn't break anything else

## Issues

Closes #696

### DIFF
--- a/site/src/pages/RoadmapPage/Planner.scss
+++ b/site/src/pages/RoadmapPage/Planner.scss
@@ -20,4 +20,9 @@
   .header {
     top: 0px;
   }
+
+  .loading-spinner {
+    display: flex;
+    justify-content: center;
+  }
 }
--- a/site/src/pages/RoadmapPage/Planner.tsx
+++ b/site/src/pages/RoadmapPage/Planner.tsx
@@ -51,18 +51,18 @@ const Planner: FC = () => {
   return (
     <div className="planner">
       <PlannerLoader />
+      <Header
+        courseCount={courseCount}
+        unitCount={unitCount}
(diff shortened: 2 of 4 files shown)


Awesome-E (approved):
There are some other areas where we do the same thing to center the spinner, but I think that might be addressed in migrating to MUI anyways. So thus this PR looks fine to me

PR #706, opened 2025-05-24
Fix click actions not working in Import popups

Awesome-E:
## Description

Unwraps buttons from `Dropdown.Item` so that the click handler installed by `Dropdown.Item` does not interfere with our ability to perform default actions such as clicking links or opening file uploads.

Previously, the content was inside `<Dropdown.Item/>` and `<Dropdown.Item/>` installed a click handler. This click handler would call `event.preventDefault()` at some point, which disables the user's ability to click links and open file upload prompts.

## Test Plan

- [ ] Check that buttons and links inside import transcript and zot4plan schedule work again

### DIFF
--- a/site/src/pages/RoadmapPage/Header.scss
+++ b/site/src/pages/RoadmapPage/Header.scss
@@ -45,6 +45,7 @@
     padding: 8px;
     align-items: center;
     gap: 8px 12px;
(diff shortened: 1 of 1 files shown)


CadenLee2 (approved):
z-index is good now and the file upload button works

PR #709, opened 2025-05-28
Simplified prereq tree by merging 'all of' and 'one of'

charlieweinberger:
## Description

For certain courses (ex: [ICS 33](https://peterportal.org/course/I&CSCI33), [COMPSCI 113](https://peterportal.org/course/COMPSCI113), [PHYSICS 52B](https://peterportal.org/course/PHYSICS52B)), the prerequisite tree shows "all of" and then immediately "one of", and then lists a set of courses. In these trees, the "all of" label is redundant, and should be hidden in favor of just showing the "one of".

## Screenshots

Before:
Screenshot 2025-05-27 at 8 16 17 PM

After:
Screenshot 2025-05-27 at 8 16 56 PM

## Test Plan

- [ ] Verify that all courses that have "all of" and then "one of" now just have "one of"
- [ ] Verify that all other courses are unchanged
- [ ] Verify that the new prereq tree looks good

## Issues

Does not close any issues.

### DIFF
--- a/site/src/component/PrereqTree/PrereqTree.tsx
+++ b/site/src/component/PrereqTree/PrereqTree.tsx
@@ -74,9 +74,8 @@ const PrereqTreeNode: FC<TreeProps> = (props) => {
   // if value is an object, render the rest of the sub tree
   else {
     const prereqTree = prerequisite as PrerequisiteTree;
-    const prereqTreeMap = prereqTree as Record<string, PrerequisiteTree[]>;
     const prereqTreeType = Object.keys(prereqTree)[0] as 'AND' | 'OR' | 'NOT';
-    const prereqChildren = prereqTreeMap[prereqTreeType];
+    const prereqChildren = prereqTree[prereqTreeType] as PrerequisiteTree[];
 
     if (prereqTree.AND && prereqChildren.length === 1 && prereqChildren[0].OR) {
       return <PrereqTreeNode prerequisiteNames={props.prerequisiteNames} prerequisiteJSON={prereqChildren[0]} />;
--- a/site/src/component/PrereqTree/PrereqTree.tsx
+++ b/site/src/component/PrereqTree/PrereqTree.tsx
@@ -74,7 +74,7 @@ const PrereqTreeNode: FC<TreeProps> = (props) => {
   // if value is an object, render the rest of the sub tree
   else {
     const prereqTree = prerequisite as PrerequisiteTree;
-    const prereqTreeType = Object.keys(prereqTree)[0] as 'AND' | 'OR' | 'NOT';
+    const prereqTreeType = Object.keys(prereqTree)[0] as keyof PrerequisiteTree;
     const prereqChildren = prereqTree[prereqTreeType] as PrerequisiteTree[];
 
     if (prereqTree.AND && prereqChildren.length === 1 && prereqChildren[0].OR) {
--- a/site/src/component/PrereqTree/PrereqTree.tsx
+++ b/site/src/component/PrereqTree/PrereqTree.tsx
@@ -75,22 +75,24 @@ const PrereqTreeNode: FC<TreeProps> = (props) => {
   }
   // if value is an object, render the rest of the sub tree
   else {
-    const prereqTree = prerequisite as Record<string, PrerequisiteNode[]>;
+    const prereqTree = prerequisite as PrerequisiteTree;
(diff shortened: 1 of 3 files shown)


site/src/component/PrereqTree/PrereqTree.tsx
```diff
@@ -75,22 +73,24 @@ const PrereqTreeNode: FC<TreeProps> = (props) => {
   }
   // if value is an object, render the rest of the sub tree
   else {
-    const prereqTree = prerequisite as Record<string, PrerequisiteNode[]>;
+    const prereqTree = prerequisite as PrerequisiteTree;
+    const prereqTreeMap = prereqTree as Record<string, PrerequisiteTree[]>;
+    const prereqTreeType = Object.keys(prereqTree)[0] as 'AND' | 'OR' | 'NOT';
+    const prereqChildren = prereqTreeMap[prereqTreeType];
+
```
CadenLee2, line 80: The `as Record<string, PrerequisiteTree[]>` looks a bit confusing to me, why is it needed? If it's needed then maybe put the line declaring `prereqTreeMap` directly above `prereqChildren` since that's the only place where it's being used

CadenLee2 (commented):
Tested out several courses and this seems to work, the code is more readable overall and the simplified tree is also easier to understand.

site/src/component/PrereqTree/PrereqTree.tsx
charlieweinberger: Looks like I can get rid of `prereqTreeMap` by just explicitly defining `prereqChildren` as `PrerequisiteTree[]`. Just pushed that change.

site/src/component/PrereqTree/PrereqTree.tsx
```diff
@@ -75,22 +73,23 @@ const PrereqTreeNode: FC<TreeProps> = (props) => {
   }
   // if value is an object, render the rest of the sub tree
   else {
-    const prereqTree = prerequisite as Record<string, PrerequisiteNode[]>;
+    const prereqTree = prerequisite as PrerequisiteTree;
+    const prereqTreeType = Object.keys(prereqTree)[0] as 'AND' | 'OR' | 'NOT';
```
Awesome-E, line 77: try to avoid using a cast – if possible, `const prereqTreeType: keyof PrerequisiteTree` or something like that would be better

types/src/course.ts
```diff
@@ -8,6 +8,8 @@ export type Prerequisite = components['schemas']['prereq'];
 
 export type PrerequisiteTree = components['schemas']['prereqTree'];
 
+export type PrerequisiteNode = Prerequisite | PrerequisiteTree;
```
Awesome-E, line 11: good refactor

site/src/component/PrereqTree/PrereqTree.tsx
charlieweinberger: Good point, I just replaced `'AND' | 'OR' | 'NOT'` with `keyof PrerequisiteTree`.

CadenLee2: The `keyof` change is good, and it looks like the `as` cast is still needed because `Object.keys` will just return strings for some reason, which can't be made `keyof PrerequisiteTree` implicitly

site/src/component/PrereqTree/PrereqTree.tsx
```diff
@@ -68,29 +66,30 @@ const PrereqTreeNode: FC<TreeProps> = (props) => {
...
         />
       </li>
     );
   }
   // if value is an object, render the rest of the sub tree
   else {
-    const prereqTree = prerequisite as Record<string, PrerequisiteNode[]>;
+    const prereqTree = prerequisite as PrerequisiteTree;
+    const prereqTreeType = Object.keys(prereqTree)[0] as keyof PrerequisiteTree;
+    const prereqChildren = prereqTree[prereqTreeType] as PrerequisiteTree[];
```
Awesome-E, line 78: is this possible to achieve with type annotations (`const name: Type = ...`) instead of casts? If not, you can resolve this comment.

Awesome-E (approved):
Pretty much good! Just a quick final question to take a look at

site/src/component/PrereqTree/PrereqTree.tsx
charlieweinberger: Unfortunately, not possible.

PR #710, opened 2025-06-01
made it so saving a course inserts it into the course bag alphabetically

charlieweinberger:
## Description

Currently, when a course is saved, it's added to the bottom of the list of saved courses. However, after refreshing the page, the courses then get sorted alphabetically, meaning they're in a different order than when just adding courses to the bottom. If we're going to sort courses alphabetically after refreshing, and just because sorting alphabetically is more organized, we should also sort courses when initially adding them to the saved course list, instead of adding them to the bottom of the list. This PR does just that.

## Screenshots

Before:
https://github.com/user-attachments/assets/11fbcbc3-f44d-469d-a951-bc59cfc81c9b

After:
https://github.com/user-attachments/assets/60ac7bad-7441-4b2b-bde6-a14b26262767

## Test Plan

- [ ] Verify that courses are inserted into the list correctly
- [ ] Verify that nothing breaks from this change/there are no random edge cases

## Issues

Doesn't close an existing issue

### DIFF
--- a/site/src/store/slices/coursebagSlice.ts
+++ b/site/src/store/slices/coursebagSlice.ts
@@ -12,7 +12,10 @@ export const coursebagSlice = createSlice({
       state.coursebag = action.payload;
     },
     addCourseToBagState: (state, action: PayloadAction<CourseGQLData>) => {
-      state.coursebag?.push(action.payload);
+      if (!state.coursebag) return;
+      const courseIndex = state.coursebag.findIndex((c) => c.id > action.payload.id);
+      const spliceIndex = courseIndex === -1 ? state.coursebag.length : courseIndex;
+      state.coursebag.splice(spliceIndex, 0, action.payload);
     },
     removeCourseFromBagState: (state, action: PayloadAction<CourseGQLData>) => {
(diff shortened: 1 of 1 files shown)


Awesome-E (commented):
This seems to do what it says. The database likely returns it in a set order because the course ID is part of the primary key, and this PR, to my knowledge, does address the inconsistency.

However, there is a weird edge case that I ran into, which is that courses don't _quite exactly_ go in the order I'd expect all the time. For example, I'd naturally think that ICS H32 would go before 51 since it's a variant of 32, but because it starts with an H, it goes after all ICS courses.

But, I think you might still be on the right track. Since sorting on the list when adding each one, we can just add them in sorted order, and that way no matter what, courses are added correctly. (wrote the last bit in a rush)

site/src/store/slices/coursebagSlice.ts
```diff
@@ -12,7 +12,14 @@ export const coursebagSlice = createSlice({
       state.coursebag = action.payload;
```
Awesome-E, line 12: note that because this sets the initial state, you will need to sort the payload here. Consider adding a helper function that compares course order, then use that both here and in the `addCourseToBagState` function

PR #711, opened 2025-06-01
Show unread dot for imported transfers

### LINKED ISSUE #699: Unread dot for imported transcripts
Currently, when the user imports a transcript or Zot4Plan AP Exam or course, there's no indication that it was newly imported from within the import transcripts menu.

It would be nice to see an indication that an AP exam or course was added indirectly (i.e. imported) via an unread dot. This would also encourage users to check out the new transfer credits menu.

Once the user closes the menu, the unread indicator is cleared.

CadenLee2:
To draw attention to the transfer credits menu and clarify that transfers are imported, when any new transfers are added from an imported transcript/Zot4Plan schedule, a red "unread dot" will appear on the menu and on the new transfers inside it. The dots persist until the menu is viewed and closed or the page is reloaded.

## Description

- Added an `unreadTransfers` variable in the transfer credits slice, which stores arrays of any newly added AP exams, courses, or other transfers (no GEs because they can't be imported)
- Importing a Zot4Plan schedule or transcript will add the names of all *new* transfers to that state variable
- If there are unread transfers, the unread dots will appear
  - Opinions on the red dot location/size on the Transfer Credits button?
  - All new transfer items in the menu will display a "NEW" bubble in the corner
- Closing the menu clears all `unreadTransfers`
  - Also, since these aren't saved to `localStorage`, they don't persist on reload
- Note: editing an item does not remove the "NEW" status because it would be a significant amount of code addition for a very small feature; if we think this is important in the future, it can be changed.

## Screenshots

| Dot on the Transfer Credits button when there are new transfers | Dot on the new transfers |
| --- | --- |
| image | image |

## Test Plan

- [ ] Try adding some AP exams or courses then importing a transcript (with some AP exams overlapping). The menu should show a red dot and the new ones should be labeled as "NEW" when you open it, then the unread dots should disappear once you close it for the first time.
- [ ] Try importing a Zot4Plan schedule and the same behavior should happen
- [ ] Try removing some (but not all) uncategorized transfers then importing the transcript/schedule again and the ones that you left shouldn't display the unread dots
- [ ] Unread dots shouldn't appear in any other cases and the import process/transfer menu should not be modified in any other way

## Issues

Closes #699


site/src/component/UnreadDot/UnreadDot.scss
```diff
@@ -0,0 +1,26 @@
+.unread-pill,
+.unread-circle {
+  position: absolute;
+  background-color: var(--red);
```
charlieweinberger, line 4: I won't comment on the entire PR since my review wasn't requested, but I do want to point out one small thing - unless there's a specific reason to use CSS's default shade of red, which this is currently doing, shouldn't this unread dot be using our `--red-primary` color?

site/src/pages/RoadmapPage/Header.tsx
```diff
@@ -18,10 +19,23 @@ interface HeaderProps {
...
+    if (showTransfers) {
+      // After closing the menu, clear all the unread markers
+      dispatch(clearUnreadTransfers());
+    }
+    dispatch(setShowTransfersMenu(!showTransfers));
+  };
+
+  const unreadTransfers = useAppSelector((state) => state.transferCredits.unreadTransfers);
+
+  const showUnreadTransfers =
```
Awesome-E, line 35: Update variable naming so that it is clear that we mean "they exist" rather than "it should be displayed"
```suggestion
  const hasUnreadTransfers =
```

site/src/pages/RoadmapPage/transfers/APExamsSection.tsx
```diff
@@ -138,7 +146,7 @@ const APExamsSection: FC = () => {
         Enter the names of AP Exams that you&rsquo;ve taken to clear course prerequisites.
       </SectionDescription>
       {userAPExams.map((exam) => (
-        <APCreditMenuTile key={exam.examName} userExamInfo={exam} />
+        <APCreditMenuTile key={exam.examName} userExamInfo={exam} unread={unreadExams.includes(exam.examName)} />
```
Awesome-E, line 149: hmm I'm not too sure this is the right approach. I was thinking that maybe you could add an `unread?: true` (or `unread?: boolean`) property to each of the items in the course/ap/other transfers rather than maintaining separate lists to do lookups.

Your approach to adding them would also be simpler, as when you click import, you simply need to categorize and `.forEach(t => { t.unread = true; })` (or you could take in `markUnread` as a parameter in the categorize function)

Removing unread indicates would then constitute simply looping through course, ap, and other transfers, and doing `delete t.unread` for each item in the 3 lists.

site/src/store/slices/transferCreditsSlice.ts
```diff
@@ -19,6 +19,11 @@ export const transferCreditsSlice = createSlice({
     userAPExams: [] as TransferredAPExam[],
     transferredGEs: [] as TransferredGE[],
     uncategorizedCourses: [] as TransferredUncategorized[],
+    unreadTransfers: {
+      apNames: [] as string[],
+      courseNames: [] as string[],
+      otherNames: [] as string[],
+    },
```
Awesome-E, line 26: hmm I'm not too sure this is the right approach. I was thinking that maybe you could add an `unread?: true` (or `unread?: boolean`) property to each of the items in the course/ap/other transfers rather than maintaining separate lists to do lookups.

Your approach to adding them would also be simpler, as when you click import, you simply need to categorize and `.forEach(t => { t.unread = true; })` (or you could take in `markUnread` as a parameter in the categorize function)

Removing unread indicates would then constitute simply looping through course, ap, and other transfers, and doing `delete t.unread` for each item in the 3 lists.

Awesome-E: Since we're using unified types as well, you could consider creating a templated type that applies all "client-only" properties, i.e.

```ts
type TransferWithUnread<T> = T & { unread?: true; }
```
then use that in the slice. Alternatively, you could also add `unread?` directly to the type and use a docstring to indicate when it's used, though (1) I'm not sure how you'd do it with zod and (2) that's not nearly as type-safe.

site/src/pages/RoadmapPage/transfers/UncategorizedCreditsSection.tsx
```diff
@@ -32,7 +46,11 @@ const UncategorizedCreditsSection: FC = () => {
       </SectionDescription>
 
       {courses.map((course) => (
-        <UncategorizedMenuTile key={`${course.name}-${course.units}`} name={course.name} units={course.units} />
+        <UncategorizedMenuTile
+          key={`${course.name}-${course.units}`}
+          uncategorizedInfo={course}
+          unread={course.name ? unreadOther.includes(course.name) : false}
```
Awesome-E, line 52: Looks like this code is more clunky than it has to be because of typescript `undefined` complaints?

you could also do this:
```suggestion
          unread={unreadOther.includes(course.name ?? '')}
```

site/src/pages/RoadmapPage/transfers/UncategorizedCreditsSection.tsx
```diff
@@ -32,7 +46,11 @@ const UncategorizedCreditsSection: FC = () => {
       </SectionDescription>
 
       {courses.map((course) => (
-        <UncategorizedMenuTile key={`${course.name}-${course.units}`} name={course.name} units={course.units} />
+        <UncategorizedMenuTile
+          key={`${course.name}-${course.units}`}
+          uncategorizedInfo={course}
```
Awesome-E, line 51: Is there a reason this can't be kept as two props? If it's for consistency with course transfers, I'd recommend changing the other one instead so the component's code can be nicer

site/src/pages/RoadmapPage/transfers/UncategorizedCreditsSection.tsx
```diff
@@ -6,19 +6,33 @@ import { removeUncategorizedCourse } from '../../../store/slices/transferCredits
 import { useAppDispatch, useAppSelector } from '../../../store/hooks';
 import { TransferredUncategorized } from '@peterportal/types';
 
-const UncategorizedMenuTile: FC<TransferredUncategorized> = ({ name, units }) => {
+const UncategorizedMenuTile: FC<{ uncategorizedInfo: TransferredUncategorized; unread: boolean }> = ({
+  uncategorizedInfo,
+  unread,
+}) => {
```
Awesome-E, line 12: Following the previous comment, I'd like to still be able to read name and units:

```suggestion
const UncategorizedMenuTile: FC<...> = ({ name, units, unread }) => {
```
The `FC`'s type could be a `type UncategorizedMenuTileProps = TransferredUncategorized & { unread; }`, but if you also take the suggestion of creating a frontend type that adds `unread?: boolean` on top of the transfers, then you can directly use that type, i.e. `FC<TransferWithUnread<TransferredUncategorized>>`.

(you could also name it something like `DraftTransfer` idk, this depends on how much abstraction you like)

Awesome-E (changes requested):

site/src/component/UnreadDot/UnreadDot.scss
CadenLee2: Good point, I had seen `--red` in inspect element and forgot we have our own red. I think `--red` comes from bootstrap (?). Anyway, changing this

site/src/store/slices/transferCreditsSlice.ts
CadenLee2: I ended up going with this client-only type solution because it seems to make the most sense. I'm still undecided on whether to make *all* the reducers in the transfer credits slice use `TransferWithUnread`, or *only* the ones that might directly deal with unread transfers.

Awesome-E: Assuming the slice state has the client-side type, it's probably fine to leave the existing reducers take in the shared type and just have the ones that need to deal with it take in the client type.

site/src/store/slices/transferCreditsSlice.ts
```diff
@@ -27,25 +29,25 @@ export const transferCreditsSlice = createSlice({
...
     },
-    addTransferredCourse: (state, action: PayloadAction<TransferredCourse>) => {
+    addTransferredCourse: (state, action: PayloadAction<TransferWithUnread<TransferredCourse>>) => {
       state.transferredCourses.push(action.payload);
     },
     removeTransferredCourse: (state, action: PayloadAction<string>) => {
       state.transferredCourses = state.transferredCourses.filter((course) => course.courseName !== action.payload);
     },
-    updateTransferredCourse: (state, action: PayloadAction<TransferredCourse>) => {
+    updateTransferredCourse: (state, action: PayloadAction<TransferWithUnread<TransferredCourse>>) => {
```
Awesome-E, line 38: You don't seem to call these reducer functions with the unread type here, consider removing the template type for now?

site/src/pages/RoadmapPage/ImportTranscriptPopup.tsx
```diff
@@ -205,21 +206,27 @@ const ImportTranscriptPopup: FC = () => {
...
+      const newAps = apUnread.filter(
         (imported) => !currentAps.some((existing) => existing.examName == imported.examName),
       );
       const mergedAps = currentAps.concat(newAps);
 
-      const newCourses = courses.filter(
+      const newCourses = coursesUnread.filter(
         (imported) => !currentCourses.some((existing) => existing.courseName == imported.courseName),
       );
       const mergedCourses = currentCourses.concat(newCourses);
```
Awesome-E, line 227: it looks like the merged variables may or may not have unread, and that the trpc route is called with items that might have an unread attribute. I think that's fine since we just ignore the unknown properties, but I would like a confirmation that I'm understanding the code correctly.

site/src/store/slices/transferCreditsSlice.ts
CadenLee2: Done

site/src/pages/RoadmapPage/ImportTranscriptPopup.tsx
CadenLee2: I believe this is right. Regardless, the property won't ever be used by the route, so it should be okay

Awesome-E (approved):

fiveminus1 (approved):

PR #712, opened 2025-06-02
Consolidate MUI and PeterPortal Themes under a single `<AppThemeProvider/>`

Awesome-E:
## Description

This PR adds a `ThemeProvider` for Material UI so that button icons appear in the correct color instead of always having black text. Along with this, I extracted `ThemeContext.Provider` into a new `AppThemeProvider` file so its logic can be decoupled from `App.tsx`.

This app theme provider handles the logic for the user switching PeterPortal themes as well as creating the appropriate MUI theme so that buttons (and presumably other components in the future) have the correct text color.

## Screenshots

bookmark icon is now the correct color in dark mode
image

## Test Plan

- [ ] Check that switching peterportal themes still works
- [ ] Check that Material UI icons look correct in both light and dark mode

## Issues

N/A


charlieweinberger (approved):
LGTM. In both light/dark mode and when signed in/signed out, switching themes still works properly and MUI icons look correct.

PR #713, opened 2025-06-02
Simplified app.tsx

charlieweinberger:
## Description

- Moved the `loadCoursebag` callback function into the `useCoursebag()` hook
- Removed unnecessary HTML related to the `<SideBar/>` and `<ChangelogModal/>` components in `App.tsx`

Note: this PR overlaps with #712 , so this PR should wait to be merged w/ main until that PR is.

## Screenshots

N/A

## Test Plan

- [ ] Verify that loading coursebag data acts exactly the same as before
- [ ] Verify that the sidebar looks & acts exactly the same as before
- [ ] Verify that the changlog modal looks & acts exactly the same as before

## Issues

N/A


site/src/hooks/coursebag.ts
```diff
@@ -44,5 +45,13 @@ export function useCoursebag() {
...
-  return { coursebag: coursebag ?? [], addCourseToBag, removeCourseFromBag, toggleBookmark };
+  const loadCoursebag = useCallback(async () => {
+    const courseIds = isLoggedIn
+      ? await trpc.savedCourses.get.query()
+      : JSON.parse(localStorage.getItem('coursebag') ?? '[]');
+    const coursebag = await searchAPIResults('courses', courseIds);
+    dispatch(setCoursebag(Object.values(coursebag)));
+  }, [dispatch, isLoggedIn]);
+
+  return { coursebag: coursebag ?? [], addCourseToBag, removeCourseFromBag, toggleBookmark, loadCoursebag };
```
Awesome-E, line 56: instead of returning it as a function, you should just use the `useCourseBag` hook itself to load the user data, similar to how `useLoadTransferredCredits()` works.

site/src/component/SideBar/SideBar.tsx
```diff
@@ -115,7 +115,7 @@ const SideBar = () => {
   );
 
   return (
-    <>
+    <div>
```
Awesome-E, line 118: Do not convert fragments to `div`s for no reason. Unless you have a reason, it's not explicitly better to do this, and it can mess with styles

site/src/App.tsx
```diff
@@ -21,10 +21,10 @@ import ThemeContext from './style/theme-context';
 import trpc from './trpc';
 import { Theme } from '@peterportal/types';
 import { useAppDispatch } from './store/hooks';
-import { searchAPIResults } from './helpers/util';
-import { setCoursebag } from './store/slices/coursebagSlice';
 import { useIsLoggedIn } from './hooks/isLoggedIn';
 
+import { useCoursebag } from './hooks/coursebag';
```
Awesome-E, line 26: if we're refactoring course bag in general, we should be updating verbiage to say "saved courses" or "bookmarks". I think "saved courses" might be the better term here since (1) that's how it's stored in the db and (2) we don't actually use the terminology "bookmark"... only the icon.

site/src/hooks/coursebag.ts
Awesome-E: although, maybe it's still better that the loader is called once. I do like having in a separate file from app.tsx, but i think maybe it's better to have as a different function than useCoursebag...

Awesome-E (changes requested):

site/src/App.tsx
charlieweinberger: For this entire PR, I wasn't sure when to use "coursebag" vs. "saved courses" vs. "bookmark", since we do use all 3 terms to varying extents throughout the codebase. As far as I could tell, coursebag was the most commonly used term, so I stuck with that. I do think that using one term for as much as possible, instead of 3, would be ideal, and I don't particularly care which of the 3 terms we choose to use primarily. However, I think that that's a larger scale change, not something to think about for this PR.

CadenLee2 (commented):
Things seem to work and look the same on staging, but I left a comment on the side bar div thing mentioned above

site/src/component/SideBar/SideBar.tsx
charlieweinberger: I think it makes more sense to have `<div>` in `SideBar.tsx`, because
- the code in `SideBar.tsx` needs to be wrapped by something (currently `<>`), and IMO it makes more sense to replace the `<>` with `<div>` than it does to keep `<>` but also have `<div>` in another file
- `<div>` doesn't have a class (since `app-sidebar` doesn't exist)
- `<ChangelogModal />` isn't wrapped by a `<div>`, so wrapping `SideBar` with a `<div>` seems inconsistent

Awesome-E: Oh I see the confusion. The reason for `<></>` is because these are actually two _distinct_ sidebar elements (one `.sidebar.mini` for persistent, and one `.sidebar` for opened by hamburger button). Because these are two distinct fixed position elements who are both locked to the viewport, it makes more sense to use the "natural" parent of the app body rather than create a new div to be the shared parent of the two child elements.


> `<ChangelogModal />` isn't wrapped by a `<div>`, so wrapping `SideBar` with a `<div>` seems inconsistent.

that doesn't make sense to me?

charlieweinberger:
Just pushed an update to the `useCoursebag()` hook (now named `useSavedCourses()`) in #714

site/src/component/SideBar/SideBar.tsx
charlieweinberger: > that doesn't make sense to me?

I was referring to the symmetry of having `<SideBar />` and `<ChangelogModal />` on the same indentation level and equidistant from the routes. It's a minor thing, but it does look better imo, in conjunction with what I said above about `<>` vs. `<div>`.

Awesome-E: I see. It is the same as it was before, but I think the wrapper of `app-sidebar` in its entirety should not exist since both child elements are fixed to the viewport.

I do agree that it's better than `<div>` then `<></>`

Honestly `<ChangelogModal/>` should also not be wrapped in a div when called.

charlieweinberger:
Merged main (aka the code from #714) into this branch, so this PR should now be good to merge into main

charlieweinberger:
@Awesome-E Oh my bad, I misunderstood that convo. Just removed the `<div>`. If you want me to change anything else, lmk 👍

Awesome-E (approved):

PR #714, opened 2025-06-02
Create the `useSavedCourses()` and `useLoadSavedCourses()` hooks

charlieweinberger:
## Description

I noticed that we had several instances of searching for if a course is included in the `coursebag`, so I created the helper function named `coursebagIncludes()` within `useCoursebag()` to do this functionality, and then simplified existing code that used this new function.

Edit: in addition to this ^^^, I renamed every instance of cousebag/bookmark to saved courses.

## Screenshots

N/A

## Test Plan

- [ ] Verify that the modified code doesn't change how the website performs in any way

## Issues

N/A


site/src/component/CourseInfo/CourseInfo.tsx
```diff
@@ -14,9 +14,8 @@ interface CourseProp {
 }
 
 export const CourseBookmarkButton: FC<CourseProp> = ({ course }) => {
-  const { coursebag: bookmarks, toggleBookmark } = useCoursebag();
-  const isBookmarked = bookmarks.some((c) => c.id === course.id);
-
+  const { coursebagIncludes, toggleBookmark } = useCoursebag();
+  const isBookmarked = coursebagIncludes(course);
   return (
```
CadenLee2, line 19: As Ethan mentioned on your other PR, the different terms ("saved courses" vs "course bag" vs "bookmarked") make the code a bit more confusing, and I'm not sure about "saved courses" or "course bag", but I agree that "bookmarked" should only be in this bookmark button itself (which could also be renamed to Save Course Button in the future). For now I think `coursebag` where you have it now is okay since it's used a lot in related code

CadenLee2 (approved):
The course bag still works on staging (adding, removing, including while logged in). Changes seem straightforward and for now using `coursebag` in names seems okay

Awesome-E (approved):
Made a tiny change but otherwise LGTM!

PR #716, opened 2025-06-03
Convert buttons with only an icon inside them to MUI Icon Buttons

Awesome-E:
## Screenshots

Before|After
:-:|:-:
image|image
image|image

## Test Plan

- [ ] Ensure these buttons now exhibit MUI behavior (circular background on hover, click and hold does the MUI circle outwards animation)
- [ ] Check that there are no other buttons that have only an icon in them.

## Issues

N/A


charlieweinberger (changes requested):
Looks mostly good, just 2 small comments, one here and one in a separate comment.

When the window is narrow, starting at 900px, the trashcan icon in the roadmap multiplan dropdown becomes smaller than the other two icons, and all icons start to become un-centered from their buttons.

Screenshot 2025-06-06 at 7 58 31 AM

site/src/component/Review/Review.scss
```diff
@@ -63,6 +63,15 @@
...
     gap: 12px;
+    height: 24px;
+  }
+
+  .edit-buttons {
+    display: flex;
+    flex-direction: row;
+    float: right;
+    gap: 4px;
+    margin-right: -8px;
```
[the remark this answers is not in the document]
Awesome-E: yeah float is probably a leftover property. I don't think they are too close to the top as it should be the same as the x and y padding on the heading on the left

CadenLee2 (approved):
Buttons all look and behave as expected.

I don't really like how we have to change the font size of the trash icon, especially considering that its sharp edges look slightly different (at least in my browser), but it doesn't seem like there's much way around it and it's less noticeable than having the trash can be a completely different size.

image

site/src/component/Review/Review.scss
```diff
@@ -63,6 +63,14 @@
     justify-content: space-between;
     align-items: center;
     gap: 12px;
+    height: 24px;
+  }
+
+  .edit-buttons {
+    display: flex;
+    flex-direction: row;
+    gap: 0px;
```
charlieweinberger, line 72: Why keep `gap: 0px`? I do think we should keep the gap, but if we're gonna not have a gap, I'd suggest just deleting this line too.

Here's what it looks like to me, where imo its a bit strange that the buttons are right next to each other. I think a gap of `4px` is good.

https://github.com/user-attachments/assets/c370273d-d15e-43f5-adbd-ba5d5e1bccdc

charlieweinberger (commented):
Looks good, the visual bug I mentioned is fixed, one small comment tho

site/src/component/Review/Review.scss
Awesome-E: I believe I changed it since I'd rather have the buttons look correctly spaced apart when they're not hovered

charlieweinberger (approved):
> I believe I changed it since I'd rather have the buttons look correctly spaced apart when they're not hovered

Makes sense, ig it's just a personal preference of mine. All good to merge 👍

charlieweinberger (changes requested):
Wait, just realized - the `navbar-menu` CSS class is now unused, but still exists the code. If it's not needed anymore, then the code for it should be deleted.

PR #717, opened 2025-06-04
Simplify Chart (GradeDist/QuarterTooltip) Colors/Styling

charlieweinberger:
## Description

- Added chart colors to `index.css`, and simplified them
  - Created secondary-light colors, that are the same as existing secondary colors but don’t have the corresponding dark mode variants
- Used `getCSSVariable()` to get colors from CSS variables in `GradeDist/Chart.tsx`, `GradeDist/Pie.tsx`, and `QuarterTooltip/Chart.tsx`
- Abstracted shared components/helper functions into `helpers/styling.ts` and `components/ChartTooltip.tsx`
- Deleted `GradeDist/gradeColors.ts` and `GradeDist/colors.tsx`
- Moved `PPCOverlayTrigger.tsx` into its own folder in `components/` for consistency

## Screenshots

Although most of this PR doesn't affect how the frontend looks, there are a few small things.

Before:
Screenshot 2025-06-03 at 6 21 16 PM

After:
Screenshot 2025-06-03 at 6 21 43 PM

## Test Plan

- [ ] Verify that all colors still appear as expected
- [ ] Verify that nothing else unexpectedly breaks/stops working
- [ ] Verify that the minor frontend changes are acceptable

## Issues

N/A


site/src/helpers/styling.ts
```diff
@@ -0,0 +1,38 @@
...
+        },
+      },
+    },
+  };
+};
+
+export const getCssVariable = (name: string): string => {
+  const bodyStyles = getComputedStyle(document.body);
+  const valueFromBody = bodyStyles.getPropertyValue(name).trim();
+  return valueFromBody || '#ffffff';
```
Awesome-E, line 37: This should mirror the behavior of getting a CSS variable, as the function name states. I believe it would be more correct to return an empty string instead of defaulting to white.

If you need a default value for special cases, you should instead take an optional second parameter `(..., defaultValue: string = '')` and return `valueFromBody || defaultValue`

site/src/helpers/styling.ts
```diff
@@ -0,0 +1,38 @@
...
+      fontSize: '1.2rem',
+      outline: 'none',
+      margin: 0,
+      padding: '0.25em 0.5em',
+      borderRadius: '2px',
+    },
+  },
+};
+
+export const getTheme = (darkMode: boolean): Theme => {
```
Awesome-E, line 17: I think this should be called `getChartTheme` since it uses properties specific to the chart.

site/src/component/PPCOverlayTrigger/PPCOverlayTrigger.tsx
Awesome-E, line 1: fine with the change, but this was always intended to be a stopgap popover wrapper and not long term, if that makes sense.

site/src/component/GradeDist/Pie.tsx
```diff
@@ -153,20 +143,15 @@ export default class Pie extends React.Component<PieProps> {
...
-        <strong>
-          {props.datum.id}: {((props.datum.value / this.total) * 100).toFixed(2)}%
-        </strong>
-      </div>
-    );
+    return <ChartTooltip label={props.datum.id} value={((props.datum.value / this.total) * 100).toFixed(2) + '%'} />;
   };
 
   render() {
+    const data = this.getClassData();
```
Awesome-E, line 150: what is the actual class data? Assuming it's a grade distribution, we should make these variables names more clear:

`const gradeCounts = ...`
`gradeCounts.map(grade => grade.color)`

site/src/component/QuarterTooltip/CourseQuarterIndicator.tsx
```diff
@@ -123,7 +123,7 @@ const CourseQuarterIndicator: FC<CourseQuarterIndicatorProps> = (props) => {
 
               {termsInOrder.includes(`${prevYear} Spring`) && (
                 <span>
-                  <span className={emojiSize}>🌸</span>
+                  <span className={emojiSize}>🌱</span>
```
Awesome-E, line 126: I don't think changing this emoji "to get rid of the pink color" is the right move here. If it's just for that, we should instead just replace the course offerings tooltip chart with the table as seen in recent offerings (since it's something we discussed anyways).

I suggest reverting this emoji change for the PR, since it may introduce a change that existing users would have to get used to.

site/src/helpers/courseRequirements.ts
Awesome-E, line 1: Nice, I like that you were able to find and use the same logic as present already!

site/src/index.css
```diff
@@ -36,14 +36,25 @@ code {
...
   --ring-road-white: #ffffff;
 
+  /* NOTE: unlike the secondary colors above, these secondary-light
+  colors aren't overridden by the dark mode variants defined in the
+  body below, since these are defined at the root, not in the body */
+  --blue-secondary-light: var(--blue-secondary);
+  --green-secondary-light: var(--green-secondary);
+  --yellow-secondary-light: var(--yellow-secondary);
+  --orange-secondary-light: var(--orange-secondary);
+  --red-secondary-light: var(--red-secondary);
```
Awesome-E, line 46: Perhaps "secondary light" isn't all that intuitive, but I do think the comment does a good job of explaining it. You don't need to say `NOTE:` at the beginning of the comment though, since well, a comment is a note.

Awesome-E (changes requested):
Not bad all around. a couple of considerations though

site/src/index.css
charlieweinberger: I agree its not great, but I couldn't think of a better naming system that wasnt too verbose. Lmk if you have any suggestions

Awesome-E: yeah what you have is fine.

Awesome-E (commented):
pretty much all good, just one thing

Awesome-E (approved):

PR #718, opened 2025-06-04
simplified course quarter offerings indicator

charlieweinberger:
## Description

Simplified the logic of this file by moving more of the logic out the JSX elements themselves. 

Note: this does use 🌱 instead of 🌸, so it should wait for #717 to be merged.

## Screenshots

N/A

## Test Plan

- [ ] Verify that the application doesn't change at all because of this PR
- [ ] Verify that the new file is simpler/better than the existing code

## Issues

N/A

### DIFF
--- a/site/src/component/QuarterTooltip/CourseQuarterIndicator.tsx
+++ b/site/src/component/QuarterTooltip/CourseQuarterIndicator.tsx
@@ -9,6 +9,7 @@ interface CourseQuarterIndicatorProps {
 }
 
 const CourseQuarterIndicator: FC<CourseQuarterIndicatorProps> = (props) => {
+  // determine the size of all emojis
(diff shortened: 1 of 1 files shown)


Awesome-E:
> I agree that converting to a table makes more sense... will put up a PR (a new one?) when I'm done/need feedback

I think it would be better to use `main` as the base branch and have a new PR (as you suggested) to change that to a table. Consider having the recent offerings table take in a `size` prop (defaults to `large`, can be set to `small`) to be consistent with MUI prop naming, and shrinking the texts like Caden said to "25-26" instead of "2025-2026" and "Year" instead of "Academic Year"

Awesome-E:
Once you put up the new PR, we will probably close this PR without merge.

PR #719, opened 2025-06-06
Created a `<LoadingSpinner />` component

charlieweinberger:
## Description

Created a shared `<LoadingSpinner />` component, defined once in `components/` and called everywhere where the current bootstrap `<Spinner />` is used. The only thing I'm not sure of for this PR is what the margin should be.

## Screenshots

N/A (some spinners did change from 8px to 20px, but its barely noticable.)

## Test Plan

- [ ] Verify that the application looks functionally the same
- [ ] Verify that nothing breaks

## Issues

N/A


site/src/component/LoadingSpinner/LoadingSpinner.scss
```diff
@@ -1,5 +1,5 @@
-.requirements-loading {
+.loading-spinner {
   display: flex;
   justify-content: center;
-  margin: 8px auto;
+  margin: 20px;
```
Awesome-E, line 4: `8px auto` for now, since its usage in visually smaller components (i.e. program requirements) will suffer more from bigger changes to the margin.

You can also override this in other places in their respective style sheets, i.e. `.planner > .loading-spinner { margin-block: 20px }` as appropriate

Awesome-E (commented):
Gap here looks a bit big, but otherwise fine
image

Awesome-E (commented):
Code changes look mostly fine to me. Would need to take a look at everywhere it's used though

CadenLee2 (commented):
I think the code changes are fine, there's not much to be opinionated about here

I did notice an inconsistency in loading spinner thickness between most places (top image) and the course search results in the sidebar (bottom image). Is it related to the gap/padding thing mentioned above?
image
image

charlieweinberger:
@CadenLee2 Nice catch, I didn't notice that difference. This is also b/c of the `SearchPlaceholder` component in `AllCourseSearch.tsx`, since `SearchPlaceholder` has a font size of 20px, which makes the spinner larger, since usually it uses the default font size of 16px. I just pushed a fix by making it so the CSS styling of `SearchPlaceholder` doesn't apply to the loading spinner, just the no results text.

CadenLee2 (approved):
I double-checked around the codebase and it doesn't look like you missed anything; all the changes look good. Remember to update from main, then you should be good to merge

PR #720, opened 2025-06-06
Created the `<NoResults />` component

charlieweinberger:
## Description

Throughout PeterPortal, there are many cases where users (1) wait for something to load, and then either (2) view the results of that search, or sometimes (3) don't get any results, and get an error message indicating so. This 3rd case happens in at least 2 instances, which currently use very similar code. This PR creates a new component named `NoResults` to be used in those 2 instances.

### Additional Discussion

There may be other places in the codebase that could use similar logic handling, but that don't fit this exact code. It depends on how we handle the logic between the loading spinner, the no results message, and the results. Some features don't need an error message when there are no results, such as the roadmap itself, which just uses the empty shell of a roadmap. Other features do need a no results message to indicate when loading is done, such as the roadmap sidebar course search, which would be confusingly blank without a no results message. Below are 3 options for handling this logic.

1. Treat all 3 separately, and render any number of the three components. For example, given a `loading` boolean and `results` list: if `loading` is true, show `LoadingSpinner`; if `results` has length 0, show a no results message; and if `results` doesn't have length 0, show the results. This is often the codebase's current logic, but this doesn't make sense, since one and only one component should appear at a time, but this logic allows potentially all three to be shown.
2. Treat all 3 separately, but only show 1 at a time, using ternary operators to render the correct component. This is ideal IMO, since it ensures only 1 component is rendered at a time and is the easiest to understand. This is one change I made in #721.
3. Instead of checking for both `LoadingSpinner` and `NoResults`, move `LoadingSpinner` inside of `NoResults`. This is the current logic of `SearchPlaceholder` in `AllCourseSearch.tsx`, but I think the spinner should be separate, since it's usually used on its own elsewhere.

## Screenshots

N/A

## Test Plan

- [ ] Verify that the application isn't changed at all (this PR just changes how the code is organized).
- [ ] Decide what is the best use of this component

## Issues

N/A

### DIFF
--- a/site/src/component/NoResults/NoResults.tsx
+++ b/site/src/component/NoResults/NoResults.tsx
@@ -3,14 +3,14 @@ import './NoResults.scss';
 import noResultsImg from '../../asset/no-results-crop.webp';
 
 interface NoResultsProps {
-  notSearching: boolean;
-  placeholderText: string;
+  showPrompt: boolean;
+  prompt: string;
 }
-const NoResults: FC<NoResultsProps> = ({ notSearching, placeholderText }) => {
+const NoResults: FC<NoResultsProps> = ({ showPrompt, prompt }) => {
   return (
     <div className="no-results">
       <img src={noResultsImg} alt="No results found" />
-      {notSearching ? placeholderText : "Sorry, we couldn't find any results for that search!"}
+      {showPrompt ? prompt : "Sorry, we couldn't find any results for that search!"}
     </div>
   );
 };
--- a/site/src/component/SearchHitContainer/SearchHitContainer.tsx
+++ b/site/src/component/SearchHitContainer/SearchHitContainer.tsx
@@ -55,10 +55,7 @@ const SearchHitContainer: FC<SearchHitContainerProps> = ({ index, CourseHitItem,
   return (
     <div ref={containerDivRef} className="search-hit-container">
       {noResults && (
-        <NoResults
-          notSearching={query === ''}
-          placeholderText={`Start typing in the search bar to search for ${index}...`}
-        />
+        <NoResults showPrompt={query === ''} prompt={`Start typing in the search bar to search for ${index}...`} />
       )}
       {searchInProgress && <LoadingSpinner />}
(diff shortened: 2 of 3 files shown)


site/src/component/NoResults/NoResults.tsx
```diff
@@ -0,0 +1,18 @@
...
+interface NoResultsProps {
+  notSearching: boolean;
+  placeholderText: string;
+}
+const NoResults: FC<NoResultsProps> = ({ notSearching, placeholderText }) => {
+  return (
+    <div className="no-results">
+      <img src={noResultsImg} alt="No results found" />
+      {notSearching ? placeholderText : "Sorry, we couldn't find any results for that search!"}
+    </div>
```
CadenLee2, line 14: Calling the prop `notSearching` seems a bit confusing because of the negation, especially since it's already for a component called `NoResults`. Since it's used when the user hasn't started taking an action at all, maybe something like `showPrompt` (and update `placeholderText` to be `prompt` to clarify that the two are related)?

CadenLee2 (commented):
Staging still works, code changes are mostly good, and I agree with the change since it's not fun seeing all that duplicate code

## Discussion
I agree with either (1) or (2) of the options you mentioned and I don't think it makes a major difference.
1. Personally I think `{searchInProgress && <LoadingSpinner />}` looks the nicest because it's compact and the conditional sections can each be their own line without interdependence, but you're right that it requires more thought about the conditions to prevent overlap.
2. One consideration with ternary operators is that logically they're nested (i.e. `a ? b : c ? d : e` -> `a ? (b) : (c ? d : e)`) but using them on multiple lines makes them look more flat, which could make them more confusing to read. However, they can create more clean logic. If we end up doing this, I like the structure of:
```ts
{searchInProgress ? (
  <LoadingSpinner />
) : shownCourses.length == 0 ? (
  /* No results */
) : (
  /* Show results */
)}
```
4. Option (3) feels odd because "loading" isn't really the same as actually having no results, so the spinner shouldn't be inside `NoResults`.

CadenLee2 (approved):
Looks good and still works!

PR #708, opened 2025-05-26
Refactor roadmap helpers

Awesome-E:
## Description

A partial refactor of functions in the helpers and roadmap slice files

## Screenshots

N/A

## Test Plan

Staging still works for all roadmap operations
- [ ] Editing/adding/deleting years, including add/remove quarters.
- [ ] Adding/dragging/removing courses on both desktop and mobile
- [ ] Prerequisite validation should still work

## Issues
N/A

PR #722, opened 2025-06-06
Display users' previously viewed roadmap on reload

### LINKED ISSUE #556: Similar to AntAlmanac, we should save the user's last selected roadmap
Currently, the 4-year plan loaded when you enter the page is always the first one saved to the user's account. This may not always be desired, so we should save which 4-year plan is currently active when the user saves their roadmap.

charlieweinberger:
## Description

A few days/weeks (idk?) ago, I spent like an hour trying to figure out the issue below, relating to saving the roadmap that the user last viewed and showing that whenever the user re-enters the application, instead of just showing the first roadmap. However, If I understand correctly, this involves a change to the database, so I wasn't sure if/how I would test it. Also, at least some of the related code has been refactored in the previous few days (see #708 ), so it's possible that most of this code is out-of-date, if it was every anything in the first place. Anyways, here it is, in all it's glory. Since it's been some time since I wrote this, I don't exactly remember how it all works, but like the other PRs I've put up recently, I'll come back to them by next week or so and finish them all, if not sooner.

## Screenshots

N/A (untested for now)

## Test Plan

- [ ] Verify that the roadmap that you most recently viewed is the one that loads when you reload the page. 

## Issues

Closes #556


Awesome-E (commented):
Looks good at first glance. I will say, though, we should be careful about database changes just because migrations aren't drizzle's strength. Nothing here looks wrong, but as a precautionary measure I think database changes should be discussed with leads.

api/src/controllers/roadmaps.ts
```diff
@@ -6,32 +6,37 @@ import { and, asc, eq, inArray, not } from 'drizzle-orm';
...
+
+    return {
       planners: planners as SavedPlannerData[],
       transfers: [],
-      timestamp: timestamp[0].timestamp?.toISOString(),
-    };
-    return roadmap;
+      timestamp: users[0].timestamp?.toISOString(),
+      currentPlanIndex: users[0].currentPlanIndex ?? undefined,
+    } as SavedRoadmap;
```
Awesome-E, line 33: `const roadmap: SavedRoadmap = {}` and `return {} as SavedRoadmap` are not the same type-checking. Prefer to define variables with type annotation instead of casting a return value

site/src/helpers/planner.ts
```diff
@@ -240,7 +241,7 @@ export async function upgradeLocalRoadmap(): Promise<SavedRoadmap> {
 }
 
 export const saveRoadmap = async (isLoggedIn: boolean, planners: SavedPlannerData[], showToasts: boolean = true) => {
-  const roadmap: SavedRoadmap = { timestamp: new Date().toISOString(), planners, transfers: [] };
+  const roadmap: SavedRoadmap = { planners, transfers: [], timestamp: new Date().toISOString(), currentPlanIndex: 0 };
```
Awesome-E, line 244: This just doesn't save the current plan index?

Awesome-E (changes requested):

api/src/controllers/roadmaps.ts
charlieweinberger: You're right, I thought they were the same but I understand now how they're not

api/src/controllers/roadmaps.ts
```diff
@@ -6,32 +6,38 @@ import { and, asc, eq, inArray, not } from 'drizzle-orm';
...
     if (planners.length === 0) {
       return undefined;
     }
+
     const roadmap: SavedRoadmap = {
       planners: planners as SavedPlannerData[],
       transfers: [],
-      timestamp: timestamp[0].timestamp?.toISOString(),
+      timestamp: users[0].timestamp?.toISOString(),
+      currentPlanIndex: users[0].currentPlanIndex ?? undefined,
```
Awesome-E, line 32: is this just to change `null` to `undefined`? Why is that necessary?

site/src/helpers/planner.ts
```diff
@@ -184,6 +184,7 @@ function addMultiPlanToRoadmap(roadmap: SavedRoadmap | LegacyRoadmap): SavedRoad
       ],
       transfers: roadmap.transfers,
       timestamp: roadmap.timestamp,
+      currentPlanIndex: roadmap.currentPlanIndex,
```
Awesome-E, line 187: This took a while to understand that this addition is purely to preserve the data in roadmap, and is thus a bit confusing imo. (My first thought was: "why is this even in a function that only converts something to multiplan?" in fact, that line will probably never be called since `'planners' in roadmap` should always be true for plans with a current index)

To me, it would make more sense to _only override_ one property, i.e. return `{ ...roadmap, planners: [...] }` to make it more clear that you're returning a "modified copy of the original" rather than a "completely new `SavedRoadmap`"

Awesome-E (commented):
Looks mostly good to me. I won't explicitly approve until the change is addressed or until after a second review goes through just so we don't accidentally merge early.

api/src/controllers/roadmaps.ts
charlieweinberger: The type `SavedRoadmap` says `currentPlanIndex` is type `number | undefined`, but `users[0].currentPlanIndex` is type `number | null`, so it must be converted to `undefined`

site/src/helpers/planner.ts
Awesome-E: I think that reworking < 10 lines of logic when you add to it is a more reasonable decision than adding 1 line of logic that will need to be reworked to scale later. You are changing the functionality overall, so I am in favor of making that change properly.

Awesome-E (commented):
As with the other PRs, I am willing to make some changes and merge after that.

charlieweinberger:
@Awesome-E I've kinda forgotten about how this change works/what it's actually doing, so if you wanna finish it up and push to main, its all yours 👍

anthonyj33:
Closing because this was implemented in #983

PR #724, opened 2025-06-06
Redesign Planner for Material UI

Awesome-E:
## Description

Redesigns the planner page according to our Figma design. With the exception of modals, the planner year/quarter components no longer rely on react bootstrap components.

## Screenshots

These are subject to change.

Before|After
:-:|:-:
image|image

## Test Plan

- [ ] Ensure everything looks right on staging in both light and dark mode
- [ ] Check whether operations to edit/add/remove years/quarters/courses still work

## Issues

N/A yet

### DIFF
--- a/site/src/component/ExpandMore/ExpandMore.tsx
+++ b/site/src/component/ExpandMore/ExpandMore.tsx
@@ -1,0 +1,22 @@
+import { IconButton, IconButtonProps, styled } from '@mui/material';
+
+interface ExpandMoreProps extends IconButtonProps {
+  expand: boolean;
+}
+
+export const ExpandMore = styled((props: ExpandMoreProps) => <IconButton {...props} />)(({ theme }) => ({
+  marginLeft: 'auto',
+  transition: theme.transitions.create('transform', {
+    duration: theme.transitions.duration.shortest,
+  }),
+  variants: [
+    {
+      props: ({ expand }) => !expand,
+      style: { transform: 'rotate(0deg)' },
+    },
+    {
+      props: ({ expand }) => !!expand,
+      style: { transform: 'rotate(180deg)' },
+    },
+  ],
+}));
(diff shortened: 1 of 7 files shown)


site/src/component/ExpandMore/ExpandMore.tsx
```diff
@@ -0,0 +1,22 @@
+import { IconButton, IconButtonProps, styled } from '@mui/material';
+
+interface ExpandMoreProps extends IconButtonProps {
+  expand: boolean;
+}
```
CadenLee2, line 5: The name `expand` feels a bit contradictory, since it seems that it's *false* when collapsed (and the button is thus used to *expand*) and true when expanded (and the button is thus used to *collapse*).

I think it would make more sense to rename it to `expanded`.

site/src/pages/RoadmapPage/Year.tsx
```diff
@@ -22,172 +100,89 @@ interface YearProps {
...
+          <ExpandMore
+            expand={showContent}
+            onClick={() => setShowContent(!showContent)}
+            aria-expanded={showContent}
+            aria-label="show more"
           >
-            Remove
-          </Button>
+            <ExpandMoreIcon />
+          </ExpandMore>
```
CadenLee2, line 135: Is there a reason for putting the icon as a child here rather than putting it in `ExpandMore` itself?

site/src/pages/RoadmapPage/Quarter.tsx
```diff
@@ -102,64 +89,30 @@ const Quarter: FC<QuarterProps> = ({ year, yearIndex, quarterIndex, data }) => {
...
-            onClick={() => {
-              dispatch(clearQuarter({ yearIndex: yearIndex, quarterIndex: quarterIndex }));
-              setShowQuarterMenu(false);
-            }}
-          >
-            Clear
-          </Button>
-          <Button
-            variant={buttonVariant}
-            className="quarter-menu-btn red-menu-btn"
```
CadenLee2, line 121: It looks like `red-menu-btn`, `quarter-menu-btn`, and `quarter-menu-popover` aren't used anywhere else, but are still defined in `Quarter.scss`, so that should be removed

site/src/component/ExpandMore/ExpandMore.tsx
Awesome-E: I don't fully agree that `expand` is contradictory, but I do think `expanded` is a better name. Changed that

PR #725, opened 2025-06-13
Fix size of roadmap course being dragged on Safari

Awesome-E:
## Description

For some reason, Safari's "fit content" thought that for dragged roadmap courses, that meant making it the full screen height. This PR fixes that by simply setting the grid row height to the content's height.

## Screenshots

Before:

After:

## Test Plan

- [ ] Check that this works on Staging on Safari hello????

## Issues

N/A

### DIFF
--- a/site/src/App.tsx
+++ b/site/src/App.tsx
@@ -29,9 +29,7 @@ export default function App() {
       <AppThemeProvider>
         <AppHeader />
         <div className="app-body">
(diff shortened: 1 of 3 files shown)


charlieweinberger (approved):
Was able to reproduce the bug on safari, and then observe that this PR doesn't have the bug anymore. LGTM 👍

PR #726, opened 2025-06-13
Fix Automatic Quarter Sizing

Awesome-E:
## Description

Previously, automatically making x number of quarters take up the full container width depended on exact, hard-coded values. This PR changes it so that the breakpoints now have some tolerance (instead of relying on a pixel-perfect value) and creates logic used to determine these breakpoints.

This is made easier by the mui planner redesign as there is no longer a gap between quarters to account for.

## Screenshots

image

## Test Plan

- [ ] Make sure that for different numbers of quarters in your roadmap (i.e. all years with 1 quarter, some with 2 some with 3, some with 3 and 4, etc.), the year with the most quarters never has a gap to the right when you increase the screen size. 

Only the year with the most number of quarters should stretch quarters to be the full width. Other years may have a gap because quarters should all be the same width. This is the same as the previous intended behavior.

## Issues
N/A

### DIFF
--- a/site/src/pages/RoadmapPage/Year.scss
+++ b/site/src/pages/RoadmapPage/Year.scss
@@ -1,5 +1,7 @@
 @use '../../globals';
 
+$base-quarter-width: 250px;
+
 [data-theme='dark'] {
   .year-settings-btn:hover,
   .year-settings-btn:focus,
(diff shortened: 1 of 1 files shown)


charlieweinberger (approved):
LGTM, changes seem to work as intended, nothing looks wrong, and code looks well written. Nice use of for loops in SASS!

PR #728, opened 2025-06-14
Make (uid,courseName) in misc transfers unique

Awesome-E:
## Description

As the title states, this PR makes it so that each user id/course name pair must be unique for miscellaneous transfers. Previously, this could not be done since users could add duplicates with the same name. However, since there is no longer a way to add miscellaneous transfers, and since duplicates have been merged via our Redesigned Transfers migration script, this is now possible to enforce.

Note: Before merging and running drizzle migrate, run the transfer data migration script again on prod because there are transcripts imported between the time when Redesigned Transfers released and when Imports were updated to use the new transfers.

## Screenshots

N/A

## Test Plan

- [ ] Ensure nothing breaks on staging
- [ ] Ensure that on the dev db, you can't add rows with the same user id and course name anymore.

## Issues

N/A

### DIFF
--- a/api/drizzle/0013_modern_rhino.sql
+++ b/api/drizzle/0013_modern_rhino.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "transferred_misc" ADD CONSTRAINT "transferred_misc_user_id_course_name_pk" PRIMARY KEY("user_id","course_name");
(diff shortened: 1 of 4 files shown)


CadenLee2 (approved):
Tested on staging and everything works still (adding misc via transcript import, deleting).

Importing a transcript continues to successfully avoid adding duplicate misc transfers, so as long as the migration script is run there should be absolutely no way for that situation to happen.

PR #731, opened 2025-06-16
Add Question Marks to Future Offerings in Recent Offerings Table

fiveminus1:
## Description

Added question marks to future offerings in the `RecentOfferings` table component. Queries for the current term, then checks for any future terms in the table and marks them as a question mark. Additionally, should fix the vertical alignment on checks/question marks.

## Screenshots

Before|After
:-:|:-:
image|image

## Test Plan

- [ ] Courses that were offered this year should now have a question mark for offerings that are in the future. (e.g. anything after Fall 2025)
- [ ] Courses not offered this year shouldn't be affected.

## Issues

N/A

### DIFF
--- a/api/drizzle/0013_modern_rhino.sql
+++ b/api/drizzle/0013_modern_rhino.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "transferred_misc" ADD CONSTRAINT "transferred_misc_user_id_course_name_pk" PRIMARY KEY("user_id","course_name");
--- a/api/drizzle/meta/0013_snapshot.json
+++ b/api/drizzle/meta/0013_snapshot.json
@@ -1,0 +1,1054 @@
+{
+  "id": "2dcd9aac-1652-4182-bcc0-1d3d98c5a822",
+  "prevId": "1efc8066-c883-4758-a6bd-75cbf6bdb2cb",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.completed_marker_requirement": {
+      "name": "completed_marker_requirement",
+      "schema": "",
+      "columns": {
+        "user_id": {
+          "name": "user_id",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "marker_name": {
+          "name": "marker_name",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        }
+      },
+      "indexes": {},
+      "foreignKeys": {
+        "completed_marker_requirement_user_id_user_id_fk": {
+          "name": "completed_marker_requirement_user_id_user_id_fk",
+          "tableFrom": "completed_marker_requirement",
+          "tableTo": "user",
+          "columnsFrom": ["user_id"],
+          "columnsTo": ["id"],
+          "onDelete": "no action",
+          "onUpdate": "no action"
+        }
+      },
+      "compositePrimaryKeys": {
+        "completed_marker_requirement_user_id_marker_name_pk": {
+          "name": "completed_marker_requirement_user_id_marker_name_pk",
+          "columns": ["user_id", "marker_name"]
+        }
+      },
+      "uniqueConstraints": {},
+      "policies": {},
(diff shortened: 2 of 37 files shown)


site/src/component/RecentOfferings/RecentOfferings.scss
```diff
@@ -12,7 +12,7 @@
   width: 100%;
   max-width: 540px;
   .bi-check {
-    transform: scale(1.5) translateY(-1px);
+    transform: scale(1.5);
   }
```
Awesome-E, line 16: I think we should remove this entirely, as `bi-check` is a class name for **b**ootstrap **i**cons

site/src/component/RecentOfferings/RecentOfferings.scss
```diff
@@ -23,4 +23,8 @@
   :is(th, td):not(:first-child) {
     text-align: center;
   }
+
+  :is(td):not(:first-child) {
+    transform: translateY(-2px);
+  }
```
Awesome-E, line 29: use `.MuiSvgIcon-root` to select icons

site/src/helpers/util.tsx
```diff
@@ -127,6 +127,10 @@ export const sortTerms = (terms: string[]) =>
     return yearB.localeCompare(yearA) || quartersOrdered[qtrB].localeCompare(quartersOrdered[qtrA]);
   });
 
+export const isTermAfter = (termA: string, termB: string): boolean => {
```
Awesome-E, line 130: ordering of parameters is a bit confusing

site/src/component/RecentOfferings/RecentOfferings.tsx
```diff
@@ -66,14 +78,29 @@ const RecentOfferings: FC<RecentOfferingsProps> = (props) => {
...
+                    return (
+                      <td key={index}>
+                        {offered ? (
+                          <CheckIcon style={{ fontSize: 20 }} />
+                        ) : isFutureTerm ? (
+                          <QuestionMarkIcon style={{ fontSize: 20, color: theme.palette.text.secondary }} />
+                        ) : null}
+                      </td>
+                    );
+                  })}
```
Awesome-E, line 100: Consider making a helper component`RecentOfferingQuarter` in the same file.

Awesome-E (changes requested):

site/src/component/RecentOfferings/RecentOfferings.tsx
```diff
@@ -66,14 +110,24 @@ const RecentOfferings: FC<RecentOfferingsProps> = (props) => {
...
-                {quarters.map((offered, index) => (
-                  <td key={index}>{offered ? <CheckIcon style={{ fontSize: 20 }} /> : null}</td>
-                ))}
-              </tr>
-            ))}
+            .map(([year, quarters]) => {
+              return (
+                <tr key={year}>
+                  <td>{year}</td>
+                  {quarters.map((offered, index) => {
```
Awesome-E, line 117: you should be able to make this and the `return <tr .../>` a direct arrow return `(...) => <RecentOfferingsQuarter .../>`, but it's not that big of a deal

site/src/helpers/util.tsx
```diff
@@ -127,6 +127,10 @@ export const sortTerms = (terms: string[]) =>
     return yearB.localeCompare(yearA) || quartersOrdered[qtrB].localeCompare(quartersOrdered[qtrA]);
   });
 
+export const isTermAfter = (earlierTerm: string, laterTerm: string): boolean => {
```
Awesome-E, line 130: earlier and later kinda implies that the order is already known? sorry i know naming this kind of thing sucks, but I think something like `term` and `otherTerm` is the most standard thing you can do.

That way, it should read "is [term I am comparing] after otherTerm", and by that, it's obvious that `true` means `term` is after `otherTerm`. Does that make sense?

Awesome-E (approved):
Pretty much good! just two really small things you can just change yourself then merge!

PR #734, opened 2025-06-18
fix: question marks shouldn't be displayed on current summer term offerings

fiveminus1:
## Description

Some courses that do not have a Summer 2025 (or more specifically, a Summer term in the current year) offering are displaying "?". I mistakenly assumed this was something to do with that being the active term, but it's just because the `quartersOrdered` Record in `util.tsx` doesn't include "Summer" (as the offerings table doesn't differentiate for Summer 1/2,10wk). This PR just adds a "Summer" to the record to properly compare terms.   

## Screenshots

Before|After
:-:|:-:
image|image

## Test Plan

- [ ] Find a course that doesn't have a Summer 2025 offering (2024-25 AcaYear), confirm that it doesn't have a question mark for Summer anymore.
- [ ] Any courses that _are_ offered in Summer 2025 shouldn't be modified.

## Issues

N/A

### DIFF
--- a/api/src/controllers/reviews.ts
+++ b/api/src/controllers/reviews.ts
@@ -43,7 +43,6 @@ async function getReviews(
     verified?: boolean;
   },
   sessUserId?: number,
-  isAdminView?: boolean,
 ) {
   const { courseId, professorId, userId, reviewId, verified } = where;
   const userVoteSubquery = db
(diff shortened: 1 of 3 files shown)


Awesome-E:
sure I can take a look after. You can also `git revert` the commit hash for this branch

PR #735, opened 2025-06-19
Migrate the Roadmap Header Component to MUI

Awesome-E:
## Description

Migrates the roadmap header to use MUI components instead of React Bootstrap Components. This PR covers:
- Action buttons on the right, such as Save Planner
- The Dropdown button on the left to change which planner is shown
- The Dropdown menu buttons, including the ones to switch roadmaps and add new ones.

This PR does NOT cover modals, so modals triggered within the menu are not transitioned to MUI.

## Screenshots

Before|After
:-:|:-:
image|image

## Test Plan

- [ ] Ensure all the actions in the dropdown and the header still work on staging
- [ ] Make sure things look right visually on both desktop and mobile

## Issues

N/A

### DIFF
--- a/site/src/index.css
+++ b/site/src/index.css
@@ -1,5 +1,5 @@
 @import url('https://fonts.googleapis.com/css?family=Open+Sans&display=swap');
-@import url('https://fonts.googleapis.com/css?family=Roboto&display=swap');
+@import url('https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100..900;1,100..900&display=swap');
 
 html,
 body {
--- a/site/src/pages/RoadmapPage/AddYearPopup.tsx
+++ b/site/src/pages/RoadmapPage/AddYearPopup.tsx
@@ -1,11 +1,11 @@
 import { FC, useState } from 'react';
 import './AddYearPopup.scss';
-import { Button } from 'react-bootstrap';
 import YearModal from './YearModal';
 import { addYear, selectYearPlans } from '../../store/slices/roadmapSlice';
 import { useAppDispatch, useAppSelector } from '../../store/hooks';
 import { PlannerYearData } from '../../types/types';
 import AddIcon from '@mui/icons-material/Add';
+import { Button } from '@mui/material';
 
 const AddYearPopup: FC = () => {
   const [showModal, setShowModal] = useState(false);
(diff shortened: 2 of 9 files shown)


fiveminus1 (changes requested):
Mostly looks good to me!

Ran into a minor visual bug: if you create a roadmap with a long name (within 3 chars before you hit the limit, but this isn't always the case), it'll stretch some buttons down vertically. Interestingly, it's not consistent with the name cases: a single character or "abc" repeated doesn't cause an issue, "test" repeated is shown below, some combinations also resize the middle button.


This doesn't seem to be present on prod. 

image

CadenLee2:
> This doesn't seem to be present on prod.

I'm experiencing the visual bug on prod too (the name `Unofficial Transcript_testing` causes the issue on prod but not staging).

It appears to be based on the width of the container. As you repeat `a` more, the text in some of the buttons progressively has enough space to stop wrapping (notice `Blank Roadmap` is one line while `Zot4Plan Schedule` is two lines). This can be fixed by ensuring the buttons are always all the same height.

image
image

CadenLee2 (commented):
Some small issues:
- Button height issue (mentioned in above comment)
- The cursor changes from pointer to default if the roadmap multiplan dropdown is expanded, which feels a bit odd since it can still be clicked and everything else seems to always stay pointer (or maybe that's just how MUI does it, in that case it's probably fine)
- Clicking on a roadmap to select it will no longer dismiss the dropdown. Is this intentional?

Code generally looks good

Awesome-E:
There's now a breakpoint where every button switches to one line. This should address the visual bug mentioned.

image
image

Awesome-E:
> The cursor changes from pointer to default if the roadmap multiplan dropdown is expanded

I do think it's a bit weird but it is how MUI does it. Same as how AntAlmanac's switcher works, actually.

> Clicking on a roadmap to select it will no longer dismiss the dropdown. Is this intentional?

No, fixed that.

CadenLee2 (approved):
Looks good

fiveminus1 (commented):
image

Noticed this one now: if you have too short of a name, it'll cut off the right button. Deepest apologies Ethan

fiveminus1 (approved):
Let's get that money

PR #737, opened 2025-06-23
Redesign admin and review pages

CadenLee2:
Redesigned the Your Reviews page, the admin Verification page, and the admin Reports page to have a nicer layout, along with various changes to related components

## Description

Major changes include:
- Improved the aforementioned pages to conform better to the Figma
  - Added a reusable grid component for displaying review-like items
  - Mui buttons with icons for admin actions
  - Did not change the loading text to a spinner yet, since it'll be nicer to wait for Charlie's `<LoadingSpinner />`
- Improved subreview styles to mostly conform to the Figma
  - Subreviews now accept children that go next to the edit buttons (for Delete and Verify buttons)
  - Now uses Mui Paper for the background
  - Did not change the rating colors (the blue/green/yellow/etc.) yet, since we aren't completely set on those and should probably wait
- Added some global classes for Mui components (Mui Paper and Mui Button) in `App.scss` so they can be used in the future, because I saw a lot of recurrences in Figma and it seems this is the first time they're used in a consistent way

## Screenshots

| Before | After |
| --- | --- |
| image | image |
| image | image |
| image | image |
| Reason for subreviews now accepting children | image |

## Test Plan

If you don't have admin privileges on staging you can check out the branch and run locally to test instead

- [ ] Subreview and admin features should still work
  - [ ] Your Reviews: editing and deleting
  - [ ] Verification: deleting and verifying
  - [ ] Reports: ignoring and accepting
- [ ] The grid (see screenshots) should look good on all screen sizes
- [ ] If there are no action items on a page, it should still display a nice-looking message

## Issues

Closes #


fiveminus1 (commented):
Looks a lot better! 
- The double "Posted by" looks odd..is there a reason for that?
- I personally like the red/green for delete/verify used on the Figma better. Maybe the intention was that the gray/blue matches the site theme, but I feel like r/g differentiates an Admin action better.

Probably outside of this PR but some overall thoughts:
- I know this PR is mostly to match the Figma, but I really hate the "{COURSE} {ucinetid}" format on `SubReview` cards, we should change that in the future to show the full professor name.
- Is the "Report" menu broken? The "Review Content" portion is blank.
- I think I can just do this separately after this gets merged since it's related to my last PR, but it'd be nice to have the review author un-anonymized in the Reports page as well.

CadenLee2:
> I personally like the red/green for delete/verify used on the Figma better. Maybe the intention was that the gray/blue matches the site theme, but I feel like r/g differentiates an Admin action better.

I think the idea was that the icons represent the action instead of the color (like trash can for delete, check mark for verify). We can think about this more in the future when dealing with other color changes

> it'd be nice to have the review author un-anonymized in the Reports page as well

Yeah that'll be nice to do separately

> I really hate the "{COURSE} {ucinetid}" format on SubReview cards, we should change that in the future to show the full professor name

Same. Looked into this more and it seems the SubReview cards obtain their information from whoever's using them (i.e. if they're displayed on the course page, the course page passes course GQL data as a prop, which includes professor names, and as a result the name shows up properly there). Notice how it displays only the prof name for ICS 33 reviews, which is fine:

image

I'll put up an issue for this (improving the format of non-course/prof-page SubReview card identifiers)

fiveminus1 (approved):

site/src/component/Report/Reports.tsx
```diff
@@ -56,30 +80,19 @@ const Reports: FC = () => {
...
-  }
+  return (
+    <div className="content-wrapper reports-container">
+      <h1>User Review Reports</h1>
+      <p>
+        Accepting a report will <b>delete</b> the review. Ignoring a report will <b>preserve</b> the review.
+      </p>
+      {reportsLoading ? (
+        <p>Loading...</p>
+      ) : (
```
Awesome-E, line 91: iirc the Loading spinner PR is not fully merged yet? because if it is, I'd say to use that instead of the text "Loading..."

site/src/component/ReviewItemGrid/ReviewItemGrid.scss
```diff
@@ -0,0 +1,14 @@
...
+}
+
+.reviews-grid > div {
+  width: 100%;
+  height: 100%;
+}
+
+.reviews-grid > div > * {
+  height: 100%;
+}
```
Awesome-E, line 14: I believe you could instead just use the `justify-items` and `align-items` properties instead of setting them on the child. Also, what is the direct child `div` for?

site/src/component/Verify/Verify.tsx
```diff
@@ -32,33 +23,58 @@ const Verify: FC = () => {
...
-          <div key={`verify-${i}`} className="verify">
-            <br />
-            <SubReview review={review}></SubReview>
-            <div className="verify-footer">
-              <Button variant="danger" className="mr-3" onClick={() => deleteReview(review.id)}>
-                Delete
+  return (
+    <ReviewItemGrid>
+      {reviews.length == 0 && <span>There are no unverified reviews</span>}
+      {reviews.map((review) => (
```
Awesome-E, line 29: This still feels like a lot of stuff for a `.map()`. consider making an `UnverifiedReview` component within the `AdminPage` folder and calling that.

site/src/component/Verify/Verify.tsx
```diff
@@ -32,33 +23,58 @@ const Verify: FC = () => {
...
+      <p>
+        Verifying a review will display the review on top of unverified reviews. Deleting a review will remove it
+        permanently.
+      </p>
+      {reviewsLoading ? <p>Loading...</p> : <UnverifiedReviewsList reviews={reviews} />}
+    </div>
+  );
 };
 
 export default Verify;
```
Awesome-E, line 80: could benefit from a renaming to something like `UnverifiedReviewList`

site/src/pages/ReviewsPage/UserReviews.tsx
```diff
@@ -3,41 +3,48 @@ import SubReview from '../../component/Review/SubReview';
...
+const UserReviewsList: FC<{ reviews: ReviewData[] }> = ({ reviews }) => {
+  return (
+    <ReviewItemGrid>
+      {reviews.length == 0 && (
+        <span>You haven't reviewed any courses yet. Look up a course you've taken to review it!</span>
+      )}
+      {reviews.map((review) => (
+        <div key={'user-review-' + review.id!}>
+          <SubReview review={review} />
+        </div>
```
Awesome-E, line 19: (mentioned in a previous comment) is there a reason we can't just map directly to the `SubReview`?

Awesome-E (changes requested):

site/src/component/Verify/Verify.tsx
CadenLee2: Agree, renamed

fiveminus1:
> It looks a bit odd when a title wraps since the buttons aren't vertically centered with the text, but it also would look weird if they're vertically centered and not in the same top-right position as on all the other cards. Thoughts?

I think it's acceptable as an admin page, I would prefer it stay in the top-right position. Keep in mind that my PR #742 will probably make all of the identifiers long enough to break into a newline though..maybe I can make it consistent in mine so it'll keep the professor on the newline (without a separator) if we're okay with that?

site/src/component/Review/SubReview.tsx
```diff
@@ -149,30 +150,33 @@ const SubReview: FC<SubReviewProps> = ({ review, course, professor }) => {
...
+                  </Button>
+                  <Button variant="danger" onClick={() => deleteReview(review.id!)}>
+                    Delete
+                  </Button>
+                </Modal.Footer>
+              </Modal>
+            </>
+          )}
+          {children}
+        </div>
```
Awesome-E, line 179: maybe this should be a separate component in the same file?

site/src/pages/AdminPage/Reports/Reports.tsx
```diff
@@ -1,17 +1,40 @@
...
+const ReportsList: FC<{
+  data: ReviewDisplay[];
+  acceptReports: (reviewId: number) => Promise<void>;
+  denyReports: (reviewId: number) => Promise<void>;
+}> = ({ data, acceptReports, denyReports }) => {
+  return (
+    <ReviewItemGrid>
+      {data.length == 0 && <span>There are currently no reports that need attention</span>}
+      {data.map((reviewPair) => (
+        <ReportGroup
```
Awesome-E, line 23: the name `ReportGroup` was a bit vague until I saw the staging instance but since that's not the concern of this PR i think it's fine

Awesome-E (approved):
reviewed in a bit of a crunch as you know but tentatively approved

PR #739, opened 2025-06-26
Automatically fetch saved specialization requirements 

fiveminus1:
## Description

Using the major specialization redux storage from #674, this PR ensures that the roadmap sidebar automatically loads the requirements from a found specialization. 

## Screenshots

fix-specs

## Test Plan

- [ ] Add a major specialization to a roadmap and confirm that the specialization requirements persist upon reloading/switching tabs.

## Issues

N/A


CadenLee2 (approved):
Clarification note: the exact issue was that the specialization requirements *specifically* (such as the extra 4 classes for a CS specialization, in the image) weren't getting loaded properly upon reload/switching tabs
image

Let's get that MONEY 👍

PR #740, opened 2025-06-27
Improved code version of #737

charlieweinberger:
## Description

~~Same output as #737, but with less duplicated code, making use of a helper component. Should only be merged after #737 is merged.~~ Edit: We can just add these changes directly to #737, and then close this PR as not planned.

## Screenshots

N/A

## Test Plan

- [ ] make sure the application is functionally the same as #737

## Issues

N/A


charlieweinberger:
> I agree ...

Me when I agree with someone on something: 🕺😄

> I think it made more sense to keep UserReviews.tsx under pages/ReviewPage since that's the only place where it's used. There's a lot of precedent for that in the Roadmap page file structure as well

You're correct in that the `RoadmapPage/` and `SearchPage/` folders have several components within them that are only used in those folders. However, all the other subfolders within the `pages/` folder don't have any components; their components are instead in the `components/` folder.

For example, according to your logic, `Reports.tsx` and `Verify.tsx` should both be in `pages/AdminPage/`, instead of currently being in `components/`.

These are two different approaches, but they're both logical IMO, our codebase currently uses both. I'm fine with either approach, but I do think we should pick one approach and try to stick to it. However, that's a decision that affects the entire codebase, so we can discuss that another time. For this PR, I'll stick with your preference.

> I think `ReviewsGrid` is an unclear name because it suggests it's just the grid, and it's just for reviews. In reality it acts as a sort of template for those types of pages, and just happens to display stuff in a grid. Instead maybe like `ReviewsGridPageTemplate`? That name's kinda unwieldy but just something more clear would be nice

I see your point, and I'm also not sure what's best. I think I like `ReviewGridTemplate`, what do u think? IMO "review" should be singular, since "grid" already implies that there's multiple reviews, and "page" seems like the least important of the 4 words, and I'd prefer to only have 3 words, although I'm like 51/49 on this so idk.

> ReviewsGrid shouldn't need the noData prop because it can just use one of React's child helpers to check whether there are zero children

charlieweinberger:
@Awesome-E see #751, that PR has the latest changes/details

PR #741, opened 2025-06-28
Show reported reviewers' names in admin view

fiveminus1:
## Description

Per conversation in #737, this PR uses the admin route created in #733 to also show the name of anonymous reviewers being reported when viewing reports.

## Screenshots

Before|After
:-:|:-:
image|image

## Test Plan

- [ ] Make an anonymous review, report it, then check the admin `Reports` page and ensure the reported review is not anonymized.
- [ ] Confirm nothing else was changed.

## Issues

N/A


CadenLee2 (approved):
Solid

It looked a bit suspicious when testing because one review with the name Anonymous Peter did show up in reports, but turns out that's the user's actual name in the dev database

Feel free to merge this now since we can update the admin/review page redesign branches with the change

PR #742, opened 2025-06-29
Improve ReviewCard identifiers 

### LINKED ISSUE #738: Improve SubReview card identifiers that display both prof and course information
On professor pages, the title of each review is the reviewed course's name (because we already know which course the review is for). Similarly, on course pages, the title of each review is the reviewed professor:

Image

But when verifying reviews, looking at reports, or looking at your own reviews, we have to display both the course and the professor since otherwise it'd be ambiguous. Currently it's in an ugly format with the course ID and the professor's netid concatenated together:

Image

In this case, we should improve how we show the course (`COMPSCI204` -> `COMPSCI 204`) and the professor's name (`habibaf` -> `Full Name`).
- This might involve calling our `professor.get` TRPC route when the `<SubReview />` card recognizes that it doesn't have enough information
- Since this would make the identifier a much longer line, consider splitting it up into two lines somehow or otherwise changing it to look nicer

Wait until #737 gets merged to address this issue to avoid conflicts

fiveminus1:
## Description

This PR improves the display of `ReviewCard` (renamed from `SubReview`) identifiers in non-course/non-professor pages (such as Your Reviews/Verification) to display both the full course ID as well as the full professor name. 

It should also improve the course ID display by adding a space between the course department and number to better match the Professor page identifiers. Should also note that because `course_id`s within `review` columns don't account for spaces within department names, Your Reviews identifiers won't necessarily match 1-1 with Professor identifiers (i.e. on Your Reviews, it'll say I&CSCI but on a Professor page, it'll say I&C SCI). 

Additionally, introduced a `<Skeleton>` component while fetching the identifier. It's up for future discussion whether we should use Skeletons in other part of the app over loading screens, but @Awesome-E's idea is that they fit best for secondary loading in cases like this.

## Screenshots

Before|After
:-:|:-:
image|image

## Test Plan

- [ ] Your Reviews/Verification ReviewCards look better
- [ ] ReviewCards in Course/Professor pages still look the same
- [ ] **Make sure nothing broke from merging conflicts** 

## Issues

Closes #738


site/src/component/Review/ReviewCard.tsx
```diff
@@ -172,38 +218,19 @@ const SubReview: FC<SubReviewProps> = ({ review, course, professor, children })
...
-            <div>
-              <Link to={{ pathname: `/course/${review.courseId}` }}>{review.courseId}</Link>{' '}
-              <Link to={{ pathname: `/professor/${review.professorId}` }}>{review.professorId}</Link>
-            </div>
-          )}
+    <Card variant="outlined" className="reviewcard">
+      <div className="reviewcard-header">
+        <h3 className="reviewcard-identifier">
+          {loadingIdentifier ? <Skeleton variant="text" animation="wave" width={210} /> : identifier}
         </h3>
```
CadenLee2, line 225: I like how the skeleton's used here, it feels much more natural to wait while the content is loading

site/src/helpers/util.tsx
```diff
@@ -149,3 +149,9 @@ export function pluralize(count: number, pluralText: string = 's', singularText:
 export function getCourseIdWithSpaces(course: Pick<CourseGQLData, 'department'> & Pick<CourseGQLData, 'courseNumber'>) {
   return `${course.department} ${course.courseNumber}`;
 }
+
+export function splitCourseId(courseId: string): [string, string] {
+  // splits a course id into its department and number
+  const courseMatch = courseId.match(/^([A-Za-z&]+)(\d.*)$/);
+  return courseMatch ? [courseMatch[1], courseMatch[2]] : [courseId, ''];
+}
```
CadenLee2, line 157: This technically has a bug with `IN4MATX` because it sees the `4`, which is not `A-Za-z&`:
image
I'm not sure if it's possible to create a fully bulletproof solution, though, since there are also things like `ICS45C` that also have numbers embedded in the middle, making it hard.

Maybe going from the back and identifying the start of the last number (i.e. `ICS | <-- 45 <-- | C <---`)? Since this is only used in few places it might be ok to just add a comment that says "Use with caution as this may give the wrong output in some cases; e.g. IN4MATX41 -> IN 4MATX41".

You could also hardcode `IN4MATX` as an exception (seems to be the only one with a number if [the catalogue](https://catalogue.uci.edu/allcourses/) is accurate), but that seems inelegant

CadenLee2 (commented):
Review cards look much better now! I see report identifiers aren't handled yet, so assuming that's intentional (makes sense given the pr title), I think it's okay for now since it's never user-facing other than for admins.

Found one small issue with the regex for splitting course ID

site/src/helpers/util.tsx
fiveminus1: i hate regexes...newest push should fix those two cases

```
// department can consist of letters, digits, and an optional &
// number consists of digits and an optional letter
```
took a quick glance at the catalogue, afaik there are no other department names that should break these?

CadenLee2: Yeah I think the "correct" solution here is to just fetch the course info via api like how it currently fetches the professor name in `fetchProfName`, then use the resulting `department` and `courseNumber` from that. Ideally this should be easy, just one more helper and `await`.

If that's too much work though, or if it slows down the loading too much, I think for now this regex solution is alright. The function name `splitCourseId` makes it clear enough that it's just for splitting it up, not for getting the exact spacing. Changing the stored review identifier in the db is definitely overkill imo

CadenLee2 (approved):
Much better! Looks like checking the courses of the professors first was a good move, not a single course needed to be fetched:
image

Awesome-E (approved):
let's get that money 🤑

PR #748, opened 2025-07-09
Improve code readability w/ ternary operators & guard clauses

charlieweinberger:
## Description

Implemented ternary operators and guard clauses in a handful of places throughout the codebase (mostly relating to loading states) to improve code readability/understandability. Could have done more, but didn't want to make this PR too large.

## Screenshots

N/A

## Test Plan

- [ ] Verify that code is more readable
- [ ] Verify no functionality changes

## Issues

N/A


site/src/component/AppHeader/Profile.tsx
Awesome-E, line 1: indifferent on old vs new implementation for this file

site/src/component/SearchHitContainer/SearchHitContainer.tsx
Awesome-E, line 1: I like that you reordered the precondition to the top, but I personally find it extremely difficult to read code in ternary operators (especially with any nesting). This isn't your fault, I just think the way parens and line breaks work with the formatter causes it to be very difficult to see that there is conditional logic at all.

technically yes it is less computations this way, but I find line by line to be more readable because it has more consistent indentation. to me, this is a LOT more readable than any sort of ternary because you can see what exact case is associated with what rendered code purely by looking at the first indent level:
```tsx
<>
  {searchInProgress && <LoadingSpinner />}
  {!searchInProgress && results.length === 0 && <NoResults ... />}
  {!searchInProgress && results.length > 0 && (
    <div>...</div>
  )}
</>
```
yes, it does come with the slight overhead of verifying that the conditions partition all possible states, but that's pretty simple when there's only two conditions. Anything more than that probably deserves its own component anyways

site/src/pages/RoadmapPage/sidebar/AllCourseSearch.tsx
Awesome-E, line 1: I think the ordering here is better but otherwise indifferent on this file

site/src/pages/RoadmapPage/sidebar/MinorCourseList.tsx
Awesome-E, line 1: For the same reason as the other file, I don't find this to necessarily be better. The "else" with the `:` is at least consistently indented, but I personally don't find this any more readable than before.

site/src/pages/SearchPage/CourseHitItem.tsx
```diff
@@ -76,18 +76,9 @@ const CourseHitItem: FC<CourseHitItemProps> = (props) => {
...
-            )}
-            {courseIsSaved && (
-              <IconButton onClick={(e) => onUnsaveCourse(e)} size={'small'}>
-                <BookmarkIcon />
-              </IconButton>
-            )}
-          </div>
+          <IconButton onClick={(e) => (courseIsSaved ? onUnsaveCourse(e) : onSaveCourse(e))} size="small">
+            {courseIsSaved ? <BookmarkIcon /> : <BookmarkBorderIcon />}
+          </IconButton>
```
Awesome-E, line 81: this creates an implicit dependence between conditionals, which seems to me like an anti-pattern. It's more concise, but I think it'd be nice to think of it more like a "toggle button" with a state-dependent icon as opposed to a button with state-dependent functionality and a state-dependent icon.

My ideal solution for something like this would be to combine the onunsave/save functions, use that as a single handler, and then render the icon conditionally. This also isolates function logic to their definitions instead of also being changeable both inline and in definitions (if you can swap which handler is being called, now you have to check two handlers to see if one or both are misbehaving)

```suggestion
          <IconButton onClick={toggleSaveCourse} size="small">
            {courseIsSaved ? <BookmarkIcon /> : <BookmarkBorderIcon />}
          </IconButton>
```

Awesome-E (changes requested):
Neutral on this PR, but I am happy to keep some of the changes and apply them to main. I can finish up the PR by removing/rewriting some of your changes and merging the rest into main.

Open to hearing your thoughts, but will otherwise get confirmation from @CadenLee2 and proceed with the above.

charlieweinberger:
@Awesome-E Just implemented the `toggleSaveCourse` suggestion, good idea. Other than that, I guess my changes are more of my opinion than I thought 😅. Anyways, nothing here is a big deal, so you can go ahead and make whatever changes you'd like, then merge with main 👍

Awesome-E (approved):

PR #750, opened 2025-07-22
Fix professor page sidebar position

CadenLee2:
The professor page sidebar had an issue with its size and position that caused it to overflow on large screens. This fixes the issue

## Description

Restored a removed section of code that set the width of one of the sidebar's elements to `100vw` as a fix to force adaptive width to work

## Screenshots

| Before | After |
| -- | -- |
| image | image |

## Issues

Does not close any issues

### DIFF
--- a/site/src/component/SideInfo/SideInfo.scss
+++ b/site/src/component/SideInfo/SideInfo.scss
@@ -31,6 +31,16 @@
   }
 }
 
+.side-info-name {
+  // Force adaptive width to work while keeping the name width the same
+  display: flex;
+  align-items: center;
+
+  h2 {
+    width: 100vw;
+  }
+}
+
 .course-tags {
   display: flex;
   flex-wrap: wrap;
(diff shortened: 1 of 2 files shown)


site/src/component/SideInfo/SideInfo.scss
```diff
@@ -31,6 +31,16 @@
   }
 }
 
+.side-info-name {
+  // Force adaptive width to work while keeping the name width the same
```
Awesome-E, line 35: it would be helpful if the comment was adjusted more to indicate where this issue occurs. I'm not sure if the "while keeping the name width the same" is necessary, but you can keep that if you want
```suggestion
  // Force adaptive width on the professor page to work
```

site/src/component/SideInfo/SideInfo.tsx
```diff
@@ -146,7 +146,9 @@ const SideInfo: FC<SideInfoProps> = (props) => {
     <div className="side-content-wrapper">
       <div className="side-info">
         <div>
-          <h2>{props.name}</h2>
+          <div className="side-info-name">
```
Awesome-E, line 149: so there used to be a wrapper div here? or was the (flex) style previously applied to the div without a class name?

Awesome-E (approved):

PR #745, opened 2025-07-02
Unified admin page

charlieweinberger:
## Description

- PeterPortal currently has two admin pages, one for verifying user reviews and one for viewing review reports. This PR merges those two pages into one page, using tabs to navigate between the two subpages.
- This branch is a branch of #740, which itself branches from #737, which is why there are 21 files changed. However, this PR itself mostly just changes the 2 files in `AdminPage/`, as well as parts of `ReviewGridTemplate/`, `Verify/`, and `Report/`.

### Future PRs

- This PR's admin tabs in `AdminPage/index.tsx` are very similar to `RequirementsListSelector` in `RoadmapPage/SearchSidebar.tsx`. In a future PR, they should be combined into a dedicated `Tabs` component.
- `RequirementsListSelector` is the only place where the `ppc-btn` CSS class is used, so in a future PR, `ppc-btn` should be merged into bootstrap/MUI's primary styling.

## Screenshots

Before:
Screenshot 2025-07-03 at 9 56 00 AM

After:
Screenshot 2025-07-16 at 11 57 13 PM

Alternative Design, using MUI tabs:
Screenshot 2025-07-01 at 7 22 43 PM

## Test Plan

- [ ] Ensure the new page design looks good, or at least better than the existing two pages
- [ ] Ensure both pages are still accessible
- [ ] Ensure the verify and report pages themselves are unchanged

## Issues

N/A

PR #752, opened 2025-07-29
Reopen admin redesign w/ clean git history

Awesome-E:
@charlieweinberger let me know if anything got lost in the changes

---

Same as #745


charlieweinberger (changes requested):
After a quick look, the changes from #740/#751 look good. Nice!

However, the changes from #745, that merges the verifications/reports pages into one page with two tabs, is still missing. Is that intended to be in this PR?

Awesome-E:
> However, the changes from https://github.com/icssc/peterportal-client/pull/745, that merges the verifications/reports pages into one page with two tabs, is still missing. Is that intended to be in this PR?

I think we can hold off on that, and just try to reapply that branch's commits separately as a new PR. If that was the only concern, can you re-review?

PR #758, opened 2025-08-07
Allow user to select a reward for their AP exams

### LINKED ISSUE #689: User should be able to pick which reward they want from AP exams

ellisl10:
## Description
Saves the user's selected rewards and clears the corresponding course instead of always selecting the first one.

## Screenshots

Before
Screenshot 2025-08-07 at 9 11 27 AM

After
Screenshot 2025-08-07 at 9 12 10 AM

## Test Plan

- [ ] The default selected reward is the first one. Upon selecting a different reward, the planner should update to clear the new reward instead of the old one.
- [ ] Ensure that it persists between sessions

## Issues

Closes #689


site/src/pages/RoadmapPage/transfers/APExamsSection.tsx
```diff
@@ -88,28 +149,23 @@ const APCreditMenuTile: FC<{ exam: TransferWithUnread<TransferredAPExam> }> = ({
...
+        headerItems={selectBox}
+        units={units}
+        setUnits={updateUnits}
+        deleteFn={deleteFn}
+        unread={unread}
+      >
+        <Rewards examName={examName} coursesGranted={coursesGranted} />
+      </MenuTile>
+    );
   }
```
CadenLee2, line 168: I believe AP Art History issue is here, and noticed that also setting an AP exam's score to something below the requirement (like `1`) will also "hide" the exam.

If the score is not acceptable for any of the courses, the `continue` will run, the loop will end, we reach the end of the function, and nothing will be returned/rendered.

Instead, maybe we could determine how to render the `Rewards` inside this loop (i.e. which reward to do, or whether a reward is possible), but still return the parent `MenuTile` outside of it

CadenLee2 (changes requested):
Found two bugs, but the feature generally works well in the ideal cases and code is solid

One minor thing: I feel like the dropdown looks a bit unfinished since it's unstyled (see below)--would be a quick change, but we'd have to decide what it should look like. If needed we can also wait to do a future PR that improves it, but I'm wondering if you have any thoughts on how it could look.
image

**Major bugs**:
1. When I add a reward for an AP exam for the very first time, the selection on the frontend doesn't update.
    - E.g. I have AP Chemistry added but haven't selected a reward yet. When I go to staging and log in, then select ENGR 1A as my reward, it updates in the database but doesn't show that the dropdown updated on the frontend
    - When logged in, if I reload after selecting, then everything works fine for that AP exam, but the problem happens before reloading
    - When logged out, reward selection doesn't seem to work at all, even if I reload. Is it meant to work logged out? If not then probably disable the dropdown and explain the user has to log in via a tooltip/note somewhere
2. I tried adding AP Art History as an AP exam and it seems to have added in the database, but it fails to render in the list now (works on prod). Also, setting the score of an AP exam with a requirement to 1 will cause it to disappear from the list. Found the likely cause in a comment below

Also some small code details in comments

ellisl10:
> Found two bugs, but the feature generally works well in the ideal cases and code is solid
> 
> One minor thing: I feel like the dropdown looks a bit unfinished since it's unstyled (see below)--would be a quick change, but we'd have to decide what it should look like. If needed we can also wait to do a future PR that improves it, but I'm wondering if you have any thoughts on how it could look. image
> 
> **Major bugs**:
> 
> 1. When I add a reward for an AP exam for the very first time, the selection on the frontend doesn't update.
>    
>    * E.g. I have AP Chemistry added but haven't selected a reward yet. When I go to staging and log in, then select ENGR 1A as my reward, it updates in the database but doesn't show that the dropdown updated on the frontend
>    * When logged in, if I reload after selecting, then everything works fine for that AP exam, but the problem happens before reloading
>    * When logged out, reward selection doesn't seem to work at all, even if I reload. Is it meant to work logged out? If not then probably disable the dropdown and explain the user has to log in via a tooltip/note somewhere
> 2. I tried adding AP Art History as an AP exam and it seems to have added in the database, but it fails to render in the list now (works on prod). Also, setting the score of an AP exam with a requirement to 1 will cause it to disappear from the list. Found the likely cause in a comment below
> 
> Also some small code details in comments

site/src/app/roadmap/transfers/APExamsSection.tsx
```diff
@@ -88,28 +148,37 @@ const APCreditMenuTile: FC<{ exam: TransferWithUnread<TransferredAPExam> }> = ({
...
   const apiExamInfo = apExamInfo.find((exam) => exam.fullName === examName);
-  let message = '';
 
   for (const reward of apiExamInfo?.rewards ?? []) {
+    if (score === 1 || score === 2) {
+      return (
+        <MenuTile
+          title={examName}
+          headerItems={selectBox}
+          units={units}
```
CadenLee2, line 158: Since this loop includes some duplicate code (creating the `MenuTile` where only the inner rewards `coursesGranted` is different), I recommend refactoring out the logic for determining `coursesGranted` to flatten the early-return loop and make the code more clear.

You could move the loop into a helper function that takes in only the score and the exam info, then returns the best possible reward with that score or null if no reward matches. Calling it could look like:
```ts
const reward = getApplicableReward(score, apiExamInfo);
const coursesGranted = reward?.coursesGranted ?? '';

return (
  <MenuTile
    title={examName}
    // ...
  >
    <Rewards examName={examName} coursesGranted={coursesGranted} />
  </MenuTile>
);
```

CadenLee2 (commented):
Looking better, and the bugs appear to be fixed. We can also discuss the style of the dropdowns more in our meeting

site/src/app/roadmap/transfers/APExamsSection.tsx
```diff
@@ -147,38 +165,50 @@ const APCreditMenuTile: FC<{ exam: TransferWithUnread<TransferredAPExam> }> = ({
     trpc.transferCredits.deleteUserAPExam.mutate(examName);
   }, [dispatch, examName, isLoggedIn]);
 
-  const apiExamInfo = apExamInfo.find((exam) => exam.fullName === examName);
+  const coursesGranted = (getApplicableReward(score)?.coursesGranted as CoursesGrantedTree) ?? '';
```
CadenLee2, line 168: Because `getApplicableReward(score)?.coursesGranted` could be undefined, but `''` is known to be a valid `CoursesGrantedTree`, I'd move the `as` outside:
```ts
const coursesGranted = (getApplicableReward(score)?.coursesGranted ?? '') as CoursesGrantedTree;
```

site/src/app/roadmap/transfers/APExamsSection.tsx
```diff
@@ -147,38 +165,50 @@ const APCreditMenuTile: FC<{ exam: TransferWithUnread<TransferredAPExam> }> = ({
...
+      unread={unread}
+    >
+      <Rewards examName={examName} coursesGranted={coursesGranted} />
+    </MenuTile>
+  );
+
+  // for (const reward of apiExamInfo?.rewards ?? []) {
+  //   if (score === 1 || score === 2) {
+  //     return (
+  //       <MenuTile
```
CadenLee2, line 186: Outdated commented code, can be removed

CadenLee2 (changes requested):
Looks good, just some small last details

CadenLee2 (approved):
Let's get that money

Awesome-E (approved):

PR #760, opened 2025-08-17
chore: next in readme stack

laggycomputer:
y'all do not use vite anymore

it just works (it's a readme change) https://github.com/laggycomputer/peterportal-client/tree/a53bc6d4b7941c7c73631c94a9f2ab8288b6a2a3?tab=readme-ov-file#-built-with

## test plan

click the link and see for yourself

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Features include:
 - PostgreSQL
 - Drizzle ORM
 - TypeScript
-- Vite
+- Next.js
 
 ## First time setup
 


CadenLee2 (changes requested):
Yes good change

But unfortunately it does not just work

Check lines 102 and 104

laggycomputer:
this pr has been fact checked false by real peterportal patriots

will address in a bit

laggycomputer:
i `ripgrep`d for all references to vite i promise

PR #762, opened 2025-08-25
(WIP) Migrate all form inputs to MUI

fiveminus1:
## Description

(heavy WIP still) 

This PR migrates all form inputs (`Form.*`) to MUI components. This does *not* include buttons or the modal itself. 

## Test Plan

- [ ] Make sure there are no leftover files using Bootstrap inputs 
- [ ] Make sure all modals migrated still work properly
- [ ] Styling looks good 

## Issues

N/A


fiveminus1:
> I'd err on the side of caution when it comes to adding any new ppc-[...] classes, but you did a good job of denoting that it'd be replaced with MUI themes. 

More specifically, the `.ppc-checkbox` class was added because it's a bit tricky to customize the look of the checkboxes to match how they currently look without manually making a "custom icon" via a span and styling that. But yes, with the theme, we should be able to just define the icon in there and get rid of the class. 

> Once we get that and this merged, I think themes should come before migrating the rest of MUI. Thoughts on that?

Yes, I think that makes the most sense and the rest of the migrations' styling should be a lot smoother afterwards.

Awesome-E:
I find the border, especially when hovering to be a bit harsh, and it looks like the label font size and the padding have both shrunk. This seems to be the case across all modals.

image

It's kind of a nitpick and I'm not sure how much we should change, because the proper way would be to do it with a MUI theme. Curious if you have thoughts on opening the MUI Theme PR to have this as a base branch so we can merge theme and this PR together and avoid the awkward transition hiccup on prod

fiveminus1:
> One concern: is this orangey color meant to be our error color (--mui-palette-error-main)? It looks like this on light and dark theme.

I think my thought process with this was that we needed a semantic differentiation between "danger" and "error" but I do agree it looks off and it appears that the convention with MUI is to simply use "error" as "danger" anyways. I've updated error to just be the primary red.

> And speaking of the Zot4Plan error, it might make sense to have MUI errors there too? That's not too urgent though, and if you think that can wait for the future we can make a separate issue

This should be a part of another PR IMO, we currently have a mix of toasts and error messages throughout the multiplanner and it should be standardized to one or the other.

Awesome-E (approved):
money

Awesome-E:
the money was got

PR #765, opened 2025-08-28
Set cache policy id to avoid creating one every time

Awesome-E:
## Description

Uses the existing cache policy for SST deploys instead of trying to create a new one every time.

## Screenshots

N/A

## Test Plan

it deploys


CadenLee2 (approved):
It deploys ✅

PR #766, opened 2025-08-30
Use a pre-existing, non-managed, standalone cache policy

Awesome-E:
This is so it's not tied with an SST deployment and won't get removed on teardown like the previous one did.


CadenLee2 (approved):
Staging works

PR #770, opened 2025-09-16
Prevent installation of packages < 1 day old

Awesome-E:
## Description

As per an internal suggestion, this PR makes it so pnpm cannot install packages newer than 1 day. This is a precautionary measure to protect from the unlikely scenario that a package we use gets compromised and we try to install it before the malicious code gets taken down by npm.


laggycomputer:
let's get that m

CadenLee2 (approved):
Let's get that 💵

laggycomputer:
In order for this to take effect, the `pnpm` version specified in `package.json` [here](https://github.com/icssc/peterportal-client/blob/af077e0eb5a2a3e5bff3d582a53841c4d9db9d37/package.json#L19) must be bumped to 10.16 ([released](https://pnpm.io/blog/releases/10.16) 12 September) or higher.

Awesome-E:
should we revisit this and turn this pr into a pnpm version bump? @CadenLee2

PR #790, opened 2025-09-30
Fixing corequisite checking logic to use prereq node instead of coreq text

### LINKED ISSUE #723: Corequisite Checks do not always work correctly
Main Premise: we should be using the course's `prerequisiteTree.[AND|OR|NOT].coreq` property to validate corequisites instead of the `corequisite` string. The prerequisite tree also provides more information on whether alternatives can clear the corequisite.

Example Schedule:
Image

it says something is missing when in reality I do have (_Math 2B_ ~~OR AP Calc~~) and Physics 7C before it AND Math 2C + Physics 7LD in the same quarter. That means it should be fine. However, if I move Math 2D to fall, it no longer gives the warning, meaning that it was being treated like a prerequisite.

We should have `validateCoursePrerequisite` ignore the `corequisite` field and instead read information in the prequisite tree nodes to ensure the following behavior:
- Nodes _not_ marked as corequisites must be taken in previous quarters.
- Nodes that _are_ marked as corequisites are taken in previous OR the current quarter.

Bennables:
## Description

There were two issues with the corequisite checking for each course
1. we were using the prerequisite text, which was sometimes inconsistent with websoc. I fixed this issue by using the prerequisite nodes to check the coreqs instead.
2. the checking didn't check corequisites properly and treated them as prerequisites. Using the prerequisite nodes also solved this issue.

## Test Plan

Just make sure that it works properly by testing classes with many prereqs/coreqs.
Some examples are eecs170a, physics 7d, biosci 100

## Issues

Closes #723

### DIFF
--- a/site/src/app/roadmap/transfers/TransferCreditsMenu.scss
+++ b/site/src/app/roadmap/transfers/TransferCreditsMenu.scss
@@ -10,7 +10,7 @@
   &.mobile {
     @include globals.bottom-overlay(400);
   }
-
+  padding-bottom: 56px;
   transform: translateY(100%);
   transition: transform 0.3s;
   &.enter,
--- a/types/src/courseRequirements.ts
+++ b/types/src/courseRequirements.ts
@@ -1,7 +1,7 @@
 import { components, operations } from './generated/anteater-api-types';
 
 export type GEName =
-  operations['apExams']['responses'][200]['content']['application/json']['data'][number]['rewards'][number]['geCategories'][number];
+  operations['rawGrades']['responses'][200]['content']['application/json']['data'][number]['geCategories'][number];
 
 export type GETitle =
   operations['courseById']['responses'][200]['content']['application/json']['data']['geList'][number];


Awesome-E:
@CadenLee2 i helped guide a bit along the way so maybe better if you review first. let me know if you can't though, and I can cover.

site/src/helpers/planner.ts
```diff
@@ -310,19 +310,31 @@ interface ValidationInput<PreqrequisiteType> {
...
+      if (previouslyComplete || takingCorequisite) {
+        return new Set<string>();
+      }
+    }
+    return previouslyComplete ? new Set<string>() : new Set([id]);
+  }
 
+  //otherwise, it's an exam
+  const id = prerequisite.examName;
   const previouslyComplete = taken.has(id);
```
Awesome-E, line 336: Your code looks correct to me, but there is some duplicate logic here.

The following are repeated character for character, and some other places have more or less the same logic:
- `const previouslyComplete = taken.has(id)`
- `return previouslyComplete ? new Set<string>() : new Set([id])` 

Instead of having two nested ifs, how about creating a single intermediate variable for whether an "item is a course AND a corequisite AND the user is taking it this quarter"?

You could also exit early if something is previously completed or not a course if you don't want the conditional to be super long

That way, the logic in the end will read something like:
```ts
const id = ...
const previouslyComplete = ...
const isTakingCoreq = ...
return previouslyComplete || isTakingCoreq ? new Set<string>() : new Set([id])
```
or this, if you prefer guard clauses:
```ts
const id = ...
const previouslyComplete = ...

if (previouslyComplete) return new Set<string>();
if (<check is ap exam> || <check is not coreq>) return new Set([id]);

const isTaking = ...
return isTaking ? new Set<string>() : new Set([id])
```

site/src/helpers/planner.ts
```diff
@@ -273,9 +273,9 @@ export const validatePlanner = (transferNames: string[], currentPlanData: Planne
       quarter.courses.forEach((course, courseIndex) => {
         if (!course.prerequisiteTree) return;
 
-        const { prerequisiteTree: prerequisite, corequisites: corequisite } = course;
+        const { prerequisiteTree: prerequisite } = course;
```
Awesome-E, line 276: Since we are only destructuring one thing now, this might read better as:

```suggestion
        const prerequisite = course.prerequisiteTree;
```

PR #791, opened 2025-09-30
Make roadmap default tab

### LINKED ISSUE #763: Make roadmap the default tab
Currently, visiting <https://peterportal.org/> defaults to the course search page.

To reflect an increased focus on the roadmap/planner as the central feature of PeterPortal, the roadmap should be the default tab instead.

Also update the changelog modal so users familiar with the old defaults know that course and professor search are still available. Consider a screenshot that prominently includes the left sidebar menu and a message like "Peter's Roadmap is now the default page! You can still access course and professor search via the left sidebar."

After the change, we should immediately get to the roadmap page when going to PeterPortal, like in this screenshot:
Image

DerrickThrower:
Updated the default tab to the roadmap

used page.tsx to render the roadmap feature instead of the course search
updated the changelog information with a description of the switch and  screenshot of the sidebar

## Screenshots
Screenshot 2025-09-29 at 11 13 59 PM
Screenshot 2025-09-29 at 11 27 30 PM

## Test Plan

check the the initial tab is the roadmap
check the sidebar correctly toggles between tabs

## Issues

closes #763

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,31 @@
+name: Unit Tests
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  lint:
+    name: Run Jest Test Suite
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+      - name: Setup pnpm
+        uses: ./.github/actions/setup-pnpm
+
+      - name: Install Dependencies
+        run: pnpm install
+        env:
+          HUSKY: 0
+
+      - name: Run Tests
+        run: pnpm test
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,6 @@ site/build/
 .open-next
 .next
 tsconfig.tsbuildinfo
+
+# Jest
+site/coverage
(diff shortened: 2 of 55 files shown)


CadenLee2 (changes requested):
Looking good, but don't forget to update the screenshot `roadmap-default.png` to show that Roadmap is now at the top of the sidebar

DerrickThrower:
Screenshot 2025-10-02 at 4 04 33 PM

site/src/component/SideBar/SideBar.tsx
```diff
@@ -36,10 +36,19 @@ const SideBar = () => {
...
+            </div>
+            <span className="full-name">Peter's Roadmap</span>
+            <span>Roadmap</span>
+          </Link>
+        </li>
+        <li>
+          <Link href="/" className={pathname === '/search/courses' ? 'sidebar-active' : ''} onClick={closeSidebar}>
             <div>
               <ListAltRoundedIcon className="sidebar-icon" />
             </div>
```
CadenLee2, line 54: Since courses is no longer the default tab, this `/` navigates to the roadmap instead of courses when you click it--can just change the `href` to `/search/courses`

CadenLee2 (commented):
Tested out and it's mostly working! Would like to get @Awesome-E 's thoughts on the initial flash of light theme in the sidebar, which is more prominent now since roadmap is the first page: this screenshot is in the first few milliseconds and it's caused by some theming complications we were working on earlier. Not sure if it's major enough that we would want to wait on merging this until that gets fixed
Image
Also the courses link is broken, but that should be an easy fix

PR #792, opened 2025-10-01
Store which majors/minors have been expanded/collapsed

### LINKED ISSUE #657: Store which majors/minors have been collapsed when switching between tabs
Currently, PeterPortal will remember which requirement groups (i.e. "11 Upper-Div Electives") you've closed, but not the major itself.

i.e. these are currently remembered:

Image

But these are not:

Image

The requirement groups (first screenshot) already have a way of passing down/accumulating a "key" for each one to uniquely identify it. For example, in the screenshot below, "Emphasis in Cultural Comp Lit & Critical Theory" might have a key of `comLit-2-0` (major ID, third item at outer level, first item within that block).

Image

To complete this issue, you should explore how the keys are passed down. For each major, it starts with the major ID, so we can simply use that as the key for whether the major as a whole is collapsed. It's likely that you can use the same `expandedGroups` dictionary within the `courseRequirementsSlice.ts`, and just add logic to `MajorCourseList` and `MinorCourseList`.

It is probably worth consolidating the header button for both of these into a single helper component that takes in the Major/Minor name and ID, and updates the state based on that.

Note: This is labeled "good first issue" but may require a bit of exploring and learning how to read/write in Redux across different components

skylotua:
## Description
Instead of storing whether the MajorCourseList is collapsed locally, we're storing it in the courseRequirementsSlice.

## Screenshots
N/A

## Test Plan
1. Select Computer Science Major
6. Expand Computer Science Major such that you can see the specialization selection
1. Switch to Minor tab then back to Major tab
2. Major tab should stay expanded
3. Collapse Computer Science Major such that you cannot see the specialization selection
1. Switch to Minor tab then back to Major tab
2. Major tab should stay collapsed
0. Repeat for any Minor

## Issues

Closes #657

### DIFF
--- a/api/src/controllers/external/roadmap.ts
+++ b/api/src/controllers/external/roadmap.ts
@@ -1,7 +1,7 @@
 import { TRPCError } from '@trpc/server';
 import { publicProcedure, router } from '../../helpers/trpc';
 import { user } from '../../db/schema';
-import { eq } from 'drizzle-orm';
+import { eq, or } from 'drizzle-orm';
 import { z } from 'zod';
 import { queryGetPlanners } from '../../helpers/roadmap';
 
@@ -12,7 +12,11 @@ const externalRoadmapsRouter = router({
       throw new TRPCError({ code: 'UNAUTHORIZED' });
     }
 
-    const planners = await queryGetPlanners(eq(user.googleId, input.googleUserId));
+    const idLegacy = input.googleUserId;
+    const idPrefixed = `google_${idLegacy}`;
+
+    const where = or(eq(user.googleId, idPrefixed), eq(user.googleId, idLegacy))!;
+    const planners = await queryGetPlanners(where);
 
     return planners;
   }),
--- a/site/src/component/Review/Review.tsx
+++ b/site/src/component/Review/Review.tsx
@@ -6,12 +6,13 @@ import './Review.scss';
 import { selectReviews, setReviews, setFormStatus } from '../../store/slices/reviewSlice';
 import { useAppSelector, useAppDispatch } from '../../store/hooks';
 import { CourseGQLData, ProfessorGQLData } from '../../types/types';
-import { Button, MenuItem, Select } from '@mui/material';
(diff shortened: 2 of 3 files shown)


site/src/app/roadmap/sidebar/MajorCourseList.tsx
```diff
@@ -38,10 +39,15 @@ interface MajorCourseListProps {
 }
 
 const MajorCourseList: FC<MajorCourseListProps> = ({ majorWithSpec, onSpecializationChange, selectedSpecId }) => {
+  const storeKeyPrefix = `major-${majorWithSpec.major.id}`;
   const isDark = useContext(ThemeContext).darkMode;
   const [specsLoading, setSpecsLoading] = useState(false);
   const [resultsLoading, setResultsLoading] = useState(false);
-  const [open, setOpen] = useState(true);
+  const open = useAppSelector((state) => state.courseRequirements.expandedGroups[storeKeyPrefix] ?? false);
```
CadenLee2, line 46: The `false` causes majors to be closed by default the moment they're added, is this intentional? I think it would make more sense for them to be open by default (like it is currently)

Same thing with the minors

CadenLee2 (commented):
Tested and this works well! Main concern is that now majors/minors are closed by default (see comment below)

site/src/app/roadmap/sidebar/MajorCourseList.tsx
Awesome-E: that's a good point. The slice is currently implemented such that when you set expanded to false, it deletes from the `state.courseRequirements.expandedGroups` instead of setting it to false to free the key.

So, if this is to change, we'd actually want to either change that, or make it so that when we call a slice reducer function to add a major/minor, we also set its expandedGroups tag to true. The latter may also be the better approach if we want to be explicit about the starting state.

Awesome-E:
@skylotua see above comment

atile4 (approved):
Did the test plan, majors and minors save collapsed/uncollapsed state as expected, everything works. ✅💸

PR #796, opened 2025-10-07
added ability to add misc credits

### LINKED ISSUE #781: Allow users to add/edit miscellaneous transfers
Sometimes, students might have transfers from other schools that count only for elective credit. PeterPortal should allow them to add/edit "Other Transfers" within the add transfers menu instead of only restricting this to be on Transcript Imports.

(previously, this was not the case because there was no unique constraint on misc transfers)

This also means we should show "Other Transfers" regardless of whether the user has anything there.

charlieweinberger:
## Description

Currently, the "Other Transferred Credits" section acts as a temporary inbox for transferred items that weren't automatically converted to UCI courses or AP exams. This PR changes this section from being a temporary inbox to a permanent listing of uncategorized items, allowing users to add or edit items in this section that only give elective credit.

### Styling Questions

I wasn't sure how to style the input row, so lmk what yall think.

- Should the input row be styled with MUI, bootstrap, or a custom styling that looks like react-select?
- Should there be an "add" button, or should a course be added when the user unfocuses the "unit" input?
- Should the toggle transfers button be moved to below the uncategorized credits section, instead of being fixed to the bottom of the menu and therefore overlapping the uncategorized credits input row?

### Out-Of-Scope Questions

These are some questions related to the transfer credits menu that I thought of while working on this PR, that could be separate PRs.

- Should we allow users to collapse each sub-menu of the transfer credits menu? Currently, its annoying to scroll all the way past the GE section to get to the uncategorized credits section.
- Should we allow users to completely clear their transfer credits menu? Currently I don't think can do so without tediously resetting every value, which would seem annoying. For example, someone would need to do this if they uploaded the wrong transcript, or our auto-categorization didn't work for them, or they just want to manually input items.
- Should we make it more clear that the transfer credits menu applies to all roadmaps, not just the currently selected one?

## Screenshots

Before:
Screenshot 2025-10-07 at 12 39 05 PM

After (currently unstyled):
Screenshot 2025-10-07 at 12 30 39 PM

## Test Plan

- [ ] Verify that initial uncategorized transferred items are still liste
- [ ] Verify that items can be both added and edited, and that changes are saved both locally and to the db
- [ ] Give input on the UI/UX

## Issues

Closes #781

### DIFF
--- a/api/src/controllers/transferCredits.ts
+++ b/api/src/controllers/transferCredits.ts
@@ -159,6 +159,17 @@ const transferCreditsRouter = router({
       .where(eq(transferredMisc.userId, ctx.session.userId!));
     return courses;
   }),
+  addUncategorizedCourse: userProcedure.input(zodTransferredUncategorized).mutation(async ({ ctx, input }) => {
+    await db
+      .insert(transferredMisc)
+      .values({ courseName: input.name, units: input.units, userId: ctx.session.userId! });
+  }),
+  updateUncategorizedCourse: userProcedure.input(zodTransferredUncategorized).mutation(async ({ ctx, input }) => {
+    await db
+      .update(transferredMisc)
+      .set({ units: input.units })
+      .where(and(eq(transferredMisc.userId, ctx.session.userId!), eq(transferredMisc.courseName, input.name ?? '')));
+  }),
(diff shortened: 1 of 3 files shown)


CadenLee2 (approved):
💶

Awesome-E (approved):
let's get that ~~merge~~ money 💸

PR #804, opened 2025-10-10
Create a Material UI theme

### LINKED ISSUE #761: Create a MUI theme
Before continuing MUI component migrations, we should create a [Material UI theme](https://mui.com/material-ui/customization/theming/) to define colors for certain components. Previously migrated component styling may also have to be adjusted to use the new theme, such as Checkbox (whose migration will be done before the theme is implemented).

fiveminus1:
## Description

This PR (branching off of #762) establishes a [Material UI theme](https://mui.com/material-ui/customization/theming/) for future component styling re-usability. This includes:
- [Palette](https://mui.com/material-ui/customization/palette/) - all colors are now defined in `theme.ts`, and the existing CSS vars now point to the vars generated by MUI. In the future, we should consider removing all of our defined css vars and simply using the MUI ones, but they're kept for this PR.
- Previously defined colors deemed unnecessary have been either removed or refactored into other colors (gray blue -> overlay2, removed ring-road-white).
- Note: Some "secondary" colors (green, yellow, orange, red) that were used in places like `GradeDist` (and occasionally other places) have been made into their own primary/secondary tokens (orange -> orange-primary, orange-secondary). Since there is, for example, no "orange-primary," placeholders have been added for such colors (orange-primary is simply orange). Perhaps those should be replaced in the future. 
- [Spacing](https://mui.com/material-ui/customization/spacing/) - defines a base spacing of `4px` 
- Components - listed below

The following components' styling have been moved from `App.scss` to the theme component style overrides:
- `Checkbox`
- `FormLabel`
- `FormControl` (which replaces previous Bootstrap `form-group` styling)
- `TextField` (`MuiInputBase`)

I'm deferring `ppc-btn` and `ppc-mui-button` styling to be migrated when `Button` has been fully replaced, since these appear to be including styles for some remaining Bootstrap Buttons and different Button variants will likely have to be established (e.g. Roadmap header and Multiplan buttons should probably be different from standard form buttons). See the Variants section [here](https://mui.com/material-ui/customization/theme-components/) for future reference.

Additionally, global styling from `index.css` and `theme.scss` have been consolidated into:
- `App.scss`: app-wide styling and Bootstrap overrides for the time being
- `globals.scss`: CSS variables and mixins
- Fonts are now loaded by [Next.js](https://nextjs.org/docs/app/api-reference/components/font) (previously in `index.css`)

## Screenshots

Everything should look the same as the base branch aside from checkboxes, but spacing may be different/more consistent in some places.

Before|After
:-:|:-:
image|image

## Test Plan
- [ ] Colors look good in light and dark mode
- [ ] All of the components listed above still appear good
- [ ] No Bootstrap overrides that were moved into `App.scss` broke 
- [ ] No remaining global styling that should've been moved into the theme
- [ ] Spacing looks good 
(note: Delete Year modal margin namely will be fixed in the `mui-modal-inputs` branch after this is merged)

## Issues

Closes #761

### DIFF
--- a/site/src/component/ReviewForm/ReviewForm.scss
+++ b/site/src/component/ReviewForm/ReviewForm.scss
@@ -30,7 +30,7 @@
   }
 
   .rating-error .MuiRating-iconEmpty {
-    color: red;
+    color: var(--mui-palette-error-main);
   }
 
   .ppc-combobox {


site/src/App.scss
```diff
@@ -4,13 +4,36 @@ button {
...
+  height: 100%;
+}
+
+html > body {
+  margin: 0;
+  font-family: 'Open Sans', sans-serif;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+  color: var(--text);
+  background: var(--background);
```
CadenLee2: Is there a specific reason for these font smoothing options, or were they just for testing?

If they're needed we can keep them, but just checking that they're not remainders from testing

CadenLee2 (commented):
Other than the font thing, which we can look more into in the future, I think theming has been successful! Couldn't find many issues on staging and based on last week's meeting it seems the theming code is as good as it reasonably can be

One issue: Bootstrap checkboxes are light on dark theme
Image

site/src/App.scss
fiveminus1: They aren't remainders from testing, I just copied them over from `index.css`.

site/src/app/layout.tsx
```diff
@@ -28,7 +42,12 @@ export default async function RootLayout({ children }: { children: React.ReactNo
   const user = await serverTrpc.users.get.query().catch(() => null);
 
   return (
-    <html lang="en" data-theme={user?.theme} suppressHydrationWarning>
+    <html
+      lang="en"
+      data-theme={user?.theme}
+      className={`${roboto.variable} ${openSans.variable}`}
```
Awesome-E: I believe this should be just Roboto?
```suggestion
      className={roboto.variable}
```

site/src/style/theme-d.ts
```diff
@@ -0,0 +1,70 @@
...
+      secondary?: string;
+    };
+    orange?: {
+      primary?: string;
+      secondary?: string;
+    };
+    red?: {
+      primary?: string;
+      secondary?: string;
+    };
```
Awesome-E: I think part of the purpose of a theme or palette in general is that you don't have random colors unrelated to the application.

As such, I think these colors should be nested into an explicit property that self-explains its purpose, i.e. inside `chart` so that it's clear these colors exist for that purpose.

Some organization like this:
```ts
interface TwoToneColor {
  primary: string;
  secondary: string;
}

// for palette
    chart: {
      red: TwoToneColor;
      orange: TwoToneColor;
      yellow: TwoToneColor;
      green: TwoToneColor;
      pass: string;
      noPass: string;
    }
// for palette options
chart?: {
      red?: Partial<TwoToneColor>;
      orange?: Partial<TwoToneColor>;
      yellow?: Partial<TwoToneColor>;
      green?: Partial<TwoToneColor>;
      pass?: string;
      noPass?: string;
    };
```

site/src/style/theme.ts
Awesome-E: Changes in this file would correspond to changes in the previous file

Awesome-E (approved):
I think this is good enough for the scope of the PR; let's get that money

PR #807, opened 2025-10-16
Migrate all buttons and dropdowns to MUI

fiveminus1:
## Description

Migrates all Bootstrap Buttons and Dropdowns to MUI's `Button` and `Select` (excluding the requirements selector/search side bar). In the future, we may also consider switching `react-select` dropdowns to MUI's `Autocomplete` or `Select`, but none of those are covered in this PR.

## Screenshots

(not exhaustive)
Before|After
:-:|:-:
image|image
image|image
image|image
image|image
image|image
image|image

## Test Plan

- [ ] All buttons behave properly and look good 
- [ ] All dropdowns behave properly and look good
- [ ] All form submissions still work correctly
- [ ] Verify dark/light mode and mobile styling

## Issues

N/A

### DIFF
--- a/site/src/component/SideInfo/SideInfo.tsx
+++ b/site/src/component/SideInfo/SideInfo.tsx
@@ -157,8 +157,12 @@ const SideInfo: FC<SideInfoProps> = (props) => {
           </div>
         </div>
 
-        <h4>Recent Offerings</h4>
-        {props.terms && <RecentOfferingsTable terms={props.terms} size="wide" />}
+        {props.terms && props.terms.length > 0 && (
+          <>
+            <h4>Recent Offerings</h4>
+            <RecentOfferingsTable terms={props.terms} size="wide" />
+          </>
+        )}
 
         <div className="side-info-ratings">
           <h2>Average Rating</h2>


site/src/component/AppHeader/AppHeader.tsx
```diff
@@ -51,22 +47,22 @@ const AppHeader: FC = () => {
       </p>
       <div className="feedback">
         <Button
-          as="a"
+          color="inherit"
           href="https://github.com/icssc-projects/peterportal-client/issues/new"
```
Cadecraft: I know this isn't really in scope, but when testing out the button thought we might as well make this change since it's quick. This link should be <https://github.com/icssc/peterportal-client/issues/new/choose> so that the new issue templates pop up

site/src/component/GradeDist/GradeDist.tsx
```diff
@@ -168,58 +165,47 @@ const GradeDist: FC<GradeDistProps> = (props) => {
...
-          className="ppc-dropdown-btn"
-          title={selectedProfCourseName}
-          variant={buttonVariant}
-          onSelect={updateProfCourse}
-        >
+        <Select value={profCourseSelectedValue} onChange={(e) => updateProfCourse(e.target.value)}>
           {profCourseOptions?.map((q) => {
             return (
-              <Dropdown.Item key={q.value} eventKey={q.value}>
+              <MenuItem key={q.value} value={q.value}>
```
Cadecraft: When data can't be loaded, these are blank instead of saying "All instructors", etc. In this case the dropdown is useless anyway, but I think it looks more broken when it's blank
Image

Cadecraft (commented):
🔥
- Tested out buttons and dropdowns, mostly works well
- Code looks great, just added a few comments

site/src/component/ResultPageContent/ResultPageContent.scss
```diff
@@ -63,6 +63,11 @@
       }
     }
 
+    .MuiInputBase-root .MuiInputBase-input {
```
Awesome-E: same idea

site/src/component/SideInfo/SideInfo.scss
```diff
@@ -30,6 +30,10 @@
   > *:not(:last-child) {
     margin-bottom: 20px;
   }
+
+  .MuiInputBase-root .MuiInputBase-input {
+    background-color: var(--overlay2);
+  }
```
Awesome-E: maybe not for this PR since we are in a bit of a crunch to get it merged, but in the future we should define additional variants in the theme that will apply the background color. I've done this for #826 but we can have this as a @todo and/or a new issue

site/src/helpers/styling.ts
```diff
@@ -14,6 +15,18 @@ export const tooltipStyle: Theme = {
   },
 };
 
+export const VisuallyHiddenInput = styled('input')({
+  clip: 'rect(0 0 0 0)',
+  clipPath: 'inset(50%)',
```
Awesome-E: i'm ngl this feels like a pretty weird way to do it.

I usually find myself doing something more like `inset: 0; width/height: 100%; opacity: 0` instead of using clip to hide the visual content

fiveminus1: I sort of just copied that snippet from the [docs](https://github.com/mui/material-ui/blob/v7.3.4/docs/data/joy/components/button/InputFileUpload.tsx), but I briefly looked into it and apparently it's a common pattern used by a couple libraries for accessibility reasons.  

There's also a [style utility](https://mui.com/system/screen-readers/#foo) `visuallyHidden` in `@mui/utils` that just does this, but we haven't used `utils` yet. If you think that'd be cleaner to use, I can just add it in this PR

PR #808, opened 2025-10-18
Inline Recent Offerings

### LINKED ISSUE #803: Inline the recent offerings table in course popover
The course popover unnecessarily requires users to hover over the emojis to see the recent offerings table (see screenshot), making it difficult to view and causing UI inconsistencies.

Instead, the table component itself should replace the "**Previous offerings:** (emojis here)" text and appear alongside the other course info content, removing the need to hover. Ensure it stays the small variant of the table.

| From the sidebar | Visual bug caused partly by the double-hover (ofc we can blame Bootstrap though) |
| --- | --- |
| Image | Image |

MajdKatabi:
## Description
Replaced the "Previous offerings: (emojis here)" text and recent offerings table popover to appear alongside the other course info content, thus removing the need to hover.

## Screenshots
Before: 
image
After: 
image

## Test Plan

Tested new title and table on roadmap GE class, roadmap search, and SideInfo.

## Issues

Closes #803

### DIFF
--- a/api/src/controllers/transferCredits.ts
+++ b/api/src/controllers/transferCredits.ts
@@ -159,6 +159,17 @@ const transferCreditsRouter = router({
       .where(eq(transferredMisc.userId, ctx.session.userId!));
     return courses;
   }),
+  addUncategorizedCourse: userProcedure.input(zodTransferredUncategorized).mutation(async ({ ctx, input }) => {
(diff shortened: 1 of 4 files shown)


CadenLee2 (approved):
LGTM (let's get that money)!

PR #815, opened 2025-10-23
Fixed Sidebar search on default page

### LINKED ISSUE #814: Fix sidebar search results on default page
Search results are failing in the sidebar search on the default roadmap page.

## Steps to recreate
Go to https://peterportal.org and immediately go to "Search" on the right sidebar then search for any course (e.g. "math"). No results will appear; only saved courses will appear.

Go to https://peterportal.org/roadmap and perform the same steps. The search results should appear properly.

## Cause & solution
The roadmap page (`/roadmap`) was recently made the default page (accessible on `/`). Most route-related logic was updated properly, but in `component/SearchModule/SearchModule.tsx` there is conditional logic that only hides saved courses when we are on `/roadmap`. This should be fixed.
https://github.com/icssc/peterportal-client/blob/107e947ef631923cf758776094ab1105c814af51/site/src/component/SearchModule/SearchModule.tsx#L76

Additionally, check for other cases of hardcoding `/roadmap` where this might be an issue.

DerrickThrower:
## Description

Included the "/" path in conditional that only checked for "/roadmap". Searched for other uses of the '/roadmap' path in conditional logic where I did not find anything.
There maybe a way to avoid the hardcoded pathnames to future proof changes to paths. 

## Screenshots

Screenshot 2025-10-23 at 1 03 25 PM

## Test Plan

- Use the sidebar search on the default and the /roadmap paths
- verify that saved courses was not affected 

## Issues

Closes #814

### DIFF
--- a/site/src/component/SearchModule/SearchModule.tsx
+++ b/site/src/component/SearchModule/SearchModule.tsx
@@ -73,7 +73,7 @@ const SearchModule: FC<SearchModuleProps> = ({ index }) => {
 
   const searchImmediately = (query: string) => {
     if (pendingRequest) clearTimeout(pendingRequest);
-    if (location.pathname === '/roadmap') {
+    if (location.pathname === '/roadmap' || location.pathname === '/') {
       dispatch(setShowSavedCourses(!query));
     }
     if (query && query !== search.query) {


fiveminus1 (approved):
Last comment should just be a separate PR, so shippit 🔥

PR #818, opened 2025-10-24
Redirect "/roadmap" to default page

### LINKED ISSUE #817: Roadmap and default page should not have separate paths
The root page `peterportal.org` now points to the Roadmap page, but clicking on the Roadmap button on the Sidebar still goes to a separate `peterportal.org/roadmap` directory. 

There's no reason for the distinction, so we should just have `/roadmap` redirect to `/` and remove any links that directly go to `/roadmap` along with any references to `/roadmap` such as in `SearchModule`.

DerrickThrower:
## Description

When a user tries to navigate to "/roadmap" the user is redirected to the default path for simplicity(no reason for both to exist)

## Test Plan

1. ensure when searching for /roadmap that the link is now the default path
2. switch between sidebar links to see that the / path is being used
3. ensure that the sidebar search is still working as intended because that is the only place I've found with the conditional rendering based on /roadmap path

closes #817


site/src/app/roadmap/page.tsx
```diff
@@ -1,5 +1,8 @@
-import RoadmapPage from './RoadmapPage';
+import { redirect } from 'next/navigation';
 
+/*
+    Refer to next.config.mjs for the main method of redirection
+  */
```
CadenLee2: We should be able to get rid of this file entirely because redirection is now handled by NEXT.js

PR #825, opened 2025-10-30
Fixed Recent Offerings to only be rendered if there are any offerings

### LINKED ISSUE #823: Professor pages should not display "Recent Offerings"
**Describe the bug**
If you go to a professor page, there's an unnecessary "Recent Offerings" header on the left side overview.

**To Reproduce**
1. Go to any professor page

**Expected Behavior**
"Recent Offerings" should not be there.

**Actual Behavior**

Image

**Additional context**
Eventually, it would be nice to have a "Recent Courses" table. However, that's out of the scope of this issue and we should just remove the visual bug for now.

anonymous-himalayas:
## Description

Added checks for rendering Recent Offerings only if there are any term properties.

## Test Plan

Go to any professor and Recent Offerings should no longer show

## Issues

Closes #823

### DIFF
--- a/site/src/component/SearchModule/SearchModule.scss
+++ b/site/src/component/SearchModule/SearchModule.scss
@@ -1,6 +1,8 @@
 @use '../../globals.scss';
 
 .search-module {
+  z-index: 100;
+
   button {
     font: inherit;
     padding: 0;
@@ -26,6 +28,7 @@
     background-color: var(--overlay1);
     border-color: var(--overlay2);
   }
+
   .search-bar:focus {
     border-color: #80bdff;
   }


fiveminus1 (changes requested):
Hey Hamil, thanks for the contribution!

You shouldn't have to touch the `parseOfferings` function for this - that's only internally used by the `RecentOfferingsTable`. The issue is that `<h4>Recent Offerings</h4>` is displayed despite there being no `terms` in the props for a Professor page (more generally, if no terms are found for a given page, professor or course, then the header shouldn't be displayed).

site/src/component/SideInfo/SideInfo.tsx
```diff
@@ -159,8 +159,12 @@ const SideInfo: FC<SideInfoProps> = (props) => {
           </div>
         </div>
 
-        <h4>Recent Offerings</h4>
-        {props.terms && <RecentOfferingsTable terms={props.terms} size="wide" />}
+        {props.terms && props.terms.length > 0 && (
```
fiveminus1: Checking the length of `props.terms` is redundant since the `props.terms &&` condition already does the same

fiveminus1 (approved):
money moves

Awesome-E (approved):
💰🦐

PR #826, opened 2025-10-31
Sticky toolbar

### LINKED ISSUE #777: Redesign planner header menu
To support future UI changes, the header should be updated to match the new designs. It should:
- Stick to the top of the planner view so it is still visible when the user scrolls down
- Take up the full width without any margin
- Match the styles of the design below

The changed planner year appearance from the screenshot is out of scope

| |
| --- |
| Before: Image |
| After: Image |

CadenLee2:
Styled the roadmap toolbar to match the course preview toolbar

## Description

Changed toolbar positioning and styles. This mostly involves changing some padding and MUI button parameters.

Planner padding on desktop was also changed, because it looked somewhat unnatural to have wider padding, considering that the toolbar is no longer also subject to that padding. This is subjective though

**Silly styling things (might want to change, advice appreciated)**
- The border for the multiplan dropdown button now uses `var(--text-secondary)`. It looks good, but this may be cursed and we can change if needed
- To maintain the same height as the course preview toolbar, padding for `planner-left` is exactly `2.75px`. This is also cursed, but other solutions might involve hardcoding a height

**Other notes**:
- Now that the multiplan dropdown is restyled, the `--blue-accent` color is only used once (in `.ppc-modal a`). We may want to consider consolidating that with another blue at some point

## Screenshots

| Sticky toolbar | image |
| --- | --- |
| Dark theme & condensed | image |
| Mobile | image |
| Course header (for reference) | image |

## Test Plan

- [ ] Ensure functionalities (add year, roadmap selector, save, transfer credits, etc.) still work on desktop and mobile
- [ ] Ensure scrolling still works well, including in the course preview
- [ ] Test responsiveness on different screen widths

## Issues

Closes #777

### DIFF
--- a/site/src/style/theme.ts
+++ b/site/src/style/theme.ts
@@ -127,8 +127,6 @@ theme = createTheme(theme, {
           style: {
             height: '24px',
             fontSize: '11px',
-            // fontFamily: "Roboto, Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif",
-            // fontWeight: '600',
             '& .MuiButton-startIcon': {
               marginRight: 4,
             },


site/src/app/roadmap/planner/Planner.scss
```diff
@@ -10,18 +10,21 @@
   overflow: auto;
   height: 100%;
 
-  @include globals.outer-padding();
+  section.years {
+    padding-left: 24px;
+    padding-right: 24px;
```
Awesome-E: You can use `padding-inline` to represent left and right together

site/src/app/roadmap/planner/Planner.scss
```diff
@@ -10,18 +10,21 @@
...
+    @media (max-width: globals.$mobile-cutoff) {
+      padding-left: 16px;
+      padding-right: 16px;
+    }
+    padding-top: 0px;
+    padding-bottom: $planner-gap;
```
Awesome-E:
```suggestion
    padding-block: 0 $planner-gap;
```

site/src/app/roadmap/toolbar/Header.scss
```diff
@@ -84,7 +59,7 @@
         padding-inline: 8px 6px;
```
Awesome-E: These custom styles might be something to consider changing or removing since we're now using the full MUI button shape. The buttons for mobile in the screenshots look a bit off because they don't use the same padding ratios as other MUI buttons (i.e. Add Course)

site/src/app/roadmap/planner/AddYearPopup.tsx
```diff
@@ -53,7 +53,15 @@ const AddYearPopup: FC = () => {
         // When the year changes, this will force default values to reset
         key={'add-year-' + placeholderYear}
       />
-      <Button variant="text" className="header-btn" onClick={() => setShowModal(true)} startIcon={<AddIcon />}>
+      <Button
+        variant="contained"
+        color="inherit"
+        size="small"
+        disableElevation
```
fiveminus1: I'm setting `variant="contained"` and `disableElevation` to be the theme default in my Buttons PR, so you can remove that when it gets merged.

Issue #771, opened 2025-09-18
Redesign profile menu


CadenLee2:
The menu when clicking on your profile picture in the corner currently includes:
- Navigation (to Your Reviews only)
- A nested theme setting
- And a Log Out button

To make this menu more useful and extensible for the future, it should instead provide:
- Navigation to Your Reviews AND to the admin pages (if logged in as an admin)
  - The admin pages should be removed from the left navigation sidebar since they are now accessible here
- Flattened user settings (theme and potentially others in the future)
- And a log out button

Optionally, to clarify that a menu is there, the profile picture can include a dropdown icon (as shown in the design below)

| Before | After |
| --- | --- |
| <img width="328" height="314" alt="Image" src="https://github.com/user-attachments/assets/9579fa69-4af0-4ac2-b2c1-95a5c8bde249" /> | <img width="133" height="62" alt="Image" src="https://github.com/user-attachments/assets/29db715b-1602-4992-9238-08c86b13bad2" /> <img width="700" height="714" alt="Image" src="https://github.com/user-attachments/assets/a8b44607-a834-48f7-8ed3-9cbc26776b97" /> |

## Implementation details
- The new theme selector uses the same button selector bar component as the requirements list selector; this should have already been refactored into its own reusable component in #772
- All Bootstrap in the user profile menu and the button selector bar should be removed and replaced with MUI
- The changed blue color is out of scope

PR #830, opened 2025-11-04
Mui Transition For TabSelector & Profile Menu (#772 & #771)

### LINKED ISSUE #772: MUI for Search Sidebar
The search sidebar's requirements list selector currently uses Bootstrap buttons:

Image

1. These buttons should be replaced by MUI buttons
2. To make this more reusable, the selector should be refactored into a shared component. This might mean it takes in as props:
    - The list of tab names and optionally icons associated with each tab (see an intended use case in #771 )
    - The currently selected tab name
3. The new selector should look close to the original, and the requirements list selector should behave as it did before.

After this, there should be no more Bootstrap in the search sidebar aside from popovers.

MajdKatabi:
## Description

Removed bootstrap for both the Search Sidebar and Profile Menu by creating a new shared component, TabSelector. Additionally, redesigned profile menu to include theme selections in the default popover using the new TabSelector. 
### Search Sidebar
| Before | After |
| --- | --- |
| image | image | 
### Profile Menu
| Before | After |
| --- | --- |
| Image | image | 
## Test Plan

To see the new sidebar with the removed bootstrap open the roadmap tab and to see the new profile menu click on the top right profile picture.
## Issues

Closes #772 & #771


site/src/app/roadmap/sidebar/TabSelector.scss
Awesome-E: I am wondering, how much of this CSS can we eliminate? I think that in the future, we should instead look into defining the tab selectors in our theme to have these styles wherever possible (i.e. create an issue for the future)

Awesome-E (changes requested):

site/src/component/AppHeader/Profile.tsx
```diff
@@ -44,98 +39,90 @@ const Profile = () => {
...
+                  component={Link}
+                >
+                  <ListItemIcon>
+                    <FlagIcon />
+                  </ListItemIcon>
+                  <ListItemText primary="View Reports" />
+                </ListItemButton>
+              </ListItem>
+            </>
+          )}
```
Awesome-E: I'd recommend pulling this out into a helper component in the same file `function AdminProfileLinks();`

site/src/component/AppHeader/Profile.scss
```diff
@@ -76,83 +75,37 @@
...
-    background-color: var(--overlay2);
+  .MuiListItemIcon-root {
+    min-width: 32px;
   }
 }
 
-.profile-popover__link.active,
-.theme-popover__link.active {
-  color: var(--blue-primary);
-}
```
Awesome-E: Because this was removed, navigating to that page no longer highlights it in the profile menu. I imagine this behavior is not something we want to get rid of

Awesome-E (changes requested):
couple things + I have a few css-style nitpicks (but for those I can just add a commit to the branch later)

Awesome-E (approved):
I think i'll save polishing touches for after we merge MUI since it makes sense to do those at the same time for consistency. Let's get that money 💸

PR #831, opened 2025-11-04
Global redux cache for current week/quarter

### LINKED ISSUE #806: Create a global cache for static API data
There are places where PeterPortal reads data that shouldn't change within a normal user session. This is mainly the case for current week/quarter/term, but will also be the case for things like a list of all departments (this will be used with search filters).

Instead of fetching these within said components, we should either modify existing `featureSlice.ts` files or create 1 or more new ones to account for this.

The ideal behavior for a PR that closes this issue should do something like this:
1. Fetch the static data when PeterPortal first loads
2. Once fetched, store that data in a slice
3. Replace lookups in other places (i.e. the Recent Offerings table) with a simple `useAppSelector(state => state.something.currentWeek)` to avoid making requests each time.

We might also want to consider store caches for things like professor netid or course lookups since those are similar.

---

I think this should be done sooner or later so we can (ideally) have a way to do global lookups before search filters is added. That way, search filters can just do a lookup on the full list of departments instead of hard-coding or fetching on render.

charlieweinberger:
## Description

Currently, when a component needs the current week/quarter (i.e the course recent offerings table), the data is fetched from tRPC every time. This PR makes it so that the tRPC requests to get the current week and quarter are fetched once when peterportal is first loaded, and stored in a redux slice. This way, future uses of the week/quarter are accessed through the redux slice, not through duplicate tRPC calls.

### Changes

- Created a redux slice `scheduleSlice` to store the current week and quarter
- Creates a hook `useSetSchedule` that is called in `AppProvider.tsx` when the site is first loaded
- Replaced all instances of using tRPC to fetch the current week/quarter with `useAppSelector` (using the slice above).

## Screenshots

No visual changes.

## Test Plan

- [ ] Verify that the current week/quarter are properly stored in the redux slice
- [ ] Verify that `<AppHeader>`, `<RecentOfferingsTable>`, and `<Schedule>` properly access the current week/quarter from the redux slice

## Issues

Closes #806


site/src/hooks/schedule.ts
```diff
@@ -0,0 +1,21 @@
...
+    trpc.schedule.currentWeek.query().then((res) => {
+      const week = res.display.split(' • ')[0];
+      dispatch(setCurrentWeek(week));
+    });
+
+    // set current quarter
+    trpc.schedule.currentQuarter.query().then((res) => {
+      dispatch(setCurrentQuarter(res));
+    });
+  }, [dispatch]);
```
Awesome-E: If I recall correctly, this will actually run twice due to React Strict mode. If you want to prevent this, see the approach I take in `site/src/component/CoursePreview/CoursePreview.tsx`.

---
I also had the initial concern that this doesn't have any sort of guard for if the original value exists, though that doesn't really matter since this is called once at the root level.

Awesome-E (approved):
:bowtie::bowtie::bowtie::bowtie:
⁣ ∧____________________
⁣ | ⁣ ⁣Let's get that money! ⁣ ⁣|
⁣ ⎺⎺⎺⎺⎺⎺⎺⎺⎺⎺⎺⎺⎺⎺⎺⎺⎺⎺⎺⎺⎺

site/src/hooks/schedule.ts
Awesome-E: `useEffect` is double-called on first render because of react strict mode. The suggestion I had would be to use the "[set]loadTrigger" so that the double render only applies to `setLoadTrigger(true)`, and the other hooks will only get past the `if (!loadTrigger) return false` once.

> I initially had that, removed it b/c of the reasoning u mentioned, but can easily add it back.

No need; was just a thought

PR #833, opened 2025-11-05
Switch to MUI toasts

### LINKED ISSUE #811: Switch to MUI toasts
Our current toasts use `toastify-js`, but we can switch to using [MUI for toasts/"snackbars"](https://mui.com/material-ui/react-snackbar/). The change will likely mostly involve editing `helpers/toastify.ts` (and probably renaming this file to something more generic, since we're switching)

Colors can be updated with MUI theming in the future; this does not have to wait on that branch to be merged. For now we can use the same colors as the existing ones:
- `--blue-primary` for normal toasts
- `--red-primary` for errors

The toastify dependency can then be removed.

ellisl10:
## Description

Removed toastify-js and replaced all toasts with MUI Snackbars. 

## Screenshots

Before
Screenshot 2025-11-05 at 7 12 40 AM

After
Screenshot 2025-11-05 at 7 50 39 AM

## Test Plan

- Make sure the following spawns a MUI snackbar that has the correct message + color and that it appears above everything else on the page:
  - Save Roadmap
  - Trying to edit a roadmap and give it an empty name or duplicate name/year
  - Adding a year and giving it a duplicate name/year
  - Importing from Zot4Plan
  - Submitting a report
  - Voting on a review
  - Submitting a review

## Issues

Closes #811

### DIFF
--- a/site/next-env.d.ts
+++ b/site/next-env.d.ts
@@ -1,5 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
+/// <reference types="next/navigation-types/compat/navigation" />
 
 // NOTE: This file should not be edited
 // see https://nextjs.org/docs/app/api-reference/config/typescript for more information.


site/src/helpers/toast.tsx
```diff
@@ -0,0 +1,32 @@
...
+interface ToastProps {
+  text: string;
+  severity: string;
+  showToast: boolean;
+  onClose: () => void;
+}
+
+const Toast: FC<ToastProps> = ({ text, severity, showToast, onClose }) => {
+  const backgroundColor = severity === 'error' ? 'var(--red-primary)' : 'var(--blue-primary)';
+  return (
```
CadenLee2: Considering that there are only two visual variants of the toast (error or not), what do you think about either just passing a boolean `isError` or making the type of `severity` be `'error' | 'success'` (or maybe `'error' | 'success' | 'info'`)?

site/src/component/ReviewForm/ReviewForm.tsx
```diff
@@ -493,6 +510,7 @@ const ReviewForm: FC<ReviewFormProps> = ({
           </Button>
         </Box>
       </Modal.Body>
+      <Toast text={toastMsg} severity={toastSeverity} showToast={showToast} onClose={handleClose} />
     </Modal>
```
CadenLee2: When logged out, on course / course preview, clicking the "rate course" button seems to crash/freeze the website

CadenLee2 (changes requested):
I like the use of the roadmap slice since it prevents individual components from having to manage toasts themselves, and even though there's inconsistency with things outside the roadmap page, I think it's fine because we can clean those up once the unified view is finished.

Because of the left sidebar, the bug from the last review is no longer a problem, but I did find one more issue

PR #834, opened 2025-11-05
Clicking a course in a prereq tree should open it in a new tab

### LINKED ISSUE #822: Clicking a course in a prereq tree should open it in a new tab
Note: course previews in #776 are the main reason this is important, but this can be done before course previews are merged.

When the user clicks on a course in the prerequisite tree (e.g. from a course preview), we currently navigate directly to that course in the current tab. Instead, it should open up in a new tab so the user doesn't lose their place filling out their roadmap (especially if they're using the course preview).

Choollol:
## Description

Pass target="_blank" in the `Link` component in `PrereqTreeNode`.

## Screenshots

https://github.com/user-attachments/assets/2d7e0e44-23bc-4fb1-a333-01f022bf3a83

## Test Plan

1. Navigate to the full page preview for any course with prerequisites, e.g. ICS 45C.
2. Click on a course node in the prereq tree and confirm that it opens the link in a new tab.

## Issues

Closes #822


ethank2222 (approved):
Looks good, fully tested

PR #839, opened 2025-11-06
Consolidate course menus (credits, catalog, search) in the left sidebar on desktop

### LINKED ISSUE #837: Consolidate "credits" menus into the left sidebar
The "Transfer Credits" menu should be a part of the left sidebar to make it more clear that the menu to clear credits for UCI courses does in fact exist. People commonly struggle to find a way to clear course credits, and putting it alongside planner course selection makes the most sense.

In the future, we can also consider adding back the info text near prerequisite checking components to tell the user about this menu as well.

No

This
Image

This is a departure from our old approach to having "Credits" as a button the user can click to open this menu. It also consolidates search into one place so all course discovery & credits managements happen on the left sidebar

**Additional context**
The menu would be relatively unchanged on mobile (Credits is still a top-level menu instead of via "Add Course", and "add course" will still show maj/minor/ge/search, where search shows saved courses as it does currently). We may, however, consider the text "Add credits" instead for consistency.
Image

---

Note that these designs have certain changes that are out of scope of the issue (adding undo/redo + moving the save button). These are designs for future iterations, but this issue is only concerned with the 3 tabs and change of the "(Transfer) Credits" buttons

Awesome-E:
Remaining things to do:
- [ ] Ensure catalog switcher buttons (maj/min/ge/saved) work after updating from #830 
- [x] Catalog > saved tab
- [x] Semi-reimplementation of (Transferred) Credits unread indicator
- [x] update Pr description to describe more structural changes (e.g. mobile popup wrappers)

## Description

Adds a tab switcher to the left sidebar to contain "Credits", "catalog", and "search". The search section in the catalog is now replaced by "Saved", which is a list of bookmarked courses. Visuals there are not completely finalized.

On mobile, it should behave more or less the same as before, except that the text for transfer credits has been changed to "Add Credits" to be more consistent with Desktop and to reduce any confusion of whether "Credits" applies to only transfer students or also to non-transfers adding AP scores.

A couple of structural changes to take note of:
- New `<MobilePopup/>` component to take care of the logic with `<CSSTransition/>` (mostly the `refs` it creates).
- Extracting the "content" of the credits, catalog, and search menus into their own component so we can use those in both of their respective desktop and mobile containers
- Reorganization of files from `src/app/roadmap/sidebar` to `src/app/roadmap/catalog` since the sidebar is now a container component for multiple tabs instead of just the catalog.
- Loading more user data at the page-level since the content of the tabs are loaded in multiple places now.
- Renamed some Redux reducer functions to better reflect that they only control the mobile UI.

## Screenshots

image

image

## Test Plan

- [ ] Make sure the three tabs on desktop work
- [ ] Make sure that mobile adding course and transfers still work
- [ ] Test the behavior of selecting "saved" and "search", then switching between mobile and desktop.
- [ ] Test importing transcripts and making sure the unread dot behaves as expected
  - Note: delete misc transfers before doing this; duplicate misc. credits will crash right now, but that's out of scope of this PR
  - the unread dot should clear only when you enter, then exit either the tab or the mobile menu.
- [ ] Other pages, such as the course page (while it's still here) don't look broken.
- [ ] Toolbars don't break

## Issues

Closes #837

### DIFF
--- a/site/src/app/roadmap/search/SavedAndSearch.scss
+++ b/site/src/app/roadmap/search/SavedAndSearch.scss
@@ -8,9 +8,11 @@ h3.saved-courses-title {
 .roadmap-search-results {
   $header-font-size: 14px;
 
-  display: flex;
-  flex-direction: column;
-  height: 100%;
+  display: grid;
+  grid-auto-flow: row;
+  grid-auto-rows: fit-content;
+  justify-items: stretch;
+  height: fit-content;
   overflow: auto;
   gap: 8px;
   margin-top: 8px;
@@ -51,13 +53,6 @@ h3.saved-courses-title {
   .course :is(.course-and-info) {
     display: contents;
   }
-
-  display: grid;
-  grid-auto-flow: row;
-  grid-auto-rows: fit-content;
-  justify-items: stretch;
-  gap: 8px;
-  height: fit-content;
 
   &:not(.disabled) .course {
     cursor: grab;


site/src/app/roadmap/sidebar/DesktopRoadmapSidebar.tsx
```diff
@@ -0,0 +1,61 @@
...
+    <div className="roadmap-sidebar">
+      <Tabs
+        className="sidebar-tabs"
+        value={selectedIndex}
+        onChange={(_, newValue) => setSelectedIndex(newValue)}
+        variant="fullWidth"
+      >
+        <Tab
+          icon={<CompareArrowsIcon />}
+          iconPosition="start"
```
CadenLee2: The icon is inconsistent (the other one is `SwapHorizOutlinedIcon`, which I prefer)

site/src/app/roadmap/transfers/MobileCreditsMenu.tsx
```diff
@@ -0,0 +1,30 @@
...
+export const MobileCreditsMenu: FC = () => {
+  const isMobile = useIsMobile();
+  const show = useAppSelector((state) => state.transferCredits.showTransfersMenu);
+
+  const dispatch = useAppDispatch();
+
+  /** @todo move out of global state since this will no longer be conditionally rendered */
+  useEffect(() => {
+    if (!isMobile) dispatch(setShowTransfersMenu(false));
+  }, [dispatch, isMobile]);
```
CadenLee2: Is this comment meant to be addressed later? I feel like it makes sense to keep in global state?

site/src/component/CoursePreview/CoursePreview.scss
```diff
@@ -24,13 +24,17 @@
     top: 0;
     left: 0;
     width: 100%;
-    min-height: 66px; // fix to the natural height
+    min-height: globals.$roadmap-toolbar-height; // fix to the natural height
     border-width: 0 0 2px 0;
```
CadenLee2: Comment should no longer be necessary since you can just look at where else `roadmap-toolbar-height` is used

site/src/app/roadmap/sidebar/DesktopRoadmapSidebar.tsx
Awesome-E: changed!

site/src/app/roadmap/search/SavedAndSearch.scss
```diff
@@ -0,0 +1,73 @@
...
+  .course :is(.course-and-info) {
+    display: contents;
+  }
+
+  display: grid;
+  grid-auto-flow: row;
+  grid-auto-rows: fit-content;
+  justify-items: stretch;
+  gap: 8px;
+  height: fit-content;
```
CadenLee2: Unintentionally defined styling twice? (see at the top of `.roadmap-search-results`: `display`, `height`, `gap` are defined there too)

Awesome-E: surprised i didn't get a build-time warning from that

PR #840, opened 2025-11-06
GE Requirements are rendered incorrectly

### LINKED ISSUE #805: GE Requirements are rendered incorrectly
**Describe the bug**
Due to Anteater API's data for GEs being changed, there is data in a slightly different structure than we had previously assumed. This means that for GEs, we should handle top-level requirements[^1] differently.

**To Reproduce**
1. Go to roadmap page
2. Select the "GE" tab from the sidebar

**Expected Behavior**
Before expanding anything, the list should be "I. Lower Division Writing", "I. Upper Division Writing", "II. Science and Technology", etc.

**Actual Behavior**
We see a mix of the expected items and items that should be within a parent container, i.e. we should show "I. Lower Division Writing" as a requirement that contains "Writing 50" and "Writing 60" within it rather than having both writing 50/60 shown at the top-level.
Image

Additionally, we can see that Categories ii and viii do not have the correct title.

**Additional context**

Currently, we "flatten" the requirements received by API because some of the data received was in one "Group" requirement with 1 subgroup, where you had to complete that 1 subgroup. So our code, under that previous assumption, would do something like this:
```txt
category ii requirements           ||  --> II. Science and Technology
+-- II. Science and Technology     ||      +-- 1 semester transfer course
| +-- 1 semester transfer course   ||      +-- 3 courses category ii
| +-- 3 courses category ii        ||      
```

However, removing that outer layer (and repeating that process) is now causing categories 1A (lower div writing), II, and VIII to render incorrectly. We should most likely change the logic to flatten groups in 2 ways:
1. Preserve the **outer**-most name, since category ii and viii use it as their full GE description (i.e. "II. Science and Technology")
2. At the top-level for GEs only, do *not* flatten groups that contain more than 1 requirement.
    - We currently flatten "singleton groups"[^2] everywhere to avoid nesting too deep
    - However, we should NOT do this for GE top-level since we need Lower-Division writing's "complete both Writing 50 and 60" to be nested within "I. Lower-Division Writing" rather than appearing at the top level[^3].

Note that the desired behavior is not to completely skip flattening anywhere because we still need `[LdWr: [50&60: Wr50 AND Wr60]]` to be flattened into `[LdWr: Wr50 AND Wr60]`.

[^1]: Top-level requirements are defined as the blocks of requirements that you see before you expand anything.
[^2]: These are groups that require every subgroup to be complete (such as `[2: 2a AND 2b]`), meaning that instead of nesting one level (`[1], [2: 2a AND 2b], [3]`), we can pull out the child requirements to show them in the same list as the parent (`[1], [2a], [2b], [3]`)
[^3]: We don't want the behavior as shown in the last footnote for the top-level GE list since we want to preserve the supplied groupings. i.e. We want to keep `[LdWr: Wr50 AND Wr60] [2] [3]` instead of changing it to `[Wr50] [Wr 60] [2] [3]`

Choollol:
## Description

Preserve the labels of top-level requirement groups, and skip flattening when a group has more than one requirement.

### Before:
image

image

### After:
image
image

## Test Plan

1. Navigate to the GE requirements tab and confirm that each category has the correct label and levels of nesting
2. Confirm that the structure of categories I, II, and VIII are correct

## Issues

Closes #805


site/src/helpers/courseRequirements.ts
```diff
@@ -129,12 +129,38 @@ export function collapseSingletonRequirements(requirements: ProgramRequirement[]
...
-export function flattenSingletonGroups(requirements: ProgramRequirement[]): ProgramRequirement[] {
+/**
+ * Flattens requirements to prevent deep nesting.
+ * Preserves the titles of the top-level categories by default.
+ *
+ * ```typescript
+ *
+ *   requirements = flattenSingletonGroups(requirements);
+ *
+ * ```
```
Awesome-E: I don't see a reason to include this comment; usage from the function signature should be self-explanatory
```suggestion
```

site/src/helpers/courseRequirements.ts
```diff
@@ -129,12 +129,38 @@ export function collapseSingletonRequirements(requirements: ProgramRequirement[]
...
+    return flattenSingletonGroups(r.requirements, false);
   });
+  if (doPreserveTopLevelLabels) {
+    for (let i = 0; i < res.length; i++) {
+      res[i] = {
+        ...res[i],
+        label: requirements[i].label,
+      };
+    }
+  }
```
Awesome-E: This can be more simply expressed with a `.forEach()` since you're modifying a reference to an object.

```suggestion
  if (preserveOuterLabels) {
    res.forEach((r, i) => {
      r.label = requirements[i].label
    });
  }
```

PR #843, opened 2025-11-08
Fix package.json

anthonyj33:
## Description
- Fixes `lint` and `format` scripts by preventing error on unmatched pattern (i.e. no html files found)
- Adds `packageManager` field, which is needed when using corepack and has no effect if not using corepack.

## Screenshots

N/A

## Test Plan

N/A

## Issues

N/A


CadenLee2 (approved):
Looks good, running `pnpm format` and `pnpm lint` still work for me
💵

PR #844, opened 2025-11-08
Bug Fix: Prevent error on invalid course loading

### LINKED ISSUE #794: Upon loading a roadmap, any nonexistent course ID should be omitted
**Describe the bug**
Currently, loading a roadmap with nonexistent course IDs will crash PeterPortal.

**To Reproduce**
Steps to reproduce the behavior: Same as #793, as it tries to load the course ID "Loading..."

You can also recreate this easily with a local roadmap:
1. Save a normal roadmap while signed out.
2. run `localStorage.roadmap` in the console, and copy the string
3. change any course ID to something that doesn't exist
4. set `localStorage.roadmap = 'json_data_inside_quotes_so_it_saves_as_a_string'` to save the roadmap with a nonexistent course id
5. reload the page, which will try to read that roadmap

**Expected Behavior**
PeterPortal loads all courses except those with broken IDs, and the broken ones are ignored.
Nice-to-have: a toast message can let the user know which course IDs no longer exist.

**Actual Behavior**
PeterPortal crashes and shows `Application Error`.

**Additional context**
The fix for this will likely be in the functionality to "expand a planner". Roadmaps stored to the user account only store the course IDs for each quarter, but when we load that data in, we add the full course data (i.e. name, number of units, etc.) so it can be shown to the user.

In doing so, a Course ID that is in a saved roadmap will still try to look up the full course information. If such a course doesn't exist, we should not be including it in the "expanded" planner. In other words, it's fine to still _try_ to look up the course, but if that course can't be looked up, it should not be a part of the final "expanded planner".

ethank2222:
## Description

Iteratively look at the courses being pulled in, if invalid, discard, count invalid courses to be mentioned in error toast.

## Screenshots
Before: Error
After:
image

## Test Plan

1. Save a normal roadmap while signed out.
2. run localStorage.roadmap in the console, and copy the string
3. change any course ID to something that doesn't exist
4. set localStorage.roadmap = 'json_data_inside_quotes_so_it_saves_as_a_string' to save the roadmap with a nonexistent course id
5. reload the page, which will try to read that roadmap

Closes #794


Awesome-E (changes requested):
A couple suggestions

PR #845, opened 2025-11-09
Migrate all Bootstrap Modals to MUI Dialogs

fiveminus1:
## Description
Migrates all Bootstrap Modals to use the MUI [Dialog](https://mui.com/material-ui/react-dialog/) component.

## Screenshots

(not exhaustive)
Before|After
:-:|:-:
image|image
image|image
image|image
image|image
image|image
image|image
image|image|image

## Test Plan

- [ ] Confirm all modals work
- [ ] Confirm modals look fine in light and dark mode
- [ ] Test responsiveness of all modals
- [ ] Ensure no dropdowns or inputs elsewhere have been unintentionally affected by changes made to the theme
- [ ] Confirm there isn't any dead styling that still refers to Bootstrap Modals

## Issues

N/A

### DIFF
--- a/site/src/app/roadmap/transfers/APExamsSection.tsx
+++ b/site/src/app/roadmap/transfers/APExamsSection.tsx
@@ -76,10 +76,12 @@ const Rewards: FC<{ examName: string; coursesGranted: CoursesGrantedTree }> = ({
   const selectedApRewards = useAppSelector((state) => state.transferCredits.selectedApRewards);
   const selectedReward = selectedApRewards.find((reward) => reward.examName === examName);
   const dispatch = useAppDispatch();
+  const isLoggedIn = useIsLoggedIn();
 
   const handleSelect = (selectedIndex: number) => {
     // path column to be removed in a future PR
     dispatch(updateSelectedApReward({ examName, path: '', selectedIndex }));
+    if (!isLoggedIn) return;
     trpc.transferCredits.setSelectedAPReward.mutate({ examName, path: '', selectedIndex });
   };
 
@@ -192,10 +194,11 @@ const APExamsSection: FC = () => {
 
   // Set selected rewards
   useEffect(() => {
+    if (!isLoggedIn) return;
     trpc.transferCredits.getSelectedAPRewards.query().then((rewards) => {
       dispatch(setSelectedApRewards(rewards));
     });
-  }, [dispatch]);
+  }, [dispatch, isLoggedIn]);
 
   // Save AP Exam to store
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


site/src/App.scss
```diff
@@ -182,63 +183,6 @@ div.ppc-popover.bs-popover-bottom {
   }
 }
 
-.ppc-modal {
```
CadenLee2: Truly beautiful diff

fiveminus1: .aap-dialog :)

fiveminus1:
> Do we want a close button on the "what's new" modal?

Added! 
> Since all the other inputs use the accent blue in dark theme, I think the upload one should as well

Discussed on Discord with Caden but for future readers, since accent blue is more of a "focus" marker and input outlines, checkboxes, rating stars all still use the primary, I'll opt to keep the primary here for now. We may want to consider changing all of those elements to use the accent in the future.

> The search bar size got messed up as a result of changes (I expect that we do still want it to have the accent color in dark theme, but that otherwise it should look the same as the original):

Fixed. I also noticed the uncategorized transfer input was slightly off as a result of the default input padding change, so I adjusted that as well. I believe most TextField inputs should not have the `8px 10px` padding that the search bar does, so I've opted not to include that in the theme for now? Unsure if we do want to make that the default

site/src/component/ChangelogModal/ChangelogModal.scss
```diff
@@ -1,21 +1,10 @@
 @use '../../globals';
 
 .changelog-modal {
-  .modal-dialog {
+  .MuiDialog-paper {
     max-width: 500px;
   }
```
Awesome-E: i think you're able to give the Paper a className via `slotProps.paper.className`, and that can be used for the outer css selector here (`.changelog-modal`)

Awesome-E (commented):
Markup/jsx looks good, but a few questions regarding how we want to define style overrides. Perhaps something to think about in a future PR?

site/src/app/roadmap/planner/AddCoursePopup.scss
```diff
@@ -13,50 +13,44 @@
...
-    }
-  }
-
-  .modal-header {
+  .MuiDialogTitle-root {
+    display: flex;
+    align-items: center;
     gap: 8px;
-    padding-bottom: 0;
+    padding-bottom: 16px;
```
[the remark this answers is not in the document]
fiveminus1: see above

site/src/app/roadmap/planner/AddCoursePopup.scss
```diff
@@ -13,50 +13,44 @@
...
+      font-weight: normal;
+    }
+
+    .close-button {
+      margin: -2px 4px -4px 0;
+      &.MuiButtonBase-root {
+        width: 32px;
+        height: 32px;
+      }
+    }
```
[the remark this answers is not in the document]
fiveminus1: image

It's this popup, only on mobile AFAIK.

PR #847, opened 2025-11-10
Chart-styling

### LINKED ISSUE #819: Switching themes does not correctly re-render grade distribution chart colors
**Describe the bug**
If you switch themes while on a course/instructor page, the appropriate light/dark colors do not apply accordingly.

**To Reproduce**
Steps to reproduce the behavior (starting on light mode, but vice versa if you start at dark mode): 
1. Go to a course/instructor page, notice appropriate light mode colors
2. Switch to dark mode, colors are still light mode
3. Switch to light mode, now uses dark mode colors
4. Refresh the page, now uses correct colors (for either mode)

**Expected Behavior**
Light|Dark
:-:|:-:
Image|Image

**Actual Behavior**
Light|Dark
:-:|:-:
Image|Image

**Additional context**
Should investigate how the light/dark theme is provided and re-rendered in the Chart component.

NOTE: Wait for #762 to get merged before working on this.

atile4:
Chart Styling FIx

## Description

Set the overall chart theme to look like light theme regardless of light/dark mode.

Made new color variables for the chart to always set them as light theme colors.

## Screenshots

image

image

## Test Plan

- clear a course requirement on roadmap and verify the green color is correct on both themes
- go to course info and verify the chart theme on both modes is the same

## Issues

Closes #819

### DIFF
--- a/site/src/helpers/searchFilters.ts
+++ b/site/src/helpers/searchFilters.ts
@@ -1,5 +1,3 @@
-import { GE_TITLE_MAP } from './courseRequirements';
-
 export interface FilterOptions {
   levels: string[];
   geCategories: string[];
@@ -19,16 +17,10 @@ export const levels: Record<string, string> = {
 };
 
 export function stringifySearchFilters(filters: FilterOptions): StringifiedFilterOptions {
-  const removePrefix = (label: string) => label.replace(/^.*: /, '');
-  const filterCategories = Object.entries(GE_TITLE_MAP)
-    .filter(([, label]) => filters.geCategories.includes(removePrefix(label)))
-    .map((e) => e[0])
-    .join(',');
-
   return {
     // change empty string to undefined for API call purposes
     stringifiedLevels: filters.levels.join(',') || undefined,
-    stringifiedGeCategories: filterCategories || undefined,
+    stringifiedGeCategories: filters.geCategories.join(',') || undefined,
     stringifiedDepartments: filters.departments.join(',') || undefined,
   };
 }


CadenLee2:
Update: colors are now organized into `misc` (which includes success, error, etc.); `reviews` (a color scale that's light in light mode and dark in dark mode); and `chart` (a color scale that's always light). The theme is structured such that no color hex value is duplicated (even if some colors are reused).

Codewise, this means some refactors in the theme files. Visually, the only difference should be that review colors all now have dark theme variants:

image

site/src/style/theme.ts
```diff
@@ -52,16 +47,16 @@ const lightPalette: PaletteOptions = {
...
-      primary: '#ce0000',
-      ...sharedTokens.chart.red,
+  },
+  misc: {
+    ...sharedTokens.misc,
+    success: {
+      ...sharedTokens.misc.success,
+      secondary: sharedTokens.chart.green,
     },
+    warning: sharedTokens.chart.orange,
```
fiveminus1: Haven't looked too deeply into where warning is used by MUI components, nor can I recall any place where I've seen it, but I'm not sure I like it primarily relying on a chart color. Not horribly strong on this opinion, but if we can decide on another color that'd be preferred.

site/src/style/theme.ts
```diff
@@ -89,24 +84,29 @@ const darkPalette: PaletteOptions = {
...
 };
 
 const makeExtendedPalette = (base: PaletteOptions) => ({
   ...base,
-  success: { main: base.chart!.green!.primary! },
-  error: { main: base.chart!.red!.primary! },
-  warning: { main: base.chart!.orange!.secondary! },
+  success: { main: base.misc!.success!.primary! },
+  error: { main: base.misc!.error! },
+  warning: { main: base.misc!.warning! },
```
fiveminus1: Again, we can probably get rid of the token being in misc and instead just define it to use the chart color here? Would also remove the redundancy of defining it in light and dark, since `chart.orange` is a shared token.

site/src/style/theme.ts
```diff
@@ -89,24 +83,27 @@ const darkPalette: PaletteOptions = {
...
+    ...sharedTokens.misc,
+    success: {
+      ...sharedTokens.misc.success,
       secondary: '#295629',
     },
-    red: {
-      primary: '#ff3333',
-      ...sharedTokens.chart.red,
-    },
+    error: '#ff3333',
```
fiveminus1: This can be moved up a level, out of `misc`

site/src/style/theme-d.ts
```diff
@@ -17,15 +27,11 @@ declare module '@mui/material/styles' {
     };
     misc: {
       midGray: string;
+      success: TwoToneColor;
+      error: string;
```
Awesome-E: To my knowledge, success and error in most themes are defined at the top level, and I think that should be the case here as well (as opposed to putting it inside `misc`).

Awesome-E:
> ThankYouMessage

This is a component that used to show when you complete a review but has since been deleted. We should remove references to it.

> I feel as though it's unnecessary semantically to have a "second success,"

I agree with this; we should use the primary color in places such as Program Requirements for now. In the future, I think we will be restyling the completion styles for program requirements anyways.

site/src/globals.scss
```diff
@@ -11,24 +11,24 @@
...
-  --green-primary: var(--mui-palette-chart-green-primary);
-  --red-primary: var(--mui-palette-chart-red-primary);
+  --green-primary: var(--mui-palette-misc-success-primary);
+  --red-primary: var(--mui-palette-misc-error);
 
-  --green-secondary: var(--mui-palette-chart-green-secondary);
-  --yellow-secondary: var(--mui-palette-chart-yellow-secondary);
-  --orange-secondary: var(--mui-palette-chart-orange-secondary);
-  --red-secondary: var(--mui-palette-chart-red-secondary);
+  --green-secondary: var(--mui-palette-misc-success-secondary);
```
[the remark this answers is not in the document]
CadenLee2: As `--green-secondary` was only relevantly used in program requirements, it has been removed and the requirements usage has been updated to use `oklab` to set opacity. This lets us clean up a lot

site/src/style/theme.ts
```diff
@@ -89,31 +78,23 @@ const darkPalette: PaletteOptions = {
...
+    red: '#b7523e',
+    yellow: '#c49e3e',
+    orange: '#c47e38',
+  },
+  error: {
+    main: '#ff3333',
   },
 };
 
-const makeExtendedPalette = (base: PaletteOptions) => ({
```
fiveminus1: nice

fiveminus1 (approved):
Looks great; cleaned up the theme a ton and everything else looks money to me 😃 

I think a good next step after this is merged could be cleaning up how we define CSS tokens with MUI's tokens in `globals.scss` so we don't have to keep redundantly redefining variables and just make `theme.ts` the single source of truth.

PR #849, opened 2025-11-11
Fullscreen mobile search

### LINKED ISSUE #835: Fullscreen Mobile Search
For mobile users to search for courses (and eventually professors) independent of adding a course to their roadmap, the new unified roadmap page will require a dedicated search menu on mobile.

## Requirements
- The search icon in the top bar (see designs below) only appears on mobile. Tapping it opens the fullscreen search menu
- Results appear similarly to how they currently show in the course page (but should match our new design below)
- Tapping on a result opens the page for that course; tapping the bookmark icon should work to save the result course

## Implementation
Make sure to reuse existing components when possible (especially for the contents of the course search result cards). Pay attention to some details of the design, like long descriptions being cut short.

## Designs
Note that the current top bar does not closely match these designs, so deciding where to temporarily put the search button may take some discussion.

Also note that adding the search filters icon in the top right can wait until search filters get fully integrated.
| Universal page | Fullscreen mobile search | Current top bar |
| --- | --- | --- |
| Image | Image | Image |

anthonyj33:
## Description

- Search icon in navbar on mobile roadmap page
  - Opens fullscreen search page
- Clicking on a search result opens the course preview

## Screenshots

[Peter-s-Roadmap-PeterPortal.webm](https://github.com/user-attachments/assets/776222c4-79f9-4387-8fd9-f1c2d36df64e)

## Test Plan

- Test on mobile (or emulated mobile) and verify that fullscreen search, navbar, and course preview display correctly.
- Test on desktop and verify that UI is unchanged
- Check that search functionality and course preview works properly

## Issues

Closes #835


site/src/component/AppHeader/AppHeader.tsx
```diff
@@ -59,6 +82,27 @@ const AppHeader: FC = () => {
...
+            </IconButton>
+          </div>
+          <div className="fullscreen-search-row">
+            <div>
+              <SearchModule index="courses" />
+            </div>
+            <IconButton onClick={toggleMobileSearchFilters}>
+              <TuneIcon />
+            </IconButton>
+          </div>
```
CadenLee2: I'm thinking that we should only show the filters icon when filters are actually available (i.e. right now, when we're searching for a course, and in the future, when we actually have searched something). It might be unintuitive to click the button and nothing happens.

For instance, it should probably be hidden in this case:
Image

site/src/component/MobileSearchMenu/MobileSearchMenu.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { setCourse } from '../../store/slices/popupSlice';
+import CoursePreview from '../CoursePreview/CoursePreview';
+
+const MobileSearchMenu: FC = () => {
+  const dispatch = useAppDispatch();
+  const selectedCourse = useAppSelector((state) => state.popup.course);
+
+  const handleClose = () => {
+    dispatch(setCourse(null!));
+  };
```
CadenLee2: Because in `popupSlice.ts` we've defined the course to be stored as `CourseGQLData`, using `null!` seems wrong here. I checked and saw that `popupSlice.ts` seems to default the values to be `null!` anyway, so we should probably change that as the file is quite [outdated](https://github.com/icssc/peterportal-client/blame/main/site/src/store/slices/popupSlice.ts) and we already do have checks around the codebase.

In the slice, we should be able to simply make course be `CourseGQLData | null` and and do the same with professor. This will allow us to get rid of the `!` both in the slice and here.

site/src/component/SearchPopup/SearchPopup.scss
```diff
@@ -25,6 +25,18 @@
   @include globals.card-padding;
 }
 
+.mobile-search-dialog {
+  .search-popup {
+    width: 100% !important;
+    max-width: 100vw !important;
+    height: 100vh;
```
CadenLee2: It's preferred to not use `!important` as it can make things more complicated and harder to change down the road. Typically, we can achieve the desired effect by increasing the specificity instead. Is there a strong justification for its use here?

CadenLee2 (changes requested):
Tested this out and it works well! I had one concern about the behavior (see filter icon comment below) but otherwise didn't find issues while testing.

I also found an issue with toasts... if you go into a course preview from the mobile search then click the Share button, the toast doesn't appear until you close out of mobile search. If that's something easy for you to fix that's related to these changes, go ahead, but otherwise I think we can just make an issue for it and move ahead for the sake of time.

I also have a few smaller suggestions for the code

CadenLee2 (approved):
```
💵L💵💵
💵💵G💵
💵T💵💵
💵💵M💵
```

PR #851, opened 2025-11-15
Search filters

Ascent817:
## Description

Add filters to query courses in a certain department, that fulfill a certain GE, or are of a certain level (Lower, Upper, or Graduate).
Create new `SearchFilter` component, departments slice, and departments hook.

## Screenshots

image

### DIFF
--- a/site/src/component/SearchFilters/SearchFilters.scss
+++ b/site/src/component/SearchFilters/SearchFilters.scss
@@ -45,7 +45,10 @@
     position: sticky;
     top: 0;
     margin-bottom: 8px;
-    background-color: inherit;
     z-index: 1;
+    .MuiInputBase-input {
+      background-color: var(--mui-palette-background-paper);
+      background-image: var(--Paper-overlay);
+    }
   }
 }


Ascent817:
@Awesome-E I wasn't sure how to approach adding styles to the interior elements inside the MUI components, especially the `Autocomplete` component. Any direction would be appreciated!

site/src/component/SearchFilters/SearchFilters.scss
```diff
@@ -0,0 +1,42 @@
...
+}
+
+input#multiple-limit-tags.MuiInputBase-input.MuiOutlinedInput-input.MuiInputBase-inputAdornedEnd.MuiAutocomplete-input.MuiAutocomplete-inputFocused {
+  font-size: 11px;
+  padding: 0;
+}
+
+.MuiInputBase-sizeSmall.MuiInputBase-adornedStart.MuiInputBase-adornedEnd.MuiAutocomplete-inputRoot.css-tzga38-MuiInputBase-root-MuiOutlinedInput-root {
+  padding-right: 6px;
+}
```
Awesome-E: I have a hard time telling what is actually being targeted here. if it's for one specific button, we should add a className to that button and instead target it inside `.search-filters .that-class {}`

site/src/component/SearchFilters/SearchFilters.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { Icon } from '@mui/material';
+
+import { useAppSelector } from '../../store/hooks';
+
+import { FilterOptions, levels, geCategories } from '../../helpers/searchFilters.ts';
+
+interface SearchFiltersProps {
+  selectedFilters: FilterOptions;
+  updateSelectedFilters: (value: FilterOptions) => void;
+}
```
Awesome-E: tbh i think filters belongs in the `searchSlice` rather than as used state. This would make it easier to read from/write to, esp since we have nested components.

site/src/component/SearchFilters/SearchFilters.tsx
```diff
@@ -0,0 +1,157 @@
...
+    backgroundColor: 'var(--overlay1)',
+    borderColor: 'var(--overlay2)',
+    borderWidth: '1px',
+    borderRadius: '6px',
+    '& .MuiOutlinedInput-notchedOutline': {
+      borderRadius: '6px', // Targets the outer 'fieldset' element
+    },
+    width: 'fit-content',
+  };
+  // #endregion
```
Awesome-E: Are these specific to search filters or are they things that can be universally true for PeterPortal's theme in some way (i.e. a variant)

site/src/component/SearchModule/SearchModule.scss
```diff
@@ -3,7 +3,7 @@
 .search-module {
   z-index: 100;
 
-  button {
+  button.input-group-text {
```
Awesome-E: this may not exist once mui search is merged

site/src/component/SearchModule/SearchModule.scss
```diff
@@ -12,6 +12,10 @@
     text-align: center;
     justify-content: center;
   }
+
+  .form-group {
+    margin-bottom: 0;
+  }
```
Awesome-E: this also relies on bootstrap, which is being removed from the search module in #848

Awesome-E (commented):
smaller structural changes, e.g. departments, looks good. There are some large files that are definitely largely work in progress as of now.

I think this is something Caden & I will take a look at right now because we have set up time to meet for stuff relating to search anyways. We'll let you know when we're done making changes to this branch.

PR #854, opened 2025-11-18
Mobile View Review Card Fix

### LINKED ISSUE #832: Review card widths on mobile review management pages
The review cards on "your reviews" and admin sections are too wide on mobile, resulting in sections of them overflowing being impossible to see.

This is likely caused by the edit buttons.

Instead, the widths should work properly (like with review cards on course pages)

Image

CLiang531:
Mobile Review Card Fix

## Description

Changed the styling in the reviewgrid that constrained child elements (reviewcard) to have a minimum width of 500px, which caused overflowing when the screen was narrower than that. Also added some additional styling so that the edit and delete buttons would shift to its own line once screen size decreased to below 500px, and to prevent footer from overflowing as well.

## Screenshots

Before (300px)
image

After (300px and 350px)
image
image

For verifying reviews, I also added some styling so that the buttons would move to their own line as well, and if the screen got too narrow for both buttons, each button was on its own line. There was an issue with the reviewcard-detail overflowing once the screen size increased past 600px, which I fixed as well.

Before
image
image

After
image
image
image
image

## Test Plan

- write a review and then check the /reviews page
- ensure that the styling looks good for both normal screen sizes as well as ones beneath 500px (make sure that the reviewcard now adapts to the screen size and does not overflow) as well as below 350px (check for the buttons moving to next line)

## Issues

Closes #832


site/src/component/Review/ReviewCard.scss
```diff
@@ -37,6 +37,11 @@
     align-items: center;
     height: 24px;
   }
+
+  @media (max-width: 350px) {
+    flex-direction: column;
+    gap: 10px;
```
CadenLee2: Though the wrapping behavior works well below 350px, I think it's still more natural to make this breakpoint a bit wider (like maybe 460px) as currently we can get behavior like this on a small screen that's > 350px:

Image

CadenLee2 (approved):
LGTM 💵

PR #856, opened 2025-11-19
Page metadata for professors and courses

CoreyZhong:
## Description
Added metadata for every course and professor
For courses, formatted where the title is "Course ID | Course Name" with the description being the summary of the course
For professors, formatted where the title is the professor's name and the description being "{Job Title} in {Department}"

Included seperate tags of opengraph and twitter for applications like X and Discord to work correctly

## Screenshots
Course metadata for ICS 46 example 
image
Professor metadata for klefstad example
image

## Test Plan
Go on any course or professor page and press ctrl + U to view the page source and look at the <title> and meta description tags

## Issues

Closes #


site/src/app/course/[id]/page.tsx
```diff
@@ -1,5 +1,39 @@
...
+      description,
+      url: `https://peterportal.org/course/${id}`,
+      type: 'website',
+    },
+
+    twitter: {
+      title,
+      description,
+      card: 'summary',
+    },
```
Comment (Awesome-E) [inline site/src/app/course/[id]/page.tsx]: if you omit these, is Next.js smart enough to automatically use the top-level title+desc to generate opengraph/twitter tags?

site/src/app/course/[id]/page.tsx
```diff
@@ -1,5 +1,39 @@
...
+  const id = decodeURIComponent((await params).id);
+
+  const input = encodeURIComponent(JSON.stringify({ courseID: id }));
+
+  const url = `${process.env.BACKEND_ROOT_URL}/trpc/courses.get?input=${input}`;
+
+  const res = await fetch(url);
+  const json = await res.json();
+
+  const course = json.result.data;
```
Comment (Awesome-E) [inline site/src/app/course/[id]/page.tsx]: Look for how we fetch the current user (presumably in the root layout.tsx, but could be elsewhere) to see how we set up trpc routes when we need data from the server. I'd recommend a similar approach to that instead of manually crafting the result with `fetch()`

Awesome-E (changes requested):

PR #859, opened 2025-11-20
Migrate Badges, Spinner, ProgressBar to MUI Chips, CircularProgress, LinearProgress

fiveminus1:
## Description

Migrated miscellaneous remaining Bootstrap components to MUI's [Chip](https://mui.com/material-ui/react-chip/)* and [Progress](https://mui.com/material-ui/react-progress/).

*There are some chips/pills that are not Bootstrap and just simple CSS. We may want to change them to Chips at some point, but this PR does not do that.

## Screenshots

(not including loading spinners)
Before|After
:-:|:-:
image|image
image|image
image|image
image|image

## Test Plan

- [ ] Chips look good 
- [ ] Loading spinners look good 
- [ ] Accurate/good looking SoC enrollment progress
- [ ] No remaining leftover styles
- [ ] No unintended styles affected as a result of removing Bootstrap styles

## Issues

N/A

### DIFF
--- a/site/src/app/roadmap/planner/AddCoursePopup.scss
+++ b/site/src/app/roadmap/planner/AddCoursePopup.scss
@@ -54,8 +54,5 @@
     display: flex;
     gap: 6px;
     color: var(--text-secondary);
-    .quarter-indicator-container {
-      margin-left: 0;
-    }
   }
 }
--- a/site/src/app/roadmap/planner/Course.scss
+++ b/site/src/app/roadmap/planner/Course.scss
@@ -1,56 +1,51 @@
 .course {
-  box-shadow: 0px 0px 4px var(--blue-primary);
+  box-shadow: 0 0 2px 0 #60616680;
   background: var(--overlay1);
-  border-radius: 15px;
+  border-radius: 4px;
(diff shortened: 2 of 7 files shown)


site/src/component/Schedule/Schedule.tsx
```diff
@@ -109,7 +109,12 @@ const Schedule: FC<ScheduleProps> = (props) => {
             <span className="enrollment-percentage">{Math.round(enrollmentPercent)}%</span>
           </div>
           <div className="progress-bar">
```
CadenLee2: Found more unused styles: check the bottom of `Schedule.scss`, where we have `.progress-bar > div`. This seems to be using the fact that Bootstrap progress bars had a div, which is no longer relevant:

Image

charlieweinberger (approved):
I see that MUI's spinner is a bit thinner than bootstrap's spinner, and that the text inside MUI chips has a slightly larger font and isn't bolded compared to boostrap badges, but MUI's way of styling these isn't worse, just different, so it seems fine to me. 

Lets Get That Mui ~~removed~~ merged 🤑

PR #860, opened 2025-11-21
feat: oidc auth

KevinWu098:
> [!IMPORTANT]  
> This PR contains a database migration that should be run **after** a deploy to prod finishes. This will also require us to remove or merge users with empty string emails, which are presumably an artifact from old reviews when we merged #497

> In addition to the database migration, we will also need to update old `google_id`s for users without a valid email for compatibility reasons.

## Description
This PR removes Passport.js as the authentication provider and implements Arctic's OIDC provider which integrates with https://auth.icssc.club (code: https://github.com/icssc/auth). Sessions continue to be managed by `express-session`.

The primary changes are:
1. Arctic over Passport
2. Changes in `/google/callback` to facilitate the new authentication flow
3. Migration of Google users to OIDC users — this is based on a user's email, which is presumed to be unique, and overwrites the existing Google ID (which is presumed to not be in use presently) <- this is `successLogin`
4. Changes in `/logout` to also clear the OIDC session

My `env` files:

`.env`:
```
PUBLIC_API_URL=https://api.peterportal.org/rest/v0/
PUBLIC_API_GRAPHQL_URL=https://api.peterportal.org/
OIDC_CLIENT_ID=peterportal-dev
OIDC_ISSUER_URL=https://auth.icssc.club
ANTEATER_API_KEY=
DATABASE_URL=
SESSION_SECRET=local-dev-secret-change-in-production
```

`.env.local` (api):
```
PRODUCTION_DOMAIN=http://localhost:3000
OIDC_CLIENT_ID=peterportal-dev
OIDC_ISSUER_URL=https://auth.icssc.club
ANTEATER_API_KEY=
```

`.env.local` (site)
```
BACKEND_ROOT_URL="http://localhost:3000/api"
```

## Issues

Reminder: put up issue making email unique in the DB schema

Closes #


sst.config.ts
```diff
@@ -19,14 +19,18 @@ function getDomainConfig() {
...
     SESSION_SECRET: process.env.SESSION_SECRET!,
     PUBLIC_API_URL: process.env.PUBLIC_API_URL!,
-    GOOGLE_CLIENT: process.env.GOOGLE_CLIENT!,
-    GOOGLE_SECRET: process.env.GOOGLE_SECRET!,
+    OIDC_CLIENT_ID: process.env.OIDC_CLIENT_ID!,
+    OIDC_ISSUER_URL: process.env.OIDC_ISSUER_URL!,
     GRECAPTCHA_SECRET: process.env.GRECAPTCHA_SECRET!,
-    PRODUCTION_DOMAIN: process.env.PRODUCTION_DOMAIN!,
+    PRODUCTION_DOMAIN: productionDomain, // Dynamically set based on stage
+    PRODUCTION_CLIENT_DOMAIN: productionDomain, // Same as PRODUCTION_DOMAIN (frontend on same domain)
```
Awesome-E: if these two are the same value, why do we need a second variable? I do see something related in the `.env.local` related to port numbers, but even then, I believe `localhost:3000/api/path` proxies to `localhost:8080/api/path`

api/src/controllers/auth.ts
```diff
@@ -1,29 +1,56 @@
...
+    userData = await db
+      .insert(user)
+      .values({
+        googleId: sub,
+        name: name ?? '',
+        email,
+        picture: picture ?? '',
+      })
+      .returning();
+  }
```
Awesome-E: Is the separation of upsert into insert + update purely for readability, or is it more for something else, like handling matching by email instead of google ID? I would prefer there to be less "near-duplicate" code if possible

api/src/config/constants.ts
```diff
@@ -1 +1 @@
-export const SESSION_LENGTH = 30 * 86400 * 1000;
+export const SESSION_LENGTH = 1000 * 60 * 60 * 24 * 7; // 1 week
```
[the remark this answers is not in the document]
KevinWu098: Hm, frankly, I didn't do the math and assumed Cursor was simply making the session length more verbose. Reverted

sst.config.ts
KevinWu098: In development, because the client and server are on different domains, the logout functionality of the auth server, if using `PRODUCTION_DOMAIN` would redirect back to 8080. The only way to accomplish the logout redirect without a new environment variable would be to add an additional server-side callback, e.g. `/logout/google/callback`).

I wouldn't mind doing this, but I think it's nicer to simply define where the client is and redirect directly.

api/src/controllers/reviews.ts
```diff
@@ -112,6 +112,11 @@ const reviewsRouter = router({
    */
   add: userProcedure.input(reviewSubmission).mutation(async ({ input, ctx }) => {
     const userId = ctx.session.userId!;
+
+    // Get user information
+    const userData = await db.select({ name: user.name }).from(user).where(eq(user.id, userId)).limit(1);
```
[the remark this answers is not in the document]
KevinWu098: This is fixable — I put the userName back on the session object. Other fields you expect to access a lot can also be placed into the session as well (but this was the only place where we accessed ctx.session.[something] afaik)

api/src/controllers/auth.ts
KevinWu098: Upserted now — happy to put up the issue once this is merged

sst.config.ts
KevinWu098: My understanding here is that 3000 proxies to 8080, but that the inverse is not true.

The logout method must define where to return after the logout has been completed at auth.icssc.club. So production domain (which is 8080) would not be sufficient when developing locally since 8080 does not redirect to 3000.

api/src/controllers/auth.ts
```diff
@@ -1,84 +1,137 @@
...
+      picture: picture ?? '',
+    })
+    .onConflictDoUpdate({
+      target: user.email,
+      set: {
+        googleId: sub,
+        name: name ?? '',
+        picture: picture ?? '',
+      },
+    })
```
Awesome-E: After looking into the database, we saw that (presumably) from a previous migration, some users had Google IDs but not emails. Because some of these users have planner data, I believe we should still allow them to be linked to an existing user via a matching Google ID instead of an email.

This means that we should also target a conflict on `user.google_id` and update the `email`, `name`, and `picture` accordingly. When merging to `main`, we will then update all the users without an email to have a new google_id of `google_[existing_number]` so this can work properly

KevinWu098: Hm, my LLM must have hallucinated when I was investigating this -- thanks for the correction

KevinWu098: I'm game for keying on ID then (is that unique already in the DB?)

PR #861, opened 2025-11-24
fix GE categories filter

Awesome-E:
as title states


Ascent817 (approved):
Seems like it works on deployment!

PR #865, opened 2025-12-01
Automatically detect whether searches are of courses or professors

### LINKED ISSUE #813: Smart search results
Prerequisite: #812 

When the user searches for a professor's name, they expect to find professor results. When they search for "compsci", they likely expect course results.

The professor/course search switcher defined in #812 can be used, but each search should intelligently default to either professors or courses to save work.

## Logic
The exact logic may vary based on how the professor and course search endpoints function. One idea is to compare the length (if the number of returned professors > the number of returned courses, show professors). This will determine whether we want to show professors first or courses first (the user can toggle with the link)

If there are **no professor results**, then only display course results and don't include the link to view professors. Conversely, do the same with course results. If there are no results at all, display a generic "no results" message.

Awesome-E:
still needs:

- [x] Professor preview
- [x] hide the switcher link if there are no results
- [x] don't show prof results in mobile Saved & Search (when adding a course)
- [x] Always set view index to courses when we open the mobile search menu, or consider taking in a forced viewing index if we want to preserve the state of other search containers.
- [x] No overlapping course/prof preview; opening one should close the other.

Biggest Blocker currently: do we want to reset filters on new query?

for future PRs:
- infinite scroll in desktop search results // pagination

## Description

- Changes search logic to perform new searches based on what's changed
  - i.e. when we change the query, we trigger a "new query" search operation; when we change filters, we trigger a "new filters" operation, etc. and each map to a set of requests that we make to the backend.
- Make full screen mobile search work with smart search and fix related styling with course filters
- Implement `<ProfessorPreview />` for professor results since we need a way to show prof info from our new search experience.

## Screenshots

Course|Professor + Professor Preview
:-:|:-:
image|image

Mobile full screen|Mobile Catalog
:-:|:-:
image|image

## Test Plan

behavior should work as intended between (1) desktop sidebar search, (2) full screen mobile search, and (3) full screen catalog search.

Make sure each of these work when the previous search has the following:
- course results shown
- instructor results shown
- filters applied
- filters not applied
- page other than first page (in mobile full screen)

Things to test with the above conditions across all searches:
- Searching for a course-first query, e.g. `ics` or `arth`, should show course results first but have the option to switch to profs (except in mobile catalog)
- Searching for a instructor-first query, e.g. `el` or `ale`, should show instructor results first but have the option to switch to courses (in mobile catalog, it should only show the course results)
- Applying filters should work and refresh results
- Pagination should work on full screen mobile search
- Queries are currently preserved when closing and opening menus
- Course-only queries, such as `compsci`, should switch to showing only course results, with no option to switch.
- Instructor-only queries, such as `thornton`, should switch to showing only instructor results, with no option to switch. (on mobile catalog, this should show "no results")

## Issues

Closes #813

### DIFF
--- a/site/src/component/AppHeader/AppHeader.tsx
+++ b/site/src/component/AppHeader/AppHeader.tsx
@@ -95,7 +95,7 @@ const AppHeader: FC = () => {
             </IconButton>
           )}
           {!isRoadmapPage && (
-            <IconButton href="/">
+            <IconButton component={Link} href="/">
               <HomeIcon />
             </IconButton>
           )}


charlieweinberger:
Typing "stu" gives me an infinite spinner, and this error pops up in the console
Screenshot 2025-12-04 at 10 03 30 PM
Screenshot 2025-12-04 at 10 03 38 PM

charlieweinberger:
Right now, mobile course search and mobile course+prof search share data, i.e. the query/results from one will remain when switching to the other. Is this intentional?

One problem this causes - if I search a course in the course-only search, but then switch to the course+professor search, there's no option to switch to viewing professors, even if there are professors for that search. But if I were to delete and then re-type the search, then the hyperlink to view professors shows up. Not a big problem, but not ideal. For reference, I'm using the query "Irani" to test this.

PR #866, opened 2025-12-01
Mui popovers

### LINKED ISSUE #820: MUI for popovers
Course and recent offerings popovers currently use Bootstrap (see image); they should instead use [MUI popovers](https://mui.com/material-ui/react-popover/).

Popover styling and behavior should remain relatively consistent after the transition. Contents should remain the same.

Image

Image

ellisl10:
## Description

Changed all Bootstrap tooltips, popovers, and overlay triggers to MUI

## Screenshots
Before
Screenshot 2025-12-03 at 1 20 43 PM

After
Screenshot 2025-12-03 at 1 20 06 PM

## Test Plan

Test all popovers work and that certain popovers with mouse hover interactions work

## Issues

Closes #820

### DIFF
--- a/site/src/app/roadmap/planner/Quarter.scss
+++ b/site/src/app/roadmap/planner/Quarter.scss
@@ -10,6 +10,7 @@
     opacity: 0.35;
     width: auto;
     color: var(--blue-primary);
+    pointer-events: none;
 
     .course-info {
       display: contents;


site/src/app/roadmap/planner/Course.tsx
```diff
@@ -48,9 +48,10 @@ export const CourseNameAndInfo: React.FC<CourseNameAndInfoProps> = (props) => {
   return (
     <PPCOverlayTrigger
       popoverContent={popoverContent}
-      placement={isMobile ? 'bottom' : openPopoverLeft ? 'left-start' : 'right-start'}
       popupListener={popupListener}
       disabled={isMobile}
+      anchor={isMobile ? 'bottom' : openPopoverLeft ? 'left' : 'right'}
+      transform={isMobile ? 'bottom' : openPopoverLeft ? 'left' : 'right'}
```
Awesome-E: Since we open directly to the course preview, `isMobile` no longer affects placement
```suggestion
      anchor={openPopoverLeft ? 'left' : 'right'}
      transform={openPopoverLeft ? 'left' : 'right'}
```

site/src/component/PPCOverlayTrigger/PPCOverlayTrigger.tsx
```diff
@@ -1,71 +1,98 @@
...
+            className: 'popover-paper',
+            onMouseLeave: hidePopover,
+            sx: {
+              pointerEvents: 'auto',
+            },
+          },
+        }}
+        sx={{
+          pointerEvents: 'none',
+        }}
```
Awesome-E: Since these are all popovers that we can hover, I'd recommend something more like setting a className on `<Popover/>` itself, perhaps something like `.hoverable-popover`.

That way, you can also use _that_ as your more specific css selector for Paper (see previous comment) and not have to use `sx: { pointerEvents }`.

The change in the stylesheet would be something like:
```scss
.hoverable-popover {
  pointer-events: none;

  .MuiPaper-root {
    // current styles
    pointer-events: auto;
  }
}
```
then you could greatly simplify this:

```suggestion
        className="hoverable-popover"
        slotProps={{
          paper: { onMouseLeave: hidePopover },
        }}
```

Awesome-E (changes requested):
works pretty well on staging and most of the logic changes are sound. Main thing right now is to avoid usage of `!important`, ideally by standardizing a background color for popovers.

Let me know if you'd like anyone else to take a look as well btw cause it's week 10..!

site/src/App.scss
```diff
@@ -117,72 +117,11 @@ div.ppc-popover {
     padding: 16px;
   }
 
-  div.popover-arrow {
```
fiveminus1: beautiful

PR #875, opened 2025-12-05
Remove search pages

### LINKED ISSUE #871: Get rid of course/prof search pages
Get rid of the current sidebar implementation completely, only modifying other code as needed to prevent layout quirks as a result of the sidebar going away.

On mobile, the hamburger menu's code can persist for now, but you ideally we should hide the hamburger menu button so that it can't be opened for now. The rest will be taken care of alongside creating the new planner/scheduler menu.

anthonyj33:
## Description
Remove the dedicated search pages, any references to them, and anything else that's obsolete once they are removed.

## Screenshots
image
image

## Test Plan
- Check that nothing depended on the things being removed
- Make sure that there isn't anything left over that should have been removed
- Try it out on staging

## Issues
Closes #871

### DIFF
--- a/site/src/component/AppHeader/AppHeader.tsx
+++ b/site/src/component/AppHeader/AppHeader.tsx
@@ -13,6 +13,7 @@ import SearchModule from '../SearchModule/SearchModule';
 import GitHubIcon from '@mui/icons-material/GitHub';
 import SearchIcon from '@mui/icons-material/Search';
 import ArrowLeftIcon from '@mui/icons-material/ArrowBack';
+import Home from '@mui/icons-material/Home';
 import TuneIcon from '@mui/icons-material/Tune';
 import SmsIcon from '@mui/icons-material/Sms';
 import { Button, IconButton } from '@mui/material';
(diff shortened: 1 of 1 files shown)


site/src/App.scss
```diff
@@ -37,7 +37,7 @@ code {
 
 .app-body {
   display: flex;
-  height: calc(100% - 72px);
```
Awesome-E: This is a result of a different layout shift which is fixed in #865 - we should revert this

Awesome-E:
> Should I delete the old changelog image?

Up to you

CadenLee2 (changes requested):
Generally solid, but I did notice the sidebar height is off now (look in the bottom left):
Image

charlieweinberger:
Without the sidebar, the github/api/feedback/icssc links are gone - those are being added to the roadmap page, right? when will that happen?

anthonyj33:
> Without the sidebar, the github/api/feedback/icssc links are gone - those are being added to the roadmap page, right? Are those also gonna be added to `universal-planner` being its merged into main?

Those will be added to the bottom of the roadmap page and I'd agree that that should happen before we put universal planner on prod. I think that can be a separate PR though, unless anyone wants it to be in this PR.

anthonyj33:
@charlieweinberger Added a back button on all pages except the roadmap, LMK what you think

anthonyj33:
Let's move this conversation to Discord

PR #877, opened 2025-12-06
Update staging teardown to remove deploy environments when a PR is merged

### LINKED ISSUE #757: Update staging teardown to remove deploy environments when a PR is merged
Currently, inactive deployments are not removed.

Image

Image

We should change it so that when a staging instance goes through teardown, it also removes the environment from the repo

anthonyj33:
## Description
Set onlyDeactivateDeployments to false in remove-staging.yml to allow full removal of deployments instead of just deactivation.

## Screenshots
N/A

## Test Plan
Potential approach: create new branch based on this one, create PR, and close. See behavior.
Not sure if that's the best way to do it.

## Issues

Closes #757


anthonyj33:
Closing because I don't think we intend to revisit this any time soon. See #757 as well.

PR #878, opened 2025-12-06
Create a shared ICSSC Projects footer component

### LINKED ISSUE #872: Create shared footer implementation
> [!IMPORTANT]
> This is going to be a shared component across code bases. Please be extra-careful with style rules applied to individual components that make up the shared footer, as PeterPortal styles will naturally **not** be carried over to other ICSSC projects. This means we should have the **absolute minimum** number of component overrides, theme overrides, and custom styles applied to anything in this component.

In order to create more cohesive branding with AntAlmanac, we would like to create a shared footer. The design should look as seen below, and we should use [SvgIcon](https://mui.com/material-ui/icons/#svgicon) for the ICSSC logo, and buttons with icons for each of the different links (you should be able to add `href`s to the buttons).

Image

We should have "ICSSC Projects", "GitHub", and "Feedback" in that order, and each should open their respective link in a new tab. Additionally, this means we should remove the following:
1. beta badge/popover
2. links in the mobile side nav menu

Awesome-E:
## Description

- Create a `shared-components` folder, which contains components that are meant to be easily transferred across code bases. These components should have zero dependencies from other parts of the code base so that there are no additional adjustments (e.g. styles) that need to be made when using it for another project. These components will use inline styles with MUI's `sx` prop, as that is currently what AntAlmanac does.
- Created the shared `Footer.tsx`, which contains the org link, project GitHub link, and Feedback form link.
- Restored the joyful presence of the week number #lessevilpeterportal

## Screenshots

Desktop|Mobile
:-:|:-:
Screen Shot 2025-12-06 at 11 32 09|image

## Test Plan

- it look good

## Issues

Closes #872

### DIFF
--- a/site/src/component/AppHeader/AppHeader.tsx
+++ b/site/src/component/AppHeader/AppHeader.tsx
@@ -7,15 +7,12 @@ import './AppHeader.scss';
 
 import { useAppDispatch, useAppSelector } from '../../store/hooks';
 import Profile from './Profile';
-import PPCOverlayTrigger from '../PPCOverlayTrigger/PPCOverlayTrigger';
 import SearchModule from '../SearchModule/SearchModule';
 
-import GitHubIcon from '@mui/icons-material/GitHub';
 import SearchIcon from '@mui/icons-material/Search';
 import ArrowLeftIcon from '@mui/icons-material/ArrowBack';
 import HomeIcon from '@mui/icons-material/Home';
-import SmsIcon from '@mui/icons-material/Sms';
-import { Button, IconButton } from '@mui/material';
+import { IconButton } from '@mui/material';
 import Link from 'next/link';
 import { useIsMobile } from '../../helpers/util';
 import { setShowMobileFullscreenSearch } from '../../store/slices/roadmapSlice';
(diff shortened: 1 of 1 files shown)


site/src/shared-components/Footer.tsx
```diff
@@ -0,0 +1,59 @@
...
+          target="_blank"
+        >
+          {icsscText}
+        </Button>
+        <Button
+          variant="text"
+          size="large"
+          color="inherit"
+          startIcon={<GitHubIcon />}
+          href="https://github.com/icssc/peterportal-client"
```
anthonyj33: Since this is a shared component, maybe we shouldn't hardcode this. Can we make the Footer component accept the GitHub and Feedback links as props? Or at least have it accept a prop that differentiates between planner and scheduler, then maps those onto the correct URLs?

anthonyj33:
Also, does AA team not want to have the week number on AA Scheduler?

site/src/shared-components/Footer.tsx
Awesome-E: The only thing this really changes is where we're hard-coding a url. When I mean that something is reusable in the context of shared components for the merge, it's more of the idea that they should be able to use it with minimal changes after copy-pasting, as opposed to being able to call the exact same component from the code.

Should AA and PP merge _code-bases_, I would agree that putting the url here is not a good idea.

anthonyj33: That's true.

I'm not exactly sure what the workflow will look like, but if we wanted to sync the shared components, it may be worth aiming for zero changes needed.

Sounds like that's not the plan though- at least not now- so LGTM! 🤑

PR #879, opened 2025-12-06
Unified Planner view for PeterPortal

### LINKED ISSUE #837: Consolidate "credits" menus into the left sidebar

Awesome-E:
## Description

Merge changes from #839, #849, #865, #875, and #878 into main.

## Screenshots

Desktop Catalog + Footer|Mobile Footer
:-:|:-:
image|image

Universal Search Desktop|Universal Mobile|Mobile Catalog
:-:|:-:|:-:
image|image|image

## Test Plan

- nothing breaks

## Issues

Closes #837, #835, #813, #780, #871, #872


site/src/component/MobileSearchMenu/MobileSearchMenu.tsx
anthonyj33: Good change- I like the filters this way better

anthonyj33 (approved):
Skimmed all the changes (why not) and tested on staging. Let's `git` this `merge`d. Good job team 🫡

site/src/hooks/schedule.ts
```diff
@@ -19,8 +19,8 @@ export function useSetSchedule() {
     trpc.schedule.currentWeek
       .query()
       .then((res) => {
-        const week = res.display.split(' • ')[0];
-        dispatch(setCurrentWeek(week));
+        // const week = res.display.split(' • ')[0];
```
fiveminus1: forgot to remove this?

site/src/shared-components/Footer.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  const stackStyleOverrides = {
+    color: theme.palette.text.secondary,
+    gap: 2,
+    justifyContent: 'center',
+  };
+
+  const showFullText = useMediaQuery('(min-width: 440px)');
+  const icsscLogo = <SvgIcon inheritViewBox component={ICSSCLogo} />;
+  const icsscText = showFullText ? 'ICSSC Projects' : 'ICSSC';
```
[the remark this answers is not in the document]
Awesome-E: Yes, that would affect other components. However, it's not just this one thing that's otherwise problematic. The planner view is already very cramped at the 800px, and I think that making the planner view's minimum desktop width a reasonable value naturally means changing this cutoff.

PR #888, opened 2025-12-10
Migrate all CSS variables to MUI CSS variables

### LINKED ISSUE #881: Use MUI variables for colors
> [!NOTE]
> Before starting this task, wait for PR #879 to be merged and issue #880 to be completed to reduce the risk of merge conflicts.

Currently, `globals.scss` maps mui variables to variables of our own names. E.g:
```scss
  --chart-blue: var(--mui-palette-chart-blue);
  --chart-green: var(--mui-palette-chart-green);
```
The rest of the css code currently uses these custom-named variables.

Instead, we should be directly referencing these mui palette colors across the app when possible, especially considering that most of their names already map directly to the names of our custom variables. Some discretion/discussion may be needed for handling specific cases of non-MUI color variables like `--border-primary`.

This will result in a very large number of changes around the codebase (see the note above).

fiveminus1:
## Description

As an artifact of the MUI theme transition, there are still many instances in the code where we're using custom-named CSS variables. However, all of these CSS variables should now directly map to a corresponding variable generated by the theme. This PR:
* Replaces all color variables to use MUI's theme CSS variables
* Removes all of our color variables
* Defines two `borderRadius` tokens in the theme: the base at `4px`, and a `borderRadiusLg` at `8px`, which maps to our previous `var(--border-radius)`.

Note: `--border-primary` and `--border-secondary` are omitted since Material Design doesn't have a "border color" token and instead uses different shades of the primary, secondary, accent, etc. colors but we should figure out a better solution for those since border colors aren't unified  anywhere - these variables are currently only used in one instance each.

We also bump `@mui/material`, as v7.3.5 [fixed module augmentation for shapes](https://github.com/mui/material-ui/pull/47164).

## Test Plan

- [x] Ensure no instances of the previous CSS variables in the code
- [ ] Verify all variables were correctly changed to the corresponding MUI variable
- [x] Everything should look the same 

## Issues

Closes #881

### DIFF
--- a/site/src/helpers/planner.ts
+++ b/site/src/helpers/planner.ts
@@ -125,13 +125,18 @@ export const expandPlanner = async (savedPlanner: SavedPlannerYearData[]): Promi
 
   return new Promise((resolve) => {
     const planner: PlannerData = [];
+
     savedPlanner.forEach((savedYear) => {
       const year: PlannerYearData = { startYear: savedYear.startYear, name: savedYear.name, quarters: [] };
+
       savedYear.quarters.forEach((savedQuarter) => {
         const quarter: PlannerQuarterData = { name: savedQuarter.name, courses: [] };
-        quarter.courses = savedQuarter.courses.map((course) => courseLookup[course]);
+
+        quarter.courses = savedQuarter.courses.map((courseId) => courseLookup[courseId]).filter((course) => !!course);
+
         year.quarters.push(quarter);
       });
+
       planner.push(year);
     });
     resolve(planner);


site/src/component/Review/ReviewCard.scss
```diff
@@ -211,7 +211,7 @@
     font-size: 20px;
   }
   .colored-upvote {
-    color: var(--blue-primary);
+    color: var(--mui-palette-primary-main);
   }
   .colored-downvote {
     color: var(--red-primary);
```
charlieweinberger: Don't forget to update this `--red-primary`, which was recently added in #882 

```suggestion
    color: var(--mui-palette-error-main);;
```

atile4 (commented):
colors match w/ prod and looks like all CSS are replaced 👍but not 100% sure i caught everything with so many files

saw some @todo's but i'm assuming those weren't for this PR to do

CadenLee2 (approved):
Code changes are as expected. No issues found on staging, everything looks identical to prod (light and dark theme, including grade distribution colors, border radiuses)
Lgtm 💵

PR #890, opened 2025-12-14
Create a shared navigation sidebar component for AntAlmanac

### LINKED ISSUE #873: Create a common side navigation component
> [!IMPORTANT]
> This is going to be a shared component across code bases. Please be extra-careful with style rules applied to individual components that make up the navigation, as PeterPortal styles will naturally **not** be carried over to other ICSSC projects. This means we should have the **absolute minimum** number of component overrides, theme overrides, and custom styles applied to anything in this component.

In order to create more cohesive branding with AntAlmanac, we would like to create a shared side navigation. The design should look as seen below.

Image

For mobile, I will need to check again with AntAlmanac, but we will either put links in the standard Material UI appearance on the left menu, or include a link in the profile menu header. Regardless, both the new mobile menu and desktop switcher should be written using MUI components with absolute minimal styling.

fiveminus1:
## Description

Creates a new shared side navigation component to be used between AntAlmanac Scheduler and Planner.

A link to Scheduler is also added in our Profile menu in mobile, though I believe it's currently up for discussion with UI/UX if we want this to be on both desktop + mobile or just on mobile. 

Presumably, the only changes that should be made between codebases is that Scheduler will be manually setting their link buttons' colors and adjust their z-index. Therefore, those two instances are more specific to our code, using CSS vars for the colors and having a 300 z-index.

## Screenshots

image

image

## Test Plan

- [ ] SideNav looks good in light and dark mode
- [ ] SideNav Buttons work
- [ ] SideNav should not show on mobile
- [ ] Profile navigation to Scheduler works

## Issues

Closes #873

### DIFF
--- a/site/src/shared-components/SideNav.tsx
+++ b/site/src/shared-components/SideNav.tsx
@@ -16,7 +16,7 @@ const SideNav = () => {
     position: 'fixed',
     top: 0,
     left: 0,
-    borderRight: '1px solid #60616680',
+    borderRight: '1px solid var(--mui-palette-divider)',
     borderRadius: 0,
   };
 


site/src/shared-components/SideNav.tsx
```diff
@@ -0,0 +1,52 @@
...
+        left: 0,
+        borderRight: '1px solid #60616680',
+        borderRadius: 0,
+      }}
+    >
+      <Stack direction="column" alignItems="center" sx={{ gap: '16px', paddingTop: '8px' }}>
+        <Link
+          href="https://antalmanac.com"
+          style={{ textDecoration: 'none', color: 'var(--mui-palette-text-primary)' }}
+        >
```
CadenLee2: It seems this may not work with AA Scheduler as Scheduler does not define this variable (not specified in the theme, and instead the text color gets inherited). However, considering that they'll have to swap around the colors anyway, I think it's fine
Image

fiveminus1:
> I assume you mean "just on mobile" at the end?

Yes, sorry

Awesome-E (approved):
The code looks good, but for some reason it looks visually under-scaled relative to the rest of our content.

I think the biggest thing right now is that our PETERPORTAL header is currently taller than AA's, so I would probably wait until we change that in the branding changes branch before we give the header code to them.

If it still looks too small after that, we may consider bumping up the font size to 13 and making the width something like 72, though that is to be seen ofc

PR #902, opened 2025-12-23
Make the profile menu accessible even if signed out

Awesome-E:
## Description

Allows for navigation to scheduler and for theme switching to take place by showing a hamburger menu if the user is signed out. Previously, there was only the sign in button.

would like to hear if you have any suggestions for improvement on this as well

## Screenshots

Mobile|Desktop
:-:|:-:
image|image

## Test Plan

- the menu still works when signed in
- it works when signed out
- it works in light mode
- it works in dark mode

## Issues

N/A

### DIFF
--- a/site/src/component/AppHeader/AppHeader.scss
+++ b/site/src/component/AppHeader/AppHeader.scss
@@ -15,8 +15,7 @@
   position: sticky;
   z-index: 300;
   height: 52px;
-  padding-right: 14px;
-  padding-left: 8px;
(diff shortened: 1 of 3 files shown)


CoreyZhong:
tested in light and dark mode, and signed in/out, works perfectly well. 

having two sign in buttons i think is fine because of what caden said, it shows that the menu will have more options when logged in. i do find it to be a bit redundant but i think its good to keep it for now and possibly remove it later when we have more features in the menu?

PR #903, opened 2025-12-23
Disable restoring focus on Popovers

### LINKED ISSUE #901: Exiting a popover in major requirements can cause scroll jumping
**Describe the bug**
When I trigger the course popover in major requirements, it will sometimes 'randomly' change the scroll back to somewhere else in the major requirements list.

**To Reproduce**
1. Select a major (and spec if needed)
2. Click the collapse _arrow_ (not title) of a requirements block, e.g. the arrow next to "11 upper div electives", to open/close it
3. Scroll up or down
4. trigger a course popover, then move your mouse out

**Expected Behavior**
the list keeps my current position and the popover shows up

**Actual Behavior**
it jumps back such that the arrow i previously clicked is visible again

**Additional context**
I believe this is due to the fact that when a popover is closed, MUI tries to "explicitly refocus" a previous UI element for accessibility reasons. This is not the behavior we want for hovering courses as the user's scroll position should not be hijacked.

Another clear demonstration of MUI's refocus behavior can be seen by selecting course filters, then entering/exiting a popover. You'll see that the filter dropdown gets highlighted again. Though, in regards to search filters, I don't think the popovers cause a problem.

---

A temporary workaround for users is to immediately click on a dead area (e.g. empty space in planner years) after the clicking a collapse button to unfocus it. Then, leaving a popover should not cause this jumping.

fiveminus1:
## Description

As mentioned in the original issue, MUI automatically refocuses a previous element when something like popovers are closed; this was causing annoying scroll jumping behavior. This PR just disables that setting to fix the scroll jumping.

## Test Plan

- [ ] Trigger any popovers, make sure they still appear and behave as they should
- [ ] Try scrolling/move your mouse out after triggering a popover, it shouldn't refocus/change your scroll position

## Issues

Closes #901


Awesome-E (approved):
looks right

PR #907, opened 2025-12-29
Hide description for course dragged into quarter

CadenLee2:
## Description

Restored the `::after` for the `: ` in course descriptions and added a span around the description, allowing them to be selected and hidden by `Quarter.scss` as intended

## Screenshots

| Before | After |
| --- | --- |
| image | image |

## Test Plan

- [ ] None of the places where `CourseDescription` is used should be broken (mobile search results, sidebar search results, course popover)
- [ ] `: description` text is hidden in the ghost when a course is dragged from sidebar search over a quarter

## Issues

Closes #

### DIFF
--- a/site/src/app/roadmap/search/ProfessorResult.scss
+++ b/site/src/app/roadmap/search/ProfessorResult.scss
@@ -27,6 +27,9 @@
       color: inherit;
     }
   }
+  .no-courses {
+    font-style: italic;
+  }
 
   > p {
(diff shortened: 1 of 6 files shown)


Awesome-E (commented):
staging seems to work properly. just one note about naming

Ascent817 (approved):
Deploy looks good, approved

Awesome-E (commented):
mostly good!

PR #915, opened 2026-01-03
Handling missing data for professors

### LINKED ISSUE #887: Handle instructor search results with missing data cleaner
**Describe the bug**
Originally described [here](https://github.com/icssc/peterportal-client/pull/879#pullrequestreview-3553029220).

Some instructor results are missing certain portions of data - this may include a title such as "Professor," a department, or recently taught course history. Currently, the search results will simply display an empty literal for any of these cases, which looks odd especially when unnecessarily delimited. This should be handled more gracefully.

**To Reproduce**
1. Search a professor, probably a broad query like "k"
2. Note the professors with missing data

**Expected Behavior**
* The delimiting "•" if one or both of the first two (title/department) are missing should be hidden.
* Some text noting the missing course history should be displayed such as "*No recent courses*," rather than the current "Recently Taught:".

**Actual Behavior**

Image

Image

DerrickThrower:
conditionally rendered professor information such as recently taught courses or their different positions

## Screenshots
Screenshot 2026-01-02 at 8 16 49 PM

Screenshot 2026-01-02 at 8 17 08 PM

## Test Plan

Ensure that the data is properly rendered and things do not look wonky

## Issues

Closes #887

### DIFF
--- a/site/src/app/roadmap/search/ProfessorResult.scss
+++ b/site/src/app/roadmap/search/ProfessorResult.scss
@@ -27,6 +27,9 @@
       color: inherit;
     }
   }
+  .no-courses {
+    font-style: italic;
+  }
 
   > p {
     margin-bottom: 4px;


site/src/app/roadmap/search/ProfessorResult.tsx
```diff
@@ -22,28 +23,37 @@ const ProfessorResult: FC<{ data: ProfessorGQLData }> = ({ data: professor }) =>
...
-            );
-          }),
-          ', ',
-        )}{' '}
-        {courses.length > 10 && ` + ${courses.length - 10} more...`}
+        {hasCourses ? (
+          <>
+            {addDelimiter(
+              courses.slice(0, 10).map((c) => (
+                <a
```
CadenLee2: Because this is getting pretty deeply nested, I think the courses list should be refactored into a separate component in the same file (will need to take in `courses` as a prop). This lets us call it much more cleanly, like:
```tsx
{hasCourses ? (
  <RecentlyTaughtList courses={courses} />
) : (
  <span className="no-courses">No recent courses</span>
)}
```

site/src/app/roadmap/search/ProfessorResult.tsx
```diff
@@ -5,11 +5,41 @@ import { useAppDispatch } from '../../../store/hooks';
 import { setPreviewedCourse, setPreviewedProfessor } from '../../../store/slices/coursePreviewSlice';
 import { addDelimiter } from '../../../helpers/util';
 
+type RecentlyTaughtListProps = {
+  courses: Array<ProfessorGQLData['courses'][string]>;
```
fiveminus1: I think you can find a better way to type this?

PR #920, opened 2026-01-06
Fix department filter input and select focus persistence

### LINKED ISSUE #918: Department filter input doesn't work
**Describe the bug**
When trying to narrow down the dept list, typing doesn't do anything.

**To Reproduce**
1. Go to search for courses
2. open the dept filters
3. start typing

**Expected Behavior**
it filters based on what i type

**Actual Behavior**
it doesn't

Image

**Additional context**
Not sure if this is related to inputs retaining focus after clicking into a different one, but something with search filters is just being weird.

CadenLee2:
Two MUI select-related bugs have quietly emerged in the past months, and the cause will shock you. See #918 for further context
- Clicking a MUI select then clicking out persists the focus on it anyway
- Filtering department names in search filters has no effect

## Description

The first issue (selects persisting focus when they shouldn't) started in [this commit](https://github.com/icssc/peterportal-client/commit/eaaf27e2975cbc4831a11761bcca1782174d8f39) because `disableRestoreFocus` was being applied to popovers too broadly. Because selects also create popovers, those popovers incorrectly had `disableRestoreFocus` and caused the problem.

The second issue was more dubious: the [blame](https://github.com/icssc/peterportal-client/blame/main/site/src/component/SearchFilters/SearchFilters.tsx) in the relevant file showed no edits since 2 months ago, back when it was working. So what could be going wrong?

As the usual culprit we were familiar with, Bootstrap, is no longer among us (it was [booted](https://github.com/icssc/peterportal-client/pull/886) long ago), and search filters has always been pure MUI from the start, it surely could not have been Bootstrap's fault.

Wrong. Bootstrap has been haunting the codebase from beyond the grave.

To apply the filtered string to menu items, the search filters file sets the `hidden` prop on the MUI `MenuItem`. But [there is no hidden prop](https://mui.com/material-ui/api/menu-item/), and there never has been. In fact, because we passed an unknown `hidden` prop to the MUI component, MUI actually set `hidden=""` on the select's actual HTML element. And back when Bootstrap was still with us, Bootstrap [hid](https://getbootstrap.com/docs/5.3/content/reboot/#html5-hidden-attribute) anything with `[hidden]` using `[hidden] { display: none !important }`. Now that Bootstrap is gone, that no longer applies and the items are all visible regardless of whether they are filtered out or not.

## Screenshots

It's working now:
image

## Test Plan

- [ ] Recreate #918 on prod
- [ ] Ensure that it cannot be recreated on staging, and neither can #901 

## Issues

Closes #918

### DIFF
--- a/site/src/component/SearchFilters/SearchFilters.tsx
+++ b/site/src/component/SearchFilters/SearchFilters.tsx
@@ -91,10 +91,10 @@ function DepartmentSelect() {
       />
       {Object.entries(departments).map(([code, name]) => {
         const labelText = `${code}: ${name}`;
-        const hidden = !labelText.toLowerCase().includes(deptFilterText.toLowerCase());
+        const matchesFilter = labelText.toLowerCase().includes(deptFilterText.toLowerCase());
 
         return (
-          <MenuItem key={code} value={code} className={`search-filter-item ${hidden ? 'hidden' : ''}`}>
+          <MenuItem key={code} value={code} className={`search-filter-item ${!matchesFilter ? 'hidden' : ''}`}>
             {selectedFilters.departments.includes(code) ? <CheckIcon /> : <Icon />}
             <ListItemText className="item-text" primary={labelText} />
           </MenuItem>


site/src/component/SearchFilters/SearchFilters.tsx
```diff
@@ -93,8 +93,10 @@ function DepartmentSelect() {
         const labelText = `${code}: ${name}`;
         const hidden = !labelText.toLowerCase().includes(deptFilterText.toLowerCase());
 
+        if (hidden) return null;
```
Awesome-E: remember the reason we didn't do this is because sometimes the text field would stop working for some queries but not others?

site/src/component/SearchFilters/SearchFilters.tsx
```diff
@@ -94,7 +94,7 @@ function DepartmentSelect() {
         const hidden = !labelText.toLowerCase().includes(deptFilterText.toLowerCase());
```
fiveminus1: It wasn't immediately clear to me what `hidden` was, wonder if renaming to something like `matchesFilter` and swapping the boolean logic might look better?

PR #921, opened 2026-01-06
Fix space in CourseDescription

fiveminus1:
nevermind i lied, one more popover-ish PR 

## Description

I noticed while reviewing the most recent Popover incident that there's a missing space between the title and description of a course popover's description. This is because of the `display: -webkit-box` that's now being applied to it, which causes the JSX whitespace to be ignored. 

## Screenshots

Before|After
:-:|:-:
image|image

## Test Plan

- [ ] Verify the space is there in all locations that use `CourseDescription`

## Issues

N/A


Ascent817:
It looks like you have a deployment failure?

fiveminus1:
Seems to be fixed, so closing

PR #923, opened 2026-01-08
Redesign profile button

CadenLee2:
Made a small change as per prior discussion

## Description

- The profile/menu button no longer has a background, except when hovered, and its padding/sizing is adjusted
- The save button is now larger to match AntAlmanac Scheduler's current button sizes
- It is expected that Scheduler will conform to the same styling to achieve an identical appearance

## Screenshots
| Normal | Hovered | Open | Signed out |
| --- | --- | --- | --- |
| image | image | image | image |

## Test Plan

- [ ] Ensure this looks as expected when signed in and out. Suggestions for change are welcome

## Issues

Closes #


Awesome-E (changes requested):
This should be in shared components if the implementation is to be the same across both apps

KevinWu098 (commented):
LGTM — just the bit from @ ethan and then i think we're good

PR #924, opened 2026-01-08
Planner base path

CadenLee2:
AntAlmanac planner now uses the base path `/planner`, so that once the merge occurs it can be accessed at `antalmanac.com/planner`

## Description

Updated the Next.js app's basePath to be `/planner`
- This means the backend is available on `/planner/api`
- `/` redirects to `/planner` (this only applies to staging/local: on prod, going to `https://antalmanac.com` will go to Scheduler so the Planner app will never have the path `/`)
- Most links work automatically, but plain `a` components had to be changed to Next.js links and some MUI components with hrefs needed their hrefs to be updated to include `/planner`. Open to suggestions for making this more consistent

SST config is updated to use the right path and to provide it correctly in environment variables (e.g. `BACKEND_ROOT_URL`).

The staging GitHub workflow is not updated to provide `/planner` in the URL as the redirect exists and technically the Next.js app is running on the root of `planner-123.antalmanac.com`, but this can be updated if desired.

Auth redirects to the proper URLs.

## Screenshots

image
image

## Test Plan

Test locally and on staging:
- [ ] All links work (they all include the base path `/planner`)
- [ ] Root `/` directs to `/planner`
- [ ] Deploys to staging function as expected
- [ ] Logging in and out continues to work
- [ ] No images are broken

## Issues

Closes #

### DIFF
--- a/site/src/component/ResultPreview/CoursePreview.tsx
+++ b/site/src/component/ResultPreview/CoursePreview.tsx
@@ -82,7 +82,7 @@ const CoursePreview: FC<{ courseId: string; onClose: () => void }> = ({ courseId
   const dispatch = useAppDispatch();
 
   const copyCourseLink = () => {
-    const url = new URL('/course/' + encodeURIComponent(courseId), location.origin).toString();
+    const url = new URL('/planner/course/' + encodeURIComponent(courseId), location.origin).toString();
     navigator.clipboard.writeText(url);
     dispatch(setToastMsg('Copied course URL to clipboard!'));
     dispatch(setToastSeverity('success'));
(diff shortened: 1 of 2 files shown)


CadenLee2:
Converted to draft because of an issue in staging that needs to be figured out. Jeremiah had experienced this once before when updating a branch from branding changes and it seemed to resolve itself after restarting, but we'll need to make sure this kind of thing doesn't happen when eventually merging to prod as well.

From the fact that `planner/theme-script.js` does not exist but `/theme-script.js` does, it seems there's maybe some issue going on related to caching, since it should be served on the new base path like it is locally.

Will look into this more
image

PR #925, opened 2026-01-09
Set default SoC to be the most recent quarter with offerings

### LINKED ISSUE #913: SoC shouldn't always default to the current quarter
Currently, when you click on a course or instructor, their SoC defaults to the current quarter. I personally find that this isn't always convenient, and would rather see their most recently offered quarter. 

If the argument is that the intended behavior is to show that the current quarter is not offered, I believe this is discoverable enough by simply clicking on the dropdown or observing that the default selected quarter is not the current quarter.

Of course, all of this is open to discussion.

atile4:
Set default SoC 🧦

## Description
Set SoC in course preview to default to the most recent quarter with course offerings.

- modify `Schedule.tsx` variable `selectedQuarter` state to initially be set to the most recent offered quarter
- added a message to alert user if the course isn't currently being offered in that quarter

## Screenshots

Before: (uses current quarter, no classes offered)
image

After:
image

Dark mode:
image

Nothing changes if course is being offered in current quarter
image

## Test Plan
- Check a class that **has current quarter offerings** and verify the dropdown defaults to the current quarter
- Check a class that **doesn't** have current course offerings and verify dropdown defaults to a past quarter with offerings
    -  can check ART 9a and a good amount of BIOSCI classes
- Check that class offerings match up with the current quarter

**Note:** if it exists, test with a course that has never had an offering before. I couldn't find one while I was testing.

## Issues

Closes #913

### DIFF
--- a/site/src/component/Schedule/Schedule.scss
+++ b/site/src/component/Schedule/Schedule.scss
@@ -3,6 +3,9 @@
   align-items: center;
   gap: 6px;
   margin-bottom: 16px;
+  p {
+    margin: 0px;
+  }
 }
 
 .schedule-quarter {
--- a/site/src/component/Schedule/Schedule.tsx
+++ b/site/src/component/Schedule/Schedule.tsx
@@ -7,7 +7,7 @@ import { hourMinuteTo12HourString } from '../../helpers/util';
 import { useAppSelector } from '../../store/hooks';
 import trpc from '../../trpc';
 
-import { MenuItem, Select, Stack } from '@mui/material';
+import { MenuItem, Select } from '@mui/material';
 import InfoOutlineIcon from '@mui/icons-material/InfoOutline';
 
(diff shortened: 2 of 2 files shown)


.github/actions/setup-pnpm/action.yml
```diff
@@ -10,7 +10,7 @@ runs:
     - uses: pnpm/action-setup@v3
       name: Install pnpm
       with:
-        version: 10
+        version: 9
```
CadenLee2: Since pnpm was updated to 10, this file (plus `package.json`, `pnpm-lock.yaml`, and `pnpm-workspace.yaml` should not be changed. This may be the result of some local pnpm issues you have, so after resolving that to make sure you're on pnpm 10, make sure these files don't use pnpm 9

CadenLee2 (changes requested):

PR #927, opened 2026-01-09
Feat: store opened state for majors and minors

### LINKED ISSUE #625: Store which majors/minors
Similar to #624, we should store individual the collapsed/expanded state of an individual major or minor. However, this should be done after #620 is closed.

EfinZhao:
When you expand or collapse major or minor groups, it is now added to a global state, thus preserving it when switching tabs.

## Description

- Very similar methodology to @CadenLee2's PR for #624 
- Variables `expandedMajors` and `expandedMinors` were added to the course requirements slice that store the `majorId` and `minorId` of the majors and minors that have been expanded
- Thus, each major and minor now have unique `majorId` and `minorId` strings to store whether or not it has been expanded.
- New behavior resulting from this:
    - If you switch tabs between major, minor, and GE requirements, the opened states are all preserved for each, independently.
    - The opened states are preserved independently for each major of multiple majors, and each minor of multiple minors.
    - If a major is added, some groups are expanded, the major is removed, and the major is re-added, the opened states will stay the same (specializations have no impact on this)
- If at all possible, a review of the code and advice on how to improve for future PRs and contributions would be greatly appreciated. Thanks!

## Screenshots
Below: what the Major tab looks like just after having switched back from the Minor tab
Screenshot 2026-01-09 at 1 41 28 AM

Below: what the Minor tab looks like just after having switched back from the Major tab
Screenshot 2026-01-09 at 1 41 59 AM

## Test Plan
- Test all new behavior described
- Test that existing related behavior still works as intended

## Issues

Closes #625


site/src/store/slices/courseRequirementsSlice.ts
```diff
@@ -30,6 +30,8 @@ const courseRequirementsSlice = createSlice({
     geRequirements: [] as ProgramRequirement[],
     completedMarkers: {} as Record<string, boolean>,
     expandedGroups: {} as ExpandedGroupsList,
+    expandedMajors: {} as Record<string, boolean>,
+    expandedMinors: {} as Record<string, boolean>,
```
Awesome-E: Think about how we currently create keys for requirement groups -- would there be a way to apply similar techniques to generate a unique key for majors and minors so they can just be stored in `expandedGroups`?

Awesome-E (changes requested):

CadenLee2:
Because #792 already resolves this, unfortunately this should no longer be needed

PR #930, opened 2026-01-09
Remove Years Column on Planner Table

CLiang531:
Removing the years column from planner table in database. Dev database is currently migrated, prod database needs to be migrated after merging.

See #769 for more details

## Screenshots

N/A

## Test Plan

Open drizzle studio and check that the years column is removed, loading planners should continue to work as expected.

## Issues

Closes #


CadenLee2 (approved):
Looks good! Don't merge yet though, I'll message you about the migration process

PR #931, opened 2026-01-11
Fix external roadmaps endpoint for new Google IDs

CadenLee2:
Users who have logged in after unified auth now have Google IDs prefixed by `google_` in the PeterPortal database; users who have not done so do not have this prefix, so both kinds of Google IDs exist in the database. The external roadmaps `getByGoogleID` endpoint does not currently handle this case.

For context, this endpoint is intended to be used by other ICSSC applications for future integrations with PeterPortal user data.

## Description

The caller is now expected to only pass the `GOOGLEID` format. Then, both the `GOOGLEID` and the `google_GOOGLEID` formats will be checked.

It is expected that there is only one copy of the user in the db (i.e. there is no id `ABC` for which the database has both an `ABC` and a `google_ABC` id simultaneously).

## Screenshots

N/A

## Test Plan

When testing, you will need to set the `EXTERNAL_USER_READ_SECRET` environment variable locally (to anything you want). Check the dev database (e.g. via Drizzle Studio) for testing.

To test locally you can use the endpoint like:
```js
fetch(`http://localhost:3000/api/trpc/external.roadmaps.getByGoogleID?input=${encodeURIComponent(`{
  "googleUserId": "GOOGLEIDGOESHERE"
}`)}`, {
  method: "GET",
  headers: {
    "Authorization": "Bearer AUTHGOESHERE"
  }
});
```

- [ ] Test a user in the dev database who has a legacy google ID: try the endpoint with this ID
  - The user's roadmaps should be returned
- [ ] Test a user in the dev database who has a new (`google_`) google ID: try the endpoint with the non-prefixed version
  - The user's roadmaps should be returned
- [ ] Test a nonexistent user (such as a google ID `idontexist`)
  - An empty array should be returned (as per the original behavior)

## Issues

Closes #


api/src/controllers/external/roadmap.ts
```diff
@@ -12,7 +12,11 @@ const externalRoadmapsRouter = router({
       throw new TRPCError({ code: 'UNAUTHORIZED' });
     }
 
-    const planners = await queryGetPlanners(eq(user.googleId, input.googleUserId));
+    const idLegacy = input.googleUserId.replace('google_', '');
+    const idPrefixed = `google_${idLegacy}`;
```
Awesome-E: does this imply that `input.googleUserId` only has `google_` sometimes? because otherwise:
```suggestion
    const idPrefixed = input.googleUserId;
    const idLegacy = idPrefixed.replace('google_', '');
```

Awesome-E: if so that should be commented

CadenLee2: This was originally the expectation but it sounds like in actual project use cases the provided google ID will never have the `google_` prefix, so it probably will be better to do this instead

CLiang531 (approved):
LGTM! All three cases work as expected.

PR #932, opened 2026-01-11
Preview links + back button

### LINKED ISSUE #914: Update SoC sections for better interactivity and to conform better to Scheduler's
Our's
Image

Scheduler's
Image

**1. Click on an instructor/course to preview their page (would probably need some form of "return" feature to go back to the previous preview)**
2. Click on an code to copy it
3. Sections/Units -> Type

In the further future, it would be cool to have some cross-functionality with Scheduler s.t. you could add a current quarter's course offering to Scheduler, but I think that'd be pretty far off.

I didn't look too far into this, but I believe the API endpoint only provides instructor names rather than IDs, so #1 would likely be quite involved.

Not much more to add than what I described earlier.

Instead of preview, can just open their page in new tab if the previous preview is too complicated to implement.

**Additional context**
1 is motivated by my personal gripe in using PP to discover courses/instructors...one probably common flow is looking up a course and wanting to quickly view all of the instructors that teach that course. You can either just look up the instructor or try to scroll for a review that has a link to the instructor. The obvious answer is just do the former, but it's inconvenient if you're lazy and it'd be better UX to just have the direct links available somehow.

2 may not be especially important since you'd usually use Scheduler to copy course codes into WebReg, but I still think it'd be a nice-to-have.

Ascent817:
## Description

* Add stack for navigation
* Add back button when multiple previews deep
* Exit button clears all previews

## Test Plan

Verify that all preview links work

## Issues

Partially closes #914

### DIFF
--- a/site/src/app/roadmap/planner/Course.tsx
+++ b/site/src/app/roadmap/planner/Course.tsx
@@ -9,7 +9,6 @@ import { useIsMobile, pluralize, formatGEsTag, shortenCourseLevel } from '../../
 import { CourseGQLData } from '../../../types/types';
 import { setActiveCourse, setShowAddCourse, setActiveMissingPrerequisites } from '../../../store/slices/roadmapSlice';
 import { useAppDispatch, useAppSelector } from '../../../store/hooks';
-import { useCurrentPreview } from '../../../hooks/preview';
 
 import { IconButton } from '@mui/material';
 import WarningAmberIcon from '@mui/icons-material/WarningAmber';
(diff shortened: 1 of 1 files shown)


site/src/app/roadmap/RoadmapPage.tsx
```diff
@@ -15,13 +15,13 @@ import ProfessorPreview from '../../component/ResultPreview/ProfessorPreview';
 import MobileSearchMenu from '../../component/MobileSearchMenu/MobileSearchMenu';
 import MobilePopup from './MobilePopup';
 import { Fade, useTheme } from '@mui/material';
-import { setPreviewedCourse, setPreviewedProfessor } from '../../store/slices/coursePreviewSlice';
+import { clearPreviews, removePreview } from '../../store/slices/previewSlice.ts';
```
fiveminus1: Same here 😅

site/src/app/search/CourseHitItem.tsx
```diff
@@ -12,7 +12,7 @@ import { useSavedCourses } from '../../hooks/savedCourses';
 import { Chip, IconButton } from '@mui/material';
 import BookmarkBorderIcon from '@mui/icons-material/BookmarkBorder';
 import BookmarkIcon from '@mui/icons-material/Bookmark';
-import { setPreviewedCourse } from '../../store/slices/coursePreviewSlice';
+import { addPreview } from '../../store/slices/previewSlice.ts';
```
fiveminus1: . 😅

PR #933, opened 2026-01-12
upgrade next.js to version 16

### LINKED ISSUE #898: Consider upgrading to Next.js 16
Upgrade to the newest version of Next.js (v16)

https://nextjs.org/docs/app/guides/upgrading/version-16

We would probably also want to make generic improvements to our dev environment with this PR, which among other things, would include the ability to make HMR work properly with updated style sheets and would make page/app-wide recompiles not reset the Redux state (i.e. roadmap should not go into infinite loading...)

Awesome-E:
## Description

bump next.js to version 16. HMR with scss actually seems to be working with Turbopack, which is a nice surprise coming from when we tried to get it to work with Next.js 15.

I am seeing that changing style sheets takes effect without re-rendering everything. The nextjs status bubble seems to be stuck saying compiling but i don't think that has broken anything.

## Screenshots
N/A

## Test Plan

- deployment etc. works

## Issues

Closes #898 

let's merge after #770


Awesome-E:
lint checks should be fixed!

PR #934, opened 2026-01-12
another concept to change the sidebar to make it slightly better

Awesome-E:
## Description

Centralizes some navigation to the sidebar to make it "better". As a replacement for the missing buttons, this also adds a way to show the change log again

Currently mobile does not have the feedback/org link, but I will make further changes to mobile footer/menu if we decide we like this better for desktop.

## Screenshots

image

## Test Plan

- staging something
- test that change log still triggers when `localStorage.changelogSeen` is missing & new button works
- new links work (pending universal feedback form, cc @KevinWu098)

### DIFF
--- a/site/src/shared-components/Logo.tsx
+++ b/site/src/shared-components/Logo.tsx
@@ -29,6 +29,7 @@ import {
   Typography,
 } from '@mui/material';
 import { useState } from 'react';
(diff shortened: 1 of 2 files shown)


charlieweinberger:
@CadenLee2 this pr should be closed, right?

PR #936, opened 2026-01-14
ci: use AntAlmanacRouter

KevinWu098:
## Issues

Closes #

### DIFF
--- a/site/src/shared-components/SideNav.tsx
+++ b/site/src/shared-components/SideNav.tsx
@@ -1,8 +1,8 @@
 'use client';
 
-import { Paper, Stack, Typography } from '@mui/material';
 import EventNoteIcon from '@mui/icons-material/EventNote';
 import RouteIcon from '@mui/icons-material/Route';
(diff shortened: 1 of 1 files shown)


CadenLee2 (approved):
This generally works as intended and I think it should be merged. The following tasks can be handled in the base `planner-base-path` PR once this is merged into it:
- Planner api not working (I have an idea for the fix)
- Possible mui link-related issues as mentioned earlier (will test)
- Logic to get the correct router for stagings/shared staging/prod

KevinWu098:
> This generally works as intended and I think it should be merged. The following tasks can be handled in the base `planner-base-path` PR once this is merged into it:
> 
> * Planner api not working (I have an idea for the fix)
> * Possible mui link-related issues as mentioned earlier (will test)
> * Logic to get the correct router for stagings/shared staging/prod

Okay sounds good — I'll leave the merging to you, in case anything is in flight that matters. On the planner side, I'm going to leave the PR up for a sec since I need to configure some stuff on the staging branch (`scheduler`) but it shouldn't affect planner side

PR #939, opened 2026-01-15
Improve Prerequisite Checking

### LINKED ISSUE #868: Improve Popover & Course Preview prerequisite checking
Currently, the course preview shows a list of prerequisites, which will be shown as complete if completed by courses from any year/quarter in the current roadmap. While this is generally good, it may have confusing outcomes[^1].

No

This issue proposes that the course preview only consider courses up until the quarter that the previewed course is in (if the course appears multiple times in the roadmap, use the first quarter it appears).

There are some additional considerations:
- Should there be an intermediate state where a requirement is met but not in time for the course?
- If we don't have an intermediate state, there is now a potential inconsistency between course previewed before added to roadmap and after added to roadmap without a clear indication that it's because of ordering. How should that be addressed, if at all?

Additionally, this should replace the current "Missing prerequisites text" to be similar to the one shown in the course preview; that is, a "live checklist" of the prerequisites with each one checked or "invalid".
- This addresses the readability issue of not knowing which missing prerequisites are `AND` vs `OR` requirements.

**Additional context**
An ideal approach would involve creating some helper functions and/or hooks that returns all courses up until the first occurrence of a target course ID. It would have a function signature similar to this, since `taken` and `taking` are the terms used in prerequisite checks: `function useClearedCoursesUntil(courseId: string): { taken: Set<string>, taking: Set<string> }`

[^1]: Example: if ICS 45J comes after CS 122B, this is invalid data since 122b requires 45j. However, clicking into the course preview for 122b will show that all prerequisites are complete since 45j is still somewhere in the roadmap.

MajdKatabi:
## Description

- Update helper function to check taken courses for prerequisites 
- Update course popover, summary, and search menu course card to include error message if prerequisite is missing

## Screenshots

Before:
image

After:
image

## Test Plan

Add ICS 32 in a quarter and then ICS 31 in the quarter after. Check popover, course preview, and search menu ICS 32 card to see if the prerequisite error now shows.

## Issues

Closes #868

### DIFF
--- a/api/src/controllers/reviews.ts
+++ b/api/src/controllers/reviews.ts
@@ -71,7 +71,8 @@ async function getReviews(
     .leftJoin(vote, eq(vote.reviewId, review.id))
     .leftJoin(user, eq(user.id, review.userId))
     .leftJoin(userVoteSubquery, eq(userVoteSubquery.reviewId, review.id))
-    .groupBy(review.id, user.name, userVoteSubquery.userVote);
+    .groupBy(review.id, user.name, userVoteSubquery.userVote)
+    .orderBy(desc(sql`COALESCE(SUM(${vote.vote}), 0)`), desc(review.createdAt));
 
   if (results) {
     return results.map(({ review, score, userDisplay, userVote }) =>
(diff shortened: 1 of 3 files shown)


CadenLee2 (commented):
Testing:
- Followed the test plan, worked as expected
- Prereq checking continues to work when the course is not present on the roadmap
- No off by one errors (works as expected when a prereq is in the same quarter)

The code changes look great, just included one note about naming

PR #941, opened 2026-01-18
Add CHC Requirements to GE Tab

anthonyj33:
## Description

Add support for CHC Requirements to the GE tab. Currently, API only support 4-year CHC, so we will have to revisit 2-year soon (TM).

WARNING: Do prod DB migration before merging!!!

## Screenshots
| Sidebar CHC Reqs view |
|--------|
| image | 

## Test Plan
- [ ] Saves to DB and local storage properly
- [ ] Works with existing local storage
- [ ] UI looks good
- [ ] Clearing reqs works as expected


CadenLee2 (commented):
Tested:
- With a previous local storage (worked as expected: worked fine without `chc`, added `chc` when selected, and set `chc` to `''` when deselected.
- Requirements clear correctly
- Opened/closed sections persist when switching tabs

The research requirement is a bit funny, but the behavior is acceptable. Considering Derrick's reply, I agree with Anthony that although the requirement nesting looks somewhat odd we should avoid hardcoding the requirement structure.

The code also looks good.

CadenLee2 (approved):
Migrated prod, LGTM 💵

PR #944, opened 2026-01-22
Click to copy section code and update SoC section appearance 

### LINKED ISSUE #914: Update SoC sections for better interactivity and to conform better to Scheduler's

fiveminus1:
## Description

This PR adds a "click to copy section code" feature, reorganizes section information into a "Type" column, and adds more hover information to the "Restriction" column in order to conform better with AntAlmanac Scheduler's section table. A lot of this is adapted from Scheduler's [SectionTable](https://github.com/icssc/AntAlmanac/tree/main/apps/antalmanac/src/components/RightPane/SectionTable).

## Screenshots

Before|After
:-:|:-:
image|image

## Test Plan

- [ ] Go to a course page, click on a section code and ensure it copies to clipboard
- [ ] "Copied section code" toast is appropriate and behavior is good
- [ ] Reorganized section information is accurate and looks fine
- [ ] Ensure different section restrictions' tooltips appear fine (e.g. a single restriction A, multiple restrictions D and B)

## Issues

Partially closes #914

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -12,17 +12,24 @@ on:
 
 # do not cancel in progress, SST will be stuck in a "locked" state if cancelled mid-deployment
 concurrency:
-  group: staging-${{ github.event.pull_request.number }}
+  group: staging-${{ github.event.pull_request.number || github.event.client_payload.pull_request.number }}
 
 jobs:
   build_and_deploy:
     name: Build and deploy
     runs-on: ubuntu-latest
+    env:
+      PR_NUM: ${{ github.event.pull_request.number || github.event.client_payload.pull_request.number }}
+      PR_SHA: ${{ github.event.pull_request.head.sha || github.event.client_payload.pull_request.head.sha }}
+      FULL_NAME: ${{ github.event.pull_request.head.repo.full_name || github.event.client_payload.pull_request.head.repo.full_name }}
     # don't run if labeled "no deploy" && don't run on PRs from forks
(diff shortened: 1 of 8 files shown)


site/src/helpers/schedule.tsx
```diff
@@ -0,0 +1,32 @@
+export const restrictionsMapping: Record<string, string> = {
+  A: 'A: Prerequisite required',
+  M: 'M: Non-major only',
+  E: 'E: Freshmen only',
```
CadenLee2: Unfortunate that we have to hardcode the restrictions mapping, but API doesn't provide it so there's no choice. Perhaps can be improved in the future

CadenLee2 (approved):
Tested on various screen sizes and the new soc appearance looks good. Copying works, including on mobile. LGTM
Image

CadenLee2 (approved):
Sorting works
Image

PR #953, opened 2026-01-28
Jump to top button

### LINKED ISSUE #940: Infinite scroll jump to top button
Prerequisite: #904 

When in the infinite scroll, particularly in the sidebar search (but also in mobile search given filters), users may want to scroll to the top quickly to edit filters or change their search query. Currently (as of #904) the only ways to do so are manually scrolling or hitting the end key.

Instead, it would be helpful to have a "scroll to top" button appear in the bottom-right when the user has scrolled down sufficiently far. This is a common pattern on various websites.

Current appearance when scrolled down:
Image

charlieweinberger:
## Description

Added a button in the roadmap sidebar search (on desktop), mobile catalog search, and mobile full search that, when clicked, jumps to the top of the infinite scroll container. This button only appears when the container has been scrolled 300px, and disappears when the user scrolls back up.

## Screenshots

Before:

| Before | After |
|---|---|
| Screenshot 2026-02-07 at 4 54 50 PM | Screenshot 2026-02-07 at 4 55 16 PM |

## Test Plan

- [ ] Ensure button scrolls to the top
- [ ] Ensure button only appears when the user scrolls a certain initial amount
- [ ] Ensure visual scrolling bug that @Ascent817 discovered from previous PR is resolved

## Issues

Closes #940


charlieweinberger:
@CadenLee2 does our codebase have any instances of buttons that only have an icon (i.e. circular, no background, no text, use `<IconButton />`), but are still look like the primary buttons (i.e. blue, become a darker blue on hover, use `<Button />`)? I can only find buttons that are one or the other, so I'm not sure how to get this jump to top button to look like I'm imagining (i.e. circular, icon only, but also blue background and hover?)

site/src/component/ScrollToTopButton/ScrollToTopButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+const ScrollToTopButton: FC<ScrollToTopButtonProps> = ({ scrollableTarget, raiseButton = false }) => {
+  const [visible, setVisible] = useState(false);
+
+  useEffect(() => {
+    const el = document.getElementById(scrollableTarget);
+    if (!el) return;
+
+    const onScroll = () => {
+      setVisible(el.scrollTop > 300);
+    };
```
CadenLee2: May want to make this `300` a constant (i.e. `const minDistBeforeVisible = 300;`) so it's immediately clear what it's for

anthonyj33 (approved):

PR #960, opened 2026-01-31
Changelog

anthonyj33:
## Description
Chagelog

## Screenshots
image

## Issues

Closes #


charlieweinberger:
Should the image used contain a bit of the sidebar/other page content below the header, so that it's clear the image is showing the header? e.g. smthn like screenshot below, or maybe w/ a bit more of the page content, not sure how much tho since we want the focus to be the switcher

Screenshot 2026-01-30 at 5 37 27 PM

PR #963, opened 2026-01-31
ci: peterportal redirect

KevinWu098:
## Issues

Closes #


CadenLee2 (approved):
Works (https://peterportal.org/)

Shouldn't need to be run in the future, and based on the changes this should have no impact on the usual deploy process. Good to have for future reference

PR #968, opened 2026-02-04
Converted react-select to MUI components

### LINKED ISSUE #908: Update selects to use MUI
We currently use `react-select` in some places. Switching to MUI for all selects will fix hydration errors, enforce visual consistency, and allow us to remove some one-off styling rules.

Places `react-select` is used include:
- Transfer credits (adding AP exams and searching for a course)
- Review form (select up to 3 tags)
- Major/minor/spec selection

In all of these places, it should be replaced with the equivalent MUI component and keep the same functionality. This will likely involve updating theming, though we don't need it to look 1-1 with how it does now.

It can then be removed as a dependency.

DerrickThrower:
Migrated the React Select to Mui alternative

## Description

MajorSelector: Multi select for majors → MUI Autocomplete
MinorSelector: Multi select for minors → MUI Autocomplete 
MajorCourseList: Specialization select → MUI Select with MenuItem
CoursesSection:  course search → MUI Autocomplete 
ReviewForm: Tags select  → MUI Autocomplete 

## Test Plan

Make sure it doesn't break everything
also lmk about styling because it isn't one to one 

Closes #908


fiveminus1 (changes requested):
Haven't looked at everything yet, but on first glance some styling should be refactored appropriately into the SCSS or into the MUI theme.

On a side note, PR titles should be more descriptive since eventually this will be squash merged and future contributors will see this title when viewing revision history. Please title this something along the lines of "Converted react-select to MUI components."

CadenLee2 (commented):
Looks like there are some merge conflicts that need to get fixed, also had a comment about styling

site/src/app/roadmap/transfers/APExamsSection.scss
```diff
@@ -5,10 +5,27 @@
...
+    &::before {
+      content: '';
+      position: absolute;
+      right: 32px;
+      top: 8px;
+      bottom: 8px;
+      width: 1px;
+      background-color: var(--mui-palette-divider);
+      pointer-events: none;
+    }
```
CadenLee2: I'm seeing a lot of code duplication of this, and I'm also wondering whether we want this vertical bar in the first place.

If we do want to keep it we should probably use a [mixin](https://sass-lang.com/documentation/at-rules/mixin/), but was the only purpose of this to look more like the originals? I don't think it's too necessary (without it, in the screenshot, looks fine to me)
Image

CadenLee2 (commented):
A few minor things to change...

Looks very good on staging though

CadenLee2 (approved):
LGTM! Let's see what @fiveminus1 thinks

site/src/app/roadmap/transfers/CoursesSection.tsx
```diff
@@ -47,38 +45,59 @@ const CourseCreditMenuTile: FC<{ course: TransferWithUnread<TransferredCourse> }
 
 const CoursesSection: FC = () => {
   const courses = useAppSelector((state) => state.transferCredits.transferredCourses);
-  const [timeout, setTimeout] = useState<number | null>(null);
-  const [abortFn, setAbortFn] = useState<() => void>();
+  const [options, setOptions] = useState<CourseSelectOption[]>([]);
+  const [inputValue, setInputValue] = useState('');
```
fiveminus1: Would like to rename this to `courseSearchValue` or something clearer

site/src/app/roadmap/transfers/CoursesSection.tsx
```diff
@@ -47,38 +45,59 @@ const CourseCreditMenuTile: FC<{ course: TransferWithUnread<TransferredCourse> }
...
-        if (response) resolve(response);
-      };
-      setTimeout(window.setTimeout(initSearch, 300));
-    });
-  };
+    setLoading(true);
+    const abortController = new AbortController();
+    abortControllerRef.current = abortController;
+
+    timeoutRef.current = window.setTimeout(async () => {
```
fiveminus1:
```suggestion
    const timeout = window.setTimeout(async () => {
```

site/src/app/roadmap/transfers/APExamsSection.scss
```diff
@@ -5,10 +5,12 @@
 
   .exam-input {
     width: 216px;
+    color: grey;
```
fiveminus1:
```suggestion
```
unnecessary?

site/src/app/roadmap/catalog/MajorCourseList.tsx
```diff
@@ -147,16 +146,18 @@ const MajorCourseList: FC<MajorCourseListProps> = ({ majorWithSpec, onSpecializa
...
             options={specOptions}
-            value={specOptions.find((s) => s.value.id === (majorWithSpec.selectedSpec?.id ?? selectedSpecId)) ?? null}
-            isDisabled={specsLoading}
-            isLoading={specsLoading}
-            onChange={handleSpecializationChange}
-            className="ppc-combobox"
-            classNamePrefix="ppc-combobox"
-            placeholder="Select a specialization..."
-            theme={(t) => comboboxTheme(t, isDark)}
+            value={specOptions.find((s) => s.value.id === selectedSpecIdOrSelected) ?? null}
```
fiveminus1:
```suggestion
           value={specOptions.find((s) => s.value.id === (majorWithSpec.selectedSpec?.id ?? selectedSpecId)) ?? null}
```

site/src/app/roadmap/catalog/MajorCourseList.tsx
fiveminus1: Getting `Encountered two children with the same key, `Exercise Sciences, B.S.`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.` console error from this component. Should address this

fiveminus1: Additionally, it seems like this component is kind of scuffed and is using both state and Redux as sources of truth for specializations? I think that's OOS to fix in this PR but @CadenLee2 want to take a look to make sure I'm looking at it right?

site/src/component/ReviewForm/ReviewForm.tsx
```diff
@@ -431,26 +428,28 @@ const ReviewForm: FC<ReviewFormProps> = ({
...
-              theme={(t) => comboboxTheme(t, darkMode)}
-              className="tag-select"
-              classNamePrefix="tag-select"
+              open={selectedTags.length < 3 && tagsOpen}
+              onOpen={() => setTagsOpen(true)}
+              onClose={() => setTagsOpen(false)}
+              getOptionLabel={(option) => option}
+              getOptionDisabled={() => selectedTags.length >= 3}
+              disableCloseOnSelect
+              disableClearable
```
fiveminus1:
```suggestion

```
This should be clearable

site/src/component/ReviewForm/ReviewForm.tsx
```diff
@@ -431,26 +428,28 @@ const ReviewForm: FC<ReviewFormProps> = ({
...
-              menuIsOpen={selectedTags.length < 3 && tagsOpen}
-              isOptionDisabled={() => selectedTags.length >= 3}
-              placeholder="Select up to 3 tags"
-              closeMenuOnSelect={false}
-              theme={(t) => comboboxTheme(t, darkMode)}
-              className="tag-select"
-              classNamePrefix="tag-select"
+              open={selectedTags.length < 3 && tagsOpen}
+              onOpen={() => setTagsOpen(true)}
+              onClose={() => setTagsOpen(false)}
```
fiveminus1:
```suggestion

```
You should? be able to eliminate the `tagsOpen` state if you don't want to keep the behavior of closing the menu when you hit 3 tags (which I personally think is better anyways).

site/src/component/ReviewForm/ReviewForm.tsx
```diff
@@ -431,26 +428,28 @@ const ReviewForm: FC<ReviewFormProps> = ({
...
+              getOptionLabel={(option) => option}
+              getOptionDisabled={() => selectedTags.length >= 3}
```
fiveminus1: This is unnecessary too

site/src/style/theme.ts
```diff
@@ -261,6 +264,45 @@ theme = createTheme(theme, {
...
+            fontSize: 16,
+            margin: '0 4px 0 -4px',
+          },
+          '& .MuiAutocomplete-endAdornment .MuiSvgIcon-root': {
+            color: 'var(--mui-palette-text-primary)',
+          },
+          '& .MuiOutlinedInput-root input::placeholder': {
+            color: 'var(--mui-palette-text-secondary)',
+            opacity: 1,
+          },
```
fiveminus1: Also don't think these are necessary?

fiveminus1 (changes requested):
Transfer Credits:
* Course search: Works well, but see code comments
* AP selection: Also just code comments

Catalog:
* Major/Minor Select: **"Select ..." placeholder text is persisting after selection** 

Review Form:
* **Tag select** background is too dark in dark mode 
* **Select up to 3 tags** persists on selection
* Should be allowed to clear tags
* Couple code things

A lot of the theme changes appear unnecessary as well to me, but see comments and let me know if I'm missing anything on those.

fiveminus1 (changes requested):
Code looks much better; just one or two remaining comments on that.


Aside from that, some final visual comments:
Image

It's not great that the "Score" label gets cut off..you might want to play around with this. If there isn't some way to make the cutoff threshold larger via MUI, probably reduce the size of the left one by a bit. Both should be displayed

Image

Image

These placeholders shouldn't display after selecting something


Almost good to go!

site/src/app/roadmap/catalog/MajorCourseList.tsx
CadenLee2: I sort of see what you're saying since it's a bit hard to follow, but I don't see any overlap between state and redux upon first glance. Definitely oos, but can look more closely later

CadenLee2 (commented):
One small visual concern: when a long enough major is chosen, it takes up the whole line and the textbox appears on the next line
Image
Image

Not sure of the best way to handle this: one possible solution is reducing the max width of the tags (inspect and select a tag to see it's `100% - 6px`, we can increase the 6) and also reducing the min width of the `input` so that they both fit on the same line.

Perhaps when the `input` is active, only then does it have a larger min width and thus go to the next line. Lmk on Discord if you have questions or ideas

fiveminus1 (commented):
I feel like it looks weird now that this doesn't take up the whole line 😅 

Image

fiveminus1 (changes requested):
Courses Taken search does not appear to display the same results as on prod. Try the query ICS 9, and no results can be found. If you search the full "I&C SCI 9," this will appear. More broadly, try ICS, and not as many results will appear. I'm pretty sure this is specific to this PR, as I've tried reverting to main branch locally and the results are fine. 

If this was related to the debouncing issue and you really cannot find a solution, I would be fine with you reverting any debouncing attempted fixes and just keeping the component refactor for the sake of functionality and time merging this. Just open a separate issue after.

I am still not a big fan of the solution we've gone with on the tags. See the image below in particular. Probably to be discussed on Discord...

Image

Tested and compared to prod: 
* AP Select: ✅ 
* Major/Minor Select ✅ 
* Specializations select ✅ 
* Review Form - looks kinda scuffed showing so many options and above the dropdown but honestly it'll be updated soon and I don't think it's worth the extra time trying to fix this ☑️ 

Rest of code seems fine, just need to fix up courses section mostly

DerrickThrower:
I need it

site/src/style/theme.ts
```diff
@@ -302,6 +302,28 @@ theme = createTheme(theme, {
...
+              minWidth: 0,
+            },
+          },
+          '& .MuiChip-root': {
+            height: 24,
+            fontSize: 14,
+            borderRadius: 2,
+            maxWidth: 'calc(100% - 93px)',
+          },
+          '& .MuiChip-deleteIcon': {
```
fiveminus1: This isn't applying at the moment due to specificity--you can try `& .MuiChip-root .MuiChip-deleteIcon` if you do want to keep this

fiveminus1 (approved):
LGTM at last 🤑 

All components tested, no further instances of `react-select` remaining, previous nits seem fine to me. 

One final comment about a theme styling, but you can fix it up and merge. Unsure if @CadenLee2 wants a final review too

CadenLee2 (approved):
```
L  G  T  M
O  O  O  U
O  O     I
K  D
S
```

Also look at jerb suggestion above ^

PR #971, opened 2026-02-04
Link major and minor to user instead of roadmap

### LINKED ISSUE #951: Remove broken planner/major association
Context: currently, PeterPortal attempts to tie majors and minors to specific roadmaps, such that if you switch between roadmaps it works. However, due to the nature of local storage and planner IDs, the approach is buggy and adds complexity to both the developer and user experiences.
1. Majors/minors do not save when logged out. Try changing them then logging in; they do not save. This is expected but not optimal.
2. Users can get confused when their major changes after reloading, doesn't save properly, etc. as it may not fit into their conceptual model
3. The logic of tying majors to roadmaps complicates the codebase

Image

## Replicating issues
1. When logged in, clear all majors so you have none for your first roadmap. Save and reload. You should have no majors.
2. Add a roadmap, then save. This roadmap also should have no majors.
3. Go to your second roadmap, and add the Aerospace Engineering major. Go to your first roadmap, and note that the *major still persists there* (which should not happen).
4. Save and reload. The major is now gone from the first one.
5. Switch to the second one then back to the first and you will see the same issue as in step 3. Then, add Anthropology to your first roadmap. Save and reload. You will notice that Anthropology is now included for your first and second roadmaps.

## Intended behavior
Majors should not change when switching roadmaps. Majors should also save locally when you're logged out, though this is out of scope and will be a future issue.

## Implementation suggestions
1. Update the database Zod schema (`api/src/db/schema.ts`)
    - Add `userMajor` and `userMinor` tables. They should have a similar structure as other tables that reference the user. The relation is a "one to many" because a user can have none (or up to many) majors. Note that unlike `plannerMajor` and `plannerMinor`, these tables should make the pairs `userId, majorId` and `userId, minorId` unique because it does not make sense for a user to have two of the same major
    - Do not get rid of the original tables yet, as this would cause data loss and other issues. See step 3 below for how this will be handled
    - Update backend logic to use these tables, and update the endpoint(s) so as to not need the planner information when a user changes their major/minor
    - (Reach out to the lead for support for dev database migration if necessary)
2. Remove all frontend logic for multiple majors
    - The frontend shouldn't try to refresh majors when changing roadmaps, and should simply call the API properly
4. Create a *migration script* so that existing users' majors are properly ported over to the new tables
    - Reference other migration scripts (in `api/scripts`) when making this script
    - For each user, the major(s) and minor(s) of their *first roadmap*, from `plannerMajor` and `plannerMinor`, will be inserted into the `userMajor` and `userMinor` tables. On conflict (i.e. breaking the uniqueness constraint defined in step 1) it should do nothing/continue without inserting, so that the script can be idempotent

## Future issues (out of scope of this issue)
- Save majors locally when logged out
- After ensuring the script is run on prod and the change is merged, remove the `plannerMajor` and `plannerMinor` tables, which should no longer be referenced.

CLiang531:
Major and minor information is now tied to the user #951

## Description

Updated database schema to create new tables and changed backend logic to reference these tables.
- added new tables `userMajor` and `userMinor` so that major and minor information is tied to the user's `userId` instead of one of their planners
- created a migration script so that existing information from `plannerMajor` and `plannerMinor` can be moved to these new tables, and in case users had multiple majors and minors from different roadmaps, the script chooses the user's first planner (has not been migrated yet)
- removed planner information and references to the old tables from backend logic
## Test Plan

1. After migration, check that `userMajor` and `userMinor` now includes all major and minor information from their **first** roadmap only
2. Log in and check that the user major and minor information displays correctly
3. Create a new roadmap (or navigate to a different roadmap) and edit major and minor information. Switch back to original roadmap and see if major or minor information changes. 
4. Also check that when a user switches roadmaps major and minor information don't refresh or attempt to reload

## Issues

Closes #951


site/src/app/roadmap/catalog/MajorSelector.tsx
```diff
@@ -1,4 +1,4 @@
-import { FC, useCallback, useContext, useEffect, useState, useRef } from 'react';
+import { FC, useCallback, useContext, useEffect, useState /* useRef */ } from 'react';
 import Select from 'react-select';
```
CadenLee2: Looks like this comment can be removed

PR #976, opened 2026-02-06
Update newly-created planner temp IDs to match IDs in database

charlieweinberger:
## Description

Currently, when a new roadmap plan is created, it has a temporary negative ID in the frontend (-1, -2, etc.). When the user saves this roadmap for the first time, it's assigned a proper id (e.g. 741) in the database, but the redux state is never updated. This causes problems if the user then proceeds to immediately (i.e. without refreshing the page) update or re-save their newly-created roadmap, since the frontend still tries to send the temporary negative ID to the database, but the database doesn't know what it's referring to because it assigned an actual ID.

In this PR, when the database creates a proper ID for the new roadmaps, a lookup table is returned to the frontend, containing a map from the temporary indices to the proper indices. (This table already existed, but wasn't being returned to the frontend). Now, the redux state (and local storage) can be immediately updated to store the correct ID instead of the temporary one, so the user can then edit or re-save the new roadmap without issue.

## Screenshots

N/A (backend fix)

## Test Plan

Create a new roadmap, and then save it, and then immediately (i.e. without refreshing the page) edit the roadmap (e.g. add a new course), and then save the roadmap again. You should not get an error, and when you refresh the page, your change (e.g. new course) should still be there.

- [ ] Test the above scenario when creating a new roadmap from scratch
- [ ] Test the above scenario when duplicating a roadmap
- [ ] Test the above scenario with multiple roadmaps at once
- [ ] Test the above scenario with creating a roadmap and then immediately deleting it (instead of adding a new course).

## Issues

Fixes a bug discovered by @anthonyj33 in discord.

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -12,17 +12,24 @@ on:
 
 # do not cancel in progress, SST will be stuck in a "locked" state if cancelled mid-deployment
 concurrency:
-  group: staging-${{ github.event.pull_request.number }}
+  group: staging-${{ github.event.pull_request.number || github.event.client_payload.pull_request.number }}
 
 jobs:
   build_and_deploy:
     name: Build and deploy
     runs-on: ubuntu-latest
+    env:
+      PR_NUM: ${{ github.event.pull_request.number || github.event.client_payload.pull_request.number }}
+      PR_SHA: ${{ github.event.pull_request.head.sha || github.event.client_payload.pull_request.head.sha }}
+      FULL_NAME: ${{ github.event.pull_request.head.repo.full_name || github.event.client_payload.pull_request.head.repo.full_name }}
     # don't run if labeled "no deploy" && don't run on PRs from forks
-    if: (!contains(github.event.pull_request.labels.*.name, 'no deploy')) && github.event.pull_request.head.repo.full_name == github.repository && github.event.pull_request.head.ref != 'branding-changes'
+    if: |
+      !contains(github.event.pull_request.labels.*.name || github.event.client_payload.pull_request.labels.*.name, 'no deploy') &&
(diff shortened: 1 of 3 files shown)


site/src/hooks/planner.ts
```diff
@@ -64,20 +64,26 @@ export function useSaveRoadmap() {
     const collapsedPrevious = collapseAllPlanners(lastSavedRoadmapPlans);
     const collapsedCurrent = collapseAllPlanners(planners);
 
-    const res = await saveRoadmap(isLoggedIn, collapsedPrevious, collapsedCurrent);
-    if (res && isLoggedIn) {
+    const result = await saveRoadmap(isLoggedIn, collapsedPrevious, collapsedCurrent);
+
+    if (result.success && isLoggedIn) {
       dispatch(setToastMsg('Roadmap saved to your account!'));
       dispatch(setToastSeverity('success'));
```
CadenLee2: Noticing a fair bit of duplicate code here and `PlannerLoader.tsx` for the save step, we probably want to refactor at some point so that changes like this only have to be made in one place

site/src/store/slices/roadmapSlice.ts
```diff
@@ -178,6 +178,15 @@ export const roadmapSlice = createSlice({
         plan.chc = action.payload.chc;
       }
     },
+
+    // Update the planner IDs of newly created planners that still have temporary (negative) IDs
+
+    updateTempPlannerIds: (state, action: PayloadAction<Record<number, number>>) => {
+      if (Object.keys(action.payload).length === 0) return;
```
CadenLee2: I would group this directly underneath the "modifying the roadmap" items and change the comment to just a JSDoc on the line above

anthonyj33:
LGTM

PR #980, opened 2026-02-11
Undo/redo

DerrickThrower:
added undo/redo har har har


site/src/app/roadmap/toolbar/Header.tsx
```diff
@@ -45,6 +55,12 @@ const Header: FC<HeaderProps> = ({ courseCount, unitCount }) => {
       </div>
       <div className="planner-actions">
         <ButtonGroup>
+          <Button variant="text" size="small" className="header-btn icon-only-btn" onClick={handleUndo}>
+            <UndoIcon />
+          </Button>
```
CadenLee2: I see the styling and sizing looks a bit funny given how small these buttons are. How about using a MUI IconButton like in the multiplan menu? It would probably look similar when not clicked, but the sizing and hover/clicked states will look more consistent

Image

CadenLee2 (commented):
1. Don't forget to update the buttons so that they're disabled when you can no longer undo or redo
2. See my comment about the component type

site/src/app/roadmap/toolbar/Header.tsx
```diff
@@ -20,6 +23,15 @@ const Header: FC<HeaderProps> = ({ courseCount, unitCount }) => {
...
   const dispatch = useAppDispatch();
+  const currentRevisionIndex = useAppSelector((state) => state.roadmap.currentRevisionIndex);
+  const revisions = useAppSelector((state) => state.roadmap.revisions);
+
+  const handleUndo = () => {
+    dispatch(undoRoadmapRevision());
+  };
+  const handleRedo = () => {
+    dispatch(redoRoadmapRevision());
+  };
```
CadenLee2: Error when undoing planner addition or redoing planner deletion

Steps to recreate issue:
1. Add a planner (and make sure that planner is selected)
2. Undo

You'll have to edit the roadmap revision functions to make sure the currently selected plan index is still in range

CadenLee2 (commented):
Functionally this works great. I have one small visual concern: because the button group's gap was changed, the gap between the 2 action buttons on mobile is uncomfortably bigger now.

Also, the undo/redo buttons are spaced farther apart from each other than they are from the other buttons, but according to gestalt principles related items should be closer together.
Image

Ideally the spacing should look like this:
Image

I've left comments for what should change

PR #983, opened 2026-02-12
Save previous roadmap

Ascent817:
## Description

Re-implemented most of Charlie's old pr #722

Still some bugs to iron out and I wasn't able to figure out how to access the router from the roadmap slice to save the current plan index on switch, so it currently only saves on roadmap save.

### DIFF
--- a/api/drizzle/0024_hesitant_wong.sql
+++ b/api/drizzle/0024_hesitant_wong.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "user" ADD COLUMN "current_plan_index" integer DEFAULT 0 NOT NULL;
--- a/api/drizzle/meta/0024_snapshot.json
+++ b/api/drizzle/meta/0024_snapshot.json
@@ -1,0 +1,1314 @@
+{
+  "id": "0f55d110-2818-4791-92ad-b295961d1bc6",
+  "prevId": "4285a98d-9508-4ec7-ab3c-ac54c8192aec",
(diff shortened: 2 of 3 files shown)


api/src/controllers/roadmap.ts
```diff
@@ -17,17 +24,19 @@ import {
...
-      if ('plannerId' in change)
-        return [change.plannerId]; // All non-planner operations
-      else if ('data' in change)
-        return [change.data.id]; // Planner update
-      else return [change.id]; // Planner delete
-    });
+  const plannerIds = Object.entries(input)
+    .filter(([key]) => !['overwrite', 'currentPlanIndex'].includes(key))
+    .flatMap(([, value]) => (Array.isArray(value) ? (value as RoadmapChange[]) : []))
+    .map(extractPlannerId);
```
CadenLee2: This section looks a bit convoluted, there may be a way to improve it.

One detail is that after the filter, all the values should be arrays, right? This may mean that the `Array.isArray` check is unnecessary and we can just use `value as RoadmapChange[]`. You might want to look closer and test a bit to make sure

charlieweinberger (approved):
functionality LGTM, code looks good other than the comment Caden has above. nice work finally finishing this issue!

api/src/controllers/roadmap.ts
Ascent817: Yep this was unneeded, I got rid of it.

PR #984, opened 2026-02-13
fix: normalize enrollment percent so as to prevent progress bar overflow

### LINKED ISSUE #967: Overenrollment overflows progress bar
**Describe the bug**
If a course's enrollment exceeds its maximum capacity, the progress bar appears to "overflow" and some gray will appear at the start of the bar. This is because MUI's [LinearProgress](https://mui.com/material-ui/react-progress/) expects a progress value in the range of 0-100, and the current calculation for progress does not normalize for this. An example of how to handle this is included on that doc page.

**To Reproduce**
2. [actions taken]

**Expected Behavior**
Should probably just stay fully red 

**Actual Behavior**
Gray at the start
Image

**Additional context**
N/A

jenniyan:
Fix progress bar overflow for full courses

## Description

In `site/src/component/Schedule/Schedule.tsx`, added a `normalize` function to clamp enrollment percentages for the LinearProgress bar.  
This ensures that values above 100 are capped at 100 and values below 0 are set to 0, preventing the progress bar from overflowing or showing unexpected gray segments.

## Screenshots

Before:
IMG_2016

After:
IMG_3516

## Issues

Fixes #967

### DIFF
--- a/site/src/app/not-found.tsx
+++ b/site/src/app/not-found.tsx
@@ -1,0 +1,5 @@
+import Error from '../component/Error/Error';
+
+export default function NotFound() {
+  return <Error />;
+}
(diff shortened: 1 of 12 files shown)


CadenLee2 (commented):
Mostly looks good, though I have a code suggestion

PR #994, opened 2026-02-19
Standalone page links

### LINKED ISSUE #993: Broken links on standalone pages
Steps to replicate:
1. Go to MATH 3A in a preview on the roadmap page. All links should work
2. Go to https://antalmanac.com/planner/course/MATH3A. Some links will do nothing when you try to click on them
  - Instructors on the left side
  - Names of instructors in reviews
  - Likely some other links as well

This is due to the fact that links are meant to open up previews, but the preview doesn't exist on standalone pages (course or professor pages).

To fix this, use the following rules when clicking on a link in these pages or in reviews:
- If a preview is currently open, then just open the preview (as it currently does) and prevent the link navigation
- If there is *no preview currently open*, then navigate to the page (i.e. don't prevent the link navigation)

MajdKatabi:
## Description

I did a check if the pathname is solely /planner, if so that means it is a preview so continue to default actions, but if the pathname is longer than /planner that means it is a standalone page (ex: /planner/course/MATH3A) and therefore it returns early to not push to router or dispatch.

This pull request is also pulling from Fix Preview Navigation which ensures there is a router history so the back arrows work between previews.

## Test Plan

Go to: http://localhost:3000/planner/course/MATH3A

Check the links on the page work and go to standalone pages

## Issues

Closes #993


CadenLee2 (commented):
Functionally this works well when testing, and codewise seems a lot more robust than what existed before. I do have a few code comments.

(Also since I know you're away right now, no worries about waiting to finish this)

CadenLee2 (approved):
LGTM! Everything I tested worked as expected

PR #995, opened 2026-02-19
Completion Hint for (Grouped)CourseRequirement

anthonyj33:
## Description
Clarifies number of required courses or units for (Grouped)CourseRequirement

## Screenshots
See comments below (old screenshot outdated)

## Test Plan
- Verify that all instances of ambiguous requirements are fixed
- Verify that this doesn't show up anywhere unnecesary

### DIFF
--- a/api/src/helpers/roadmap.ts
+++ b/api/src/helpers/roadmap.ts
@@ -110,7 +110,7 @@ export async function setQuarterCourses(tx: TransactionType, quarters: PlannerQu
       courseId,
       index,
     }));
-    await tx.insert(plannerCourse).values(rows);
+    await tx.insert(plannerCourse).values(rows).onConflictDoNothing();
   });
   await Promise.all(updates);
 }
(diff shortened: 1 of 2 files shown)


fiveminus1 (commented):
Looks good for the stated functionality, with a minor opinion on the code. 

However, maybe I'm just not looped into this but why doesn't the counter also apply to CourseRequirements like GE 2? It just seems weird to me (and to a user standpoint) if they see a counter for only some requirements since to them, there shouldn't be a discernible difference between GE requirements. 

(EDIT: I do realize I'm looking at this from the standpoint of GEs when I presume the overall PR is meant for all requirements, but I still do think there should be a counter in general unless I'm misunderstanding something)

CadenLee2:
> why doesn't the counter also apply to CourseRequirements like GE 2?

The counter was originally to resolve ambiguities (i.e. make it clear how many you need to take) and that's not an issue in places like GE 2. However, given that the counter also helps users tell how many more courses they need to choose, it's probably useful to have in more places.

Let's see what it looks like with this addition

anthonyj33:
image
image

PR #1002, opened 2026-02-25
Fixed mobile switcher hover color 

### LINKED ISSUE #975: Incorrect mobile switcher hover color
When hovering the AntAlmanac logo switcher on mobile, the background color is too dark:
Image
It should instead match the hover color of other buttons (faintly light):
Image

The fix should be done naturally if possible, without explicit overrides; see how the topbar buttons do it.

DerrickThrower:
## Description

Normalized the switcher hover color on the logo with the other topbar buttons 

## Screenshots
Screenshot 2026-02-24 at 10 45 22 PM

## Test Plan

check the color is the same

## Issues

Closes #975


CadenLee2 (approved):
LGTM 💵💵

Image

PR #1009, opened 2026-02-28
Add data disclaimer

### LINKED ISSUE #965: Data accuracy disclaimer
Originally suggested by Kevin, but I think some sort of disclaimer that our data/app is not perfect and to use official UCI sources to double-check things like degree requirements would be in order. We currently still have various one-off issues relating to degree req accuracy such as #744 that still need to be ironed out.  

No

Haven't quite workshopped the wording yet but some form of disclaimer, likely in the footer. Shouldn't be too intrusive, but all users should see it upon first use of the app.

Could be in the empty catalog space before adding any majors/minors.

**Additional context**
N/A

fiveminus1:
## Description

Added a brief disclaimer in the footer regarding data accuracy. Wording may not be final, open to comment

## Screenshots

image

## Test Plan

- [ ] Looks good in desktop/mobile
- [ ] Link works

## Issues

Closes #965


CadenLee2 (commented):
Lgtm except recommend some inline padding so we don't get this on awkward-sized screens
| | |
| --- | --- |
| Image | Image |

PR #1026, opened 2026-03-06
Add support for variable units

### LINKED ISSUE #353: Support for courses with variable units

Ascent817:
## Description

* Updated backend types to support new `userChosenUnits` field
* Ran a database migration to add variable units column
* Added input box and edit icon to each course in roadmap to allow variable unit edits

## Screenshots
image

### Box Behaviors
- verify input box behavior
  - enter/click check: save edits
- verify that the input box rejects wrong inputs
  - test strings, values outside of min/max range
- verify value is on course and calculated correctly in quarter, year, and planner

### Saving Behaviors
- verify the variable unit is kept on save
- verify variable unit is kept on remount (ie move course with variable unit to diff quarter/year)
- check loading from local
  - be logged out
  - drag in a course w/ variable units (CS 199)
  - specify units (or don't)
  - save and log in to an account
  - load from local and check that the variable units match with what was specified before
- check loading from account

## Issues

Closes #353


site/src/app/roadmap/planner/Course.tsx
```diff
@@ -118,8 +172,46 @@ const Course: FC<CourseProps> = (props) => {
...
+            isInRoadmap && minUnits !== maxUnits && (
+              <div className="custom-units">
+                {editUnitOpened ? (
+                  <>
+                    <ClickAwayListener onClickAway={handleClickAway}>
+                      <OutlinedInput
+                        className="unit-input"
+                        placeholder={defaultUnitsText}
+                        value={inputUnit}
+                        onChange={handleVariableUnitChange}
```
CadenLee2: In terms of behavior, I think this works well, but will need to test some more.

In terms of code, we should move this (and related logic, like the functions above) to a separate component. It can be in the same file or another file next to this one.

Example of how we might be using it here:
```tsx
          {isInRoadmap && minUnits === maxUnits && <span className="units">{defaultUnitsText}</span>}
          {isInRoadmap && minUnits !== maxUnits && <CustomUnits ...some props here... />}
```

When refactoring, consider what things should be props and what belongs inside/outside of the `<CustomUnits />` component

site/src/types/roadmap.ts
```diff
@@ -40,7 +41,19 @@ export interface PlannerQuarterEdit {
   after: PlannerQuarterChangeData;
 }
 
-export type RoadmapEdit = PlannerEdit | PlannerYearEdit | PlannerQuarterEdit;
+// use this to change variable course units?
```
CadenLee2: Remaining code comment

site/src/app/roadmap/planner/Course.scss
```diff
@@ -163,3 +164,22 @@
     opacity: 1;
   }
 }
+
+// @todo adjust spacing + class names + overall looks
+.custom-units {
```
CadenLee2: Is this todo still a concern? I think it looks solid currently

atile4: When zoomed out, when the input box is opened the course title is pushed down a tiny bit. For my screen I have to zoom out to 90% and 75% to see it. It's a really small difference, so I wasn't sure if it was a big enough issue to fix. 
I tried getting a screenshot of the difference but the difference was too small to see through screenshots.

site/src/helpers/planner.ts
```diff
@@ -240,13 +256,37 @@ function addIdsToLocalRoadmap(roadmap: SavedRoadmap): SavedRoadmap {
   return roadmap;
 }
 
+// Supporting Variable Units
```
CadenLee2: This feels like a useless comment, though I see why it was added to match the other comments.

It could be more useful if it said `// Changes courses in quarters from strings to objects`

types/src/roadmap.ts
```diff
@@ -48,13 +56,41 @@ export type ExtendedTransferData = z.infer<typeof extendedTransferData>;
...
+  startYear: z.number(),
+  name: z.string().max(35),
+  quarters: z.array(legacySavedPlannerQuarterData),
+});
+
+const legacySavedPlannerData = z.object({
+  id: z.number(),
+  name: z.string().max(35),
+  content: z.array(legacySavedPlannerYearData),
+  chc: z.enum(['', 'CHC4', 'CHC2']).optional(),
```
CadenLee2: Realize now that this might not be ideal since it's repeated between the legacy and "modern" types. We should refactor to `const chcVariant: z.enum( ...` and use that constant here + in savedPlannerData

PR #1027, opened 2026-03-07
Custom cards frontend

### LINKED ISSUE #986: Custom card menu frontend
The custom card menu will be integrated into the "saved" tab in the sidebar catalog. This tab will be renamed to "library" and will include both saved courses *and* custom courses (called "custom cards").

This PR should be based on the `custom-cards` branch.

## Requirements
- Rename the "Saved" tab to "Library"
- Add section headers much like how Credits is split into "Courses Taken", "AP Exam Credits", etc.
  - These headers should be "Saved" (above saved courses) and "Custom Cards" (above custom cards)
- Add the custom cards section as per the designs
  - Should look very similar to courses in the roadmap
  - The custom cards should be draggable into the roadmap just like other courses, though it is not expected that it will work properly with saving roadmaps. Try to reuse code when possible when it comes to dragging/structuring the components

To specify the behavior of the custom cards menu:
- Clicking the "plus" adds a blank custom card
- Clicking one of the input boxes focuses it and gives it the outline (as seen in the designs). Use MUI for this
- The trash can deletes the card

The menu should depend on **redux state** so that the state can be easily accessed in the future. Consider making a `customCardSlice`

## Designs
These are not fully accurate (due to the "library" tab rename). Reference this for the course card specifically, and check the Winter 26 Figma for further details.
Image

ellisl10:
Custom Card Frontend

## Description

1. Renamed the saved tab to "Library"
2. Added collapsible section headers
3. Created a new custom course card section

## Screenshots

Screenshot 2026-03-08 at 9 25 56 PM

## Test Plan

- [ ] Adding a card should add a blank custom card
- [ ] Should be able to edit the name, description, and units
- [ ] Cards should be draggable into the roadmap
- [ ] Changes to custom cards in the library should be reflected in the roadmap

## Issues

Closes #986


CadenLee2 (commented):
This is good progress! Since this is a large PR, I've got a lot of feedback: I'll start with behavior-related things in this review, then after some fixes, can look more closely at the code.

## Behavior/style
Image

- The "No saved courses!" header is unexpected when the Saved already is there. I think we don't need that header in the section anymore, now that Saved is collapsible and we have the explanatory text.
- I think custom cards (and the + button) need some space between each other (maybe gap of `8px`, like search results)
- Border width/color seems wrong: they maybe only need a `1px` border, or maybe even none and just keep the drop shadow, to match search results on prod

---

Image

- Search results no longer look correct because of styling changes. The appearance of these shouldn't have changed

---

Image

- To match existing courses, the cards should take up the full width
- Also, on smallish screens, the drag handles for normal courses now wrap incorrectly

---

Unless I'm mistaken, adding/managing custom course cards does not work on mobile. This is acceptable for now since it's a niche usecase and could be complex UI/UX-wise, but important to note.

## Code
I think most of the above problems can be handled cleanly by changing a how the code is structured. One important bug is in `CustomCourseCard.scss` (I left a more specific comment), though in changing that some other things may also need to be fixed.

charlieweinberger (commented):
In addition to what Caden suggested above, I've got a few comments to add.

---

When I created a custom course card, I was confused at how saving a course worked. Since the text input fields are always shown, it's not clear that the course is actually saved IMO. I'm not sure what the UI solution is here - maybe there should be an edit/save button like the courses in "courses taken" in the transfer credits menu?

Image

---

I see `CourseGQLData | CustomCourse` used as a union type several times in the codebase, can this be defined once as its own type in `types.ts` (maybe named `CourseType`) and then imported wherever used? And maybe also `(CourseGQLData | CustomCourse)[]` (maybe named `CourseListType`)?

---

Similarly, I see the code below used in `Planner.tsx` (this one also checks for `NaN`), `Year.tsx`, and `Quarter.tsx`. Can this be defined as a helper function once (maybe named `getCourseUnitCount`), and then imported wherever used?

```
if ('courseName' in course) {
  unitCount += course.units;
} else {
  unitCount += course.minUnits;
}
```

site/src/app/roadmap/catalog/CustomCourseCard.scss
```diff
@@ -56,9 +61,14 @@
     min-width: 0;
   }
 
+  .name-input {
+    color: white !important;
+  }
```
CadenLee2: We can get rid of all the `!important` in this file by:
- Get rid of the `color: inherit !important;` in `*` at the bottom of the file
- To make sure the trash can icon is still blue, add `color: inherit;` to `button.MuiIconButton-root`
- Now you can remove the `!important` on the `color: white`

However, there is also a problem where these are unreadable in light mode. Since the `*` `!important` is no longer here, we can just remove all the `color: white` altogether and let the text boxes take the default light or dark text color, which works well.

Image

CadenLee2 (commented):
One quick comment on css/colors, otherwise everything works well

CadenLee2 (approved):
LGTM 💵

Issue #1004, opened 2026-02-25
Overriding requirements backend


CadenLee2:
See #997 for the context, and for details about the frontend that'll be interfacing with this

## Tasks
- Add the table to our db schema
  - Since overrides are per-roadmap, this table needs to store the user id, the roadmap, and the requirement identifier (requirements will be identified as a string). It should be unique on (user, roadmap, requirement)
  - Should probably be indexed by (user, roadmap) so that we can load the overrides in the frontend for each roadmap
- Migrate the db (**coordinate with lead to run the migration**)
- Add endpoints for users to manage their overrides
  - Add override: takes in roadmap & requirement identifier, uses user id from context
  - Delete override: takes in roadmap & requirement identifier, uses user id from context'
  - [EDIT: added this one] Get overrides: given a plannerId, and the user id from context, get a list of all of the requirements that have been overridden

PR #1031, opened 2026-03-12
Override Requirements Frontend

### LINKED ISSUE #997: Overriding requirements frontend
Some users want to override requirements that are incorrect (i.e. due to incorrect API data). The manual override feature will allow force completion.

It will require both backend changes (storing the user's saved overriden requirements) and frontend changes (allowing users to interact and displaying the completion).

This issue is specifically for the frontend.

## Tasks
- Add a checkbox to all **expanded** requirements (see mockup below)
  - When an unchecked box is clicked, it should show a confirmation modal titled something like "Confirm force completion", with a message including the name of the roadmap like "Are you sure you want to force complete this requirement for **Peter's Roadmap**? You should only do this if you are sure the courses you've taken satisfy it."
  - Confirming should update the slice (see below)
  - When a checked box is clicked it goes away
- In a slice, store which requirements the user has overridden
- In the requirements lists, if the requirement (identified by the `storeKey`, see `ProgramRequirementsList.tsx`) is overridden, it should visibly be completed

Image

CLiang531:
## Description

- Stored overridden requirements in slice using storeKey as identification
- Added checkbox to requirement headers
- When unchecked checkbox is clicked, open a confirmation modal and if confirmed, show visual completion styling and store in slice
- When checked checkbox is clicked, clear styling and checkbox

## Screenshots

image
image
image

## Test Plan

1. Navigate to major, minor, or GE tab and open any of the requirements. Verify that checkbox is present and check it. A modal should pop up asking for confirmation. Choosing "Force Complete" should apply visual styling, and "Cancel" should do nothing.
2. Navigate away from the tab and come back to verify that overrides are being stored in slice. 
3. Find an overridden requirement (or use the one from step 1) and uncheck it. No modal should pop up, and visual styling should clear.

## Issues

Closes #997


CadenLee2 (commented):
This looks solid and I found almost no issues while testing! Only things:
1. When the modal is open, clicking outside (rather than on cancel/confirm) closes the requirement group for some reason
2. Some code suggestions

Note that this shouldn't be merged to main yet because overrides aren't saved to the user's account yet. We'll be waiting on #1004 to get done first

CadenLee2 (commented):
Again, we can't merge to main yet, but otherwise the changes lgtm! 💵

anthonyj33 (changes requested):
The code LGTM as well, but I noticed something small on staging

After expanding a requirement block, the checkbox appears and its click area slightly overlaps the click area for the toggle to collapse the block. I noticed this because I clicked to expand a block and then clicked in the same spot to close it, but ended up clicking the checkbox instead.

One possible fix for this is dealing with z-index but there might be a nicer way.

DerrickThrower (approved):
Was playing around with it to try to see if anything breaks but I couldn't. LGTM 💸

CadenLee2 (approved):
Above comments are addressed, everything still works well for me!

anthonyj33 (approved):
LGTM! Very cool, will be great to get this in prod for all the ppl requesting it :D

PR #1032, opened 2026-03-12
Custom course cards saving/loading

CoreyZhong:
## Issues

Closes #


CadenLee2 (commented):
Reminder of what we discussed in the meeting:
- The save button shouldn't be always disabled when logged out. We may want to disable the custom card menu entirely when logged out, though this can be discussed and can be a future PR to the custom cards branch
- We should allow empty custom card titles to simplify the frontend logic

site/src/app/roadmap/catalog/CustomCourseCard.tsx
```diff
@@ -0,0 +1,132 @@
...
+            <input
+              className="name-input"
+              value={newName ?? ''}
+              placeholder="Course"
+              onChange={(e) => setNewName(e.target.value)}
+              onKeyDown={handleKeyDown}
+              onBlur={onBlur}
+            />
+          ) : (
+            <>{course.courseName}</>
```
CadenLee2: Can just do `course.courseName` instead of wrapping with the `<>{`

CadenLee2 (commented):
Functionally this seems to work, but there's an issue with types (see longer comment below) that will need to be fixed. Essentially the types should reflect the real data that'll exist there (i.e. each expanded quarter could include `CourseGQLData` or `CustomCourse`).

site/src/store/slices/roadmapSlice.ts
```diff
@@ -130,8 +140,49 @@ export const roadmapSlice = createSlice({
...
+      const customCourseId = action.payload;
+
+      state.plans.forEach((plan) => {
+        plan.content.yearPlans.forEach((year) => {
+          year.quarters.forEach((quarter) => {
+            quarter.courses = quarter.courses.filter((course) => {
+              if (isCustomCourse(course)) {
+                return course.id !== customCourseId;
+              }
+              return true;
```
CadenLee2: This can be simplified to
```ts
.filter((course) => !(isCustomCourse(course) && course.id === customCourseId))
```

CadenLee2 (approved):
I have a few small code suggestions (not just the comment below, which is one example), but because a lot will possibly change when we merge, and we're on a close deadline, I believe we should handle those concerns in `custom-cards` once the frontend gets merged and we put up a PR for it, so I'll bring them up then.

Testing-wise, everything works as expected:
- Saving and loading
- Deleting custom cards that are in a planner
- Instantiating custom cards into planners
- Logged out message

PR #1052, opened 2026-04-06
Share URL Backend Set Up

CLiang531:
## Description

- Added `paralleldrive/cuid2` library to autogenerate a `share_id` when a new planner is saved for the first time
- Added an endpoint in `api/src/controllers/roadmap.ts` to fetch a planner by its `share_id`

## Test Plan

1. To verify `share_id` generation, make sure you are logged in to your account and make a new planner, then save. Check the database `planner` table and the new planner should have a `share_id` (should not be NULL)
2. For endpoint, you can check by using `fetch` in devtools console with the path `http://localhost:3000/planner/api/trpc/roadmaps.getByShareId`. Check that it returns the correct planner using a valid `share_id`.
3. Repeat step 2 with an invalid `share_id`. It should give you a bad request with the error message "Shared planner not found".
4. Test `roadmaps.get` by looking at devtools network when page loads and check that `shareId` is provided for planners that have one.

## Issues

Closes #1048

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -13,6 +13,7 @@
   },
   "dependencies": {
     "@opentelemetry/auto-instrumentations-node": "0.58.0",
+    "@paralleldrive/cuid2": "^3.3.0",
     "@trpc/server": "^10.45.2",
     "@vendia/serverless-express": "^4.12.6",
     "arctic": "^3.7.0",
--- a/api/src/controllers/roadmap.ts
+++ b/api/src/controllers/roadmap.ts
@@ -1,4 +1,4 @@
-import { router, userProcedure } from '../helpers/trpc';
+import { router, userProcedure, publicProcedure } from '../helpers/trpc';
 import {
   RoadmapDiffs,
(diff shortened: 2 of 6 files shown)


CadenLee2 (commented):
Tested in the console in staging and getting the planner from the ID worked well for me; code looks good as well.

However, we do also need to pass back each roadmap's `shareId` from the database on `roadmaps.get`, so that the frontend will eventually be able to make a shareable URL for the user. This means:
1. Modify the `savedPlannerData` to have an optional `shareId`
2. Modify `queryGetPlanners` in `api/src/helpers/roadmap.ts` to get the `shareId` from the database if it exists. Luckily you should be able to just change the outermost parts of the select (similar to how `chc: planner.chc` was added) and none of the cursed giant query lol

When testing, for any planners with a `shareId`, it should show up in the `roadmaps.get` response when the user loads into the page (check devtools network)

CadenLee2 (approved):
LGTM
Image

PR #1053, opened 2026-04-06
hide-popovers

JasonNguyen067:
## Description

Hiding popovers when you are dragging a course and it hovers over the text part of the name which causes the popover.

## Screenshots

Screenshot 2026-04-06 at 3 28 22 PM

## Test Plan

Drag a course while hovering over the name with your cursor to see no popover.

## Issues

Closes #1049


CadenLee2 (commented):
It seems you're using `document.body.classList` just as a way to store the state independently of props. However, we have Redux, so we can do the same thing with Redux (store whether a course is currently being dragged or not, and update it when a course is dragged/stops being dragged).

Notice that in places where you update the classList, generally `setActiveCourse` is being called. So, we should be able to just use `activeCourse` as a way to tell whether something's being dragged or not.

See my specific comment below

CadenLee2 (approved):
LGTM!

PR #1054, opened 2026-04-06
Overriding reqs backend endpoints 

### LINKED ISSUE #1004: Overriding requirements backend

DerrickThrower:
built the endpoints for the override table.

## Description

- addOverride inserts a row into the override table keyed by (userId, plannerId, requirement)
- deleteOverride removes the matching row 

## Issues

Closes #1004


CadenLee2 (approved):
Tested the endpoints on staging and local (e.g. post to `http://localhost:3000/planner/api/trpc/override.addOverride`), everything seems to work properly.

Adding, getting, and removing an override works well

Code lgtm

PR #1057, opened 2026-04-08
Save roadmap year collapsed state

### LINKED ISSUE #1035: Save current roadmap year collapse statuses in DB

In the roadmap, I'm currently able to collapse/open each year row, however these rows are always open when I first open the app. When opening the app, to view your schedule for later years (e.g. 3rd/4th years), you have to, **every time**, either collapse the 1st/2nd year rows, or scroll down. Ideally, I could collapse these rows once, and have that status saved to the database such that they stay closed when I open my roadmap another time.

Nope

Behavior doesn't change the user at all, in that they don't need to do anything extra.

N/A

**Additional context**

Similar to Varun's #983

atile4:
## Description

- updated planner year variable types to accommodate for the change
- Store roadmap collapsed state on save
 

## Test Plan

- Create a roadmap with one or more years
- Collapse a year (at least 1 year expanded and 1 year collapsed)
- Save and reload
- Check that the collapsed state matches what you had before
## Issues

Closes #1035


CadenLee2 (approved):
The multiple roadmaps issue I mentioned previously seems fixed

Tested:
- Collapsing and expanding with multiple roadmaps
- Saving and loading while logged in and out
- Importing transcript / zot4plan

Code looks good!

PR #1064, opened 2026-04-12
feat: collapsed years schema

CadenLee2:
Standalone database migration for #1057 due to new policy to prevent conflicts. This specific one had a conflict with main (`0028` in the original PR, now re-numbered to `0030`)

Before merging, prod should be migrated. After merging, #1057 should be updated from main to use these schema numbers

## Test Plan

- [x] Lint passes
- [x] Staging works

## Issues

Closes #


atile4 (approved):
LGTM

PR #1065, opened 2026-04-12
Better Review Cards

### LINKED ISSUE #1014: Better review cards
Update the appearance of review cards to match the new designs

## Design
Image
Image

jchau06:
## Description

This PR updates the review card component to reflect its current design in the Figma. Changes include redesign of card header, info, and footer, redesigning buttons, tags, and menus, and updating the color scheme of the card. 

>[!NOTE]
> The Figma currently only shows the dark mode design. With light mode, the light blue color in dark mode transforms into the same dark blue color, so the quality and difficulty colors will look the same.

## Screenshots

Before:
image

Mobile:
image

After:
image

image

Light Mode:
image

Mobile:
image

## Test Plan

1. Search for a course or a professor.
2. Add a review for a specific course and professor
3. Observe UI changes with review card component (redesigns + updated color schemes)
4. Press three-dot icon to check updated menu.
5. Switch between light and dark mode to verify proper styling. 

## Issues

Closes #1014


fiveminus1:
🔥🫡 Make sure this PR pulls from and merges into `better-reviews`!

site/src/component/Review/ReviewCard.scss
```diff
@@ -175,57 +221,94 @@
...
 }
 
-.reviewcard-voting {
+.reviewcard-tags {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 10px;
+
+  .MuiChip-root {
+    border-radius: 10px;
```
CadenLee2: Small detail: the border radius of the tags should not be set explicitly to `10` even though that's what the Figma has, because the goal is just that they're rounded (which is the default)

site/src/component/Review/ReviewCard.tsx
```diff
@@ -65,14 +83,82 @@ const AuthorEditButtons: FC<AuthorEditButtonsProps> = ({ review, course, profess
...
+  const openReportForm = () => {
+    setReportFormOpen(true);
+    handleMenuClose();
+  };
+
+  const menuItemSx = {
+    width: 'calc(100% - 8px)',
+    height: '22px',
+    borderRadius: '4px',
+    border: '1px solid color-mix(in srgb, var(--mui-palette-text-secondary) 50%, transparent)',
```
CadenLee2: If possible we should consolidate all these into the scss file. If this is unfeasible lmk

site/src/component/Review/ReviewCard.tsx
```diff
@@ -286,110 +405,88 @@ const ReviewCard: FC<ReviewCardProps> = ({ review, course, professor, children }
...
+            {review.verified && <div className="reviewcard-author-verified">{verifiedIcon}</div>}
+            {review.authored && <div className="reviewcard-author-author">{authorIcon}</div>}
+          </div>
+          <span className="reviewcard-date">
+            {new Date(review.createdAt).toLocaleString('default', {
+              year: 'numeric',
+              month: '2-digit',
+              day: '2-digit',
+            })}
+            {review.updatedAt && <span className="subtext edit-time"> (edited {new Date().toLocaleDateString()})</span>}
```
CadenLee2: Bug: `review.updatedAt` is not passed into the `Date()` and thus we see an incorrect edit date:
Image

Also wondering why there's an inconsistency between the code used to format the above date and this date

CadenLee2 (commented):
## Styling

The upvote/downvote buttons should always go in the right corner, even if there are no tags; this can probably be accomplished with something like `margin-left: auto;`. See the inconsistency in the image below

Also notice that the review top section is oddly tall, due to the icon button on the right. It may look more accurate to do the following:
- Decrease the padding at the top of the root review card paper
- Shrink the three dots icon button to be `32px` by `32px`. This is acceptable because it matches other icon buttons like those in the multiple planner selector
- Remove the `padding-bottom: 8px`. This does
Image

(This image is after the second piece of advice is applied. Note that spacing looks more natural):
Image

## Functionality
Seems to work well, I didn't test too thoroughly yet though. Identified one date bug (see code comments)

## Code
See comments. Didn't look over completely yet

jchau06:
## Screenshot

image

## Fixes
1. Upvote / downvote buttons are always on the bottom left of the cards regardless of if there are tags. 
2. Padding of review top section has been changed (do check if spacing is accurate). 
3. menuItemSx removed and styling changes were moved to SCSS. 
4. Date formatting and bug is resolved.

site/src/component/Review/ReviewCard.scss
```diff
@@ -16,20 +16,66 @@
...
+    }
+
+    .reviewcard-author-verified {
+      display: flex;
+      align-items: center;
+
+      svg {
+        width: 14px;
+        height: 14px;
+        color: #90b3f9;
```
CadenLee2: Can use `font-size: 14px;` instead of setting the width and height

Also, the color should be our blue MUI color variable, like the other places where blue is used

site/src/component/Review/ReviewCard.scss
```diff
@@ -175,57 +221,115 @@
...
+      }
+
+      &.colored-upvote {
+        background: transparent;
+
+        svg {
+          color: var(--mui-palette-secondary-main);
+          width: 15.5px;
+          height: 15.5px;
+          font-size: 15.5px;
```
CadenLee2: Same font-size comment on these svgs.

Also, given that `.colored-upvote` and `.colored-downvote` have the same styles, you could remove duplication between the two by unifying them into `.colored-vote` and applying that in the tsx

site/src/component/Review/ReviewCard.scss
```diff
@@ -175,57 +221,115 @@
...
-  border: none;
-  background-color: transparent;
-  text-decoration: underline;
+.MuiMenuItem-root.review-menu-item {
+  width: calc(100% - 8px);
+  height: 22px;
+  border-radius: 4px;
+  border: 1px solid color-mix(in srgb, var(--mui-palette-text-secondary) 50%, transparent);
+  color: var(--mui-palette-secondary-main);
+  font-family: Roboto;
```
CadenLee2: Was setting the font-family necessary here? If not, then this should be removed; if it was necessary, you can likely change it to `font-family: inherit;` so that it keeps the same font as the rest of the page

site/src/component/Review/ReviewCard.tsx
```diff
@@ -286,110 +395,97 @@ const ReviewCard: FC<ReviewCardProps> = ({ review, course, professor, children }
...
+            <div className="reviewcard-voting-buttons">
+              <Tooltip title="You must be logged in to vote" open={isLoggedIn ? false : undefined}>
+                <span>
+                  <button
+                    className={upvoteClassname}
+                    onClick={upvote}
+                    disabled={!isLoggedIn}
+                    style={!isLoggedIn ? { pointerEvents: 'none' } : {}}
+                  >
+                    <span>{review.score > 0 ? review.score : 0}</span>
```
CadenLee2: This score calculation is a bit inaccurate because it'll never show both upvotes and downvotes being nonzero. I made issue #1078 to fix it later, that'll be a future PR

site/src/component/Review/ReviewCard.tsx
```diff
@@ -286,110 +395,97 @@ const ReviewCard: FC<ReviewCardProps> = ({ review, course, professor, children }
...
+            <span className="reviewcard-author-name">{review.userDisplay}</span>
+            {review.verified && <div className="reviewcard-author-verified">{verifiedIcon}</div>}
+            {review.authored && <div className="reviewcard-author-author">{authorIcon}</div>}
+          </div>
+          <span className="reviewcard-date">
+            {new Date(review.createdAt).toLocaleString('default', {
+              year: 'numeric',
+              month: '2-digit',
+              day: '2-digit',
+            })}
```
CadenLee2: To reduce duplication/messiness here and below, I suggest making a helper function `displayReviewDate` that encapsulates the date display logic and calling it like `{displayReviewDate(review.createdAt)}`

CadenLee2 (commented):
Tested this out some more and it's looking very good!

1. Make sure to update from main, hopefully the merge conflict isn't too bad lol
2. A few more code comments

CadenLee2 (approved):
LGTM (just make sure to update from main, then lmk)

types/src/review.ts
```diff
@@ -1,6 +1,6 @@
 import { z } from 'zod';
 
-export const anonymousName = 'Anonymous Peter';
+export const anonymousName = 'anonymous_petr';
```
fiveminus1: I don't think this should change

site/src/component/Review/ReviewCard.tsx
```diff
@@ -65,14 +83,72 @@ const AuthorEditButtons: FC<AuthorEditButtonsProps> = ({ review, course, profess
...
+        onClose={handleMenuClose}
+        slotProps={{
+          paper: {
+            sx: {
+              width: '96px',
+              minWidth: 'unset',
+              borderRadius: '8px',
+              border: '1px solid color-mix(in srgb, var(--mui-palette-text-secondary) 40%, transparent)',
+              overflow: 'hidden',
+              boxSizing: 'border-box',
```
fiveminus1: In general, we try to avoid using `sx` prop styling so that all styling is in one spot--the SCSS. 

1. Please refactor any styling to SCSS.
2. I know this tries to follow the Figma, but we should probably use as close to the vanilla MUI look as possible for these menus as they look pretty out of place at the moment. Give it a shot and send a screenshot in the Discord

site/src/component/Review/ReviewCard.tsx
```diff
@@ -283,113 +392,97 @@ const ReviewCard: FC<ReviewCardProps> = ({ review, course, professor, children }
...
+              <Chip size="small" key={tag} label={tag} />
+            ))}
+          </div>
+        )}
+        <div className="reviewcard-footer" id={review.id.toString()}>
+          <div className="reviewcard-voting">
+            <div className="reviewcard-voting-buttons">
+              <Tooltip title="You must be logged in to vote" open={isLoggedIn ? false : undefined}>
+                <span>
+                  <button
```
fiveminus1: This should be refactored to an `IconButton`

site/src/component/Review/ReviewCard.scss
```diff
@@ -16,20 +16,65 @@
...
+    justify-content: space-between;
+    align-items: center;
+    padding-top: 8px;
+  }
 
   .reviewcard-identifier {
-    font-weight: bold;
-    font-size: 20px;
+    font-weight: 600;
+    font-size: 17px;
```
fiveminus1: 17 is an odd font size; try 18? I also don't think the font weight needs to be set

site/src/component/Review/ReviewCard.scss
```diff
@@ -68,39 +116,36 @@
 
   .rating {
```
fiveminus1, line 81: The rating boxes just look odd to me; try experimenting with the size

site/src/component/Review/ReviewCard.scss
```diff
@@ -175,57 +220,101 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  gap: 10px;
+
+  .MuiChip-root {
+    font-family: Roboto;
+    font-weight: 600;
+    font-size: 12px;
+    line-height: 100%;
+  }
```
fiveminus1: Honestly I think none of these styles are necessary

fiveminus1 (changes requested):
Great job overall! Follows Figma quite well, but there are some areas it probably shouldn't

jchau06:
## Screenshot
Screenshot 2026-04-29 at 12 15 57 AM

## New Changes
1. Reverted anonymousName change
2. Removed sx styling + using default MUI menu styling
3. Refactored voting buttons to IconButtons
4. Minor font changes + removed unnecessary styling
5. Experimented with the size of menus + rating boxes (open to further changes)

site/src/component/Review/ReviewCard.scss
```diff
@@ -16,20 +16,64 @@
 
 .reviewcard-header {
   display: flex;
-  justify-content: space-between;
-  align-items: flex-start;
-  gap: 12px;
+  flex-direction: column;
+  gap: 0;
```
fiveminus1: Can remove this

site/src/component/Review/ReviewCard.scss
```diff
@@ -16,20 +16,64 @@
...
+      }
+    }
+  }
+
+  .reviewcard-date {
+    font-size: 12px;
+    color: var(--mui-palette-text-secondary);
+    white-space: nowrap;
+  }
+
```
fiveminus1: Maybe a nit but I'd prefer if these were nested within the `reviewcard-header-top` and `reviewcard-header-bottom` rules for readability

site/src/component/Review/ReviewCard.scss
```diff
@@ -16,20 +16,64 @@
...
     a {
       color: var(--mui-palette-text-primary);
     }
     word-break: break-word;
   }
 
+  .reviewcard-author {
+    display: flex;
+    align-items: center;
+    gap: 4px;
```
fiveminus1: I think we should go with `gap: 0` here actually, the icons look a little too far to the right imo

site/src/component/Review/ReviewCard.scss
```diff
@@ -16,20 +16,64 @@
...
+
+    .reviewcard-author-name {
+      font-weight: 500;
+      font-size: 12px;
+      color: var(--mui-palette-text-secondary);
+    }
+
+    .reviewcard-author-verified {
+      display: flex;
+      align-items: center;
```
fiveminus1: These shouldn't be necessarily I believe

site/src/component/Review/ReviewCard.scss
```diff
@@ -16,20 +16,64 @@
...
+
+      svg {
```
fiveminus1: Target `.MuiSvgIcon-root` here instead

site/src/component/Review/ReviewCard.scss
```diff
@@ -16,20 +16,64 @@
...
+        font-size: 14px;
+        color: var(--mui-palette-secondary-main);
+      }
+    }
+  }
+
+  .reviewcard-date {
+    font-size: 12px;
+    color: var(--mui-palette-text-secondary);
+    white-space: nowrap;
```
fiveminus1: I also don't think you need this

site/src/component/Review/ReviewCard.scss
```diff
@@ -68,39 +115,36 @@
...
+
+  .rating-difficulty {
+    background-color: var(--mui-palette-secondary-main);
   }
-  .r4 {
-    background-color: var(--mui-palette-reviews-green);
+  .rating-label {
+    font-size: 10px;
+    font-weight: 400;
+    text-align: center;
```
fiveminus1: Can remove these

site/src/component/Review/ReviewCard.scss
```diff
@@ -68,39 +115,36 @@
...
+    line-height: 100%;
   }
-  .r5 {
-    background-color: var(--mui-palette-reviews-blue);
+  .rating-value {
+    font-size: 30px;
+    font-weight: 400;
+    text-align: center;
```
fiveminus1: Remove these too

site/src/component/Review/ReviewCard.scss
```diff
@@ -175,57 +219,96 @@
...
+      cursor: pointer;
+
+      .vote-count {
+        font-weight: 700;
+        font-size: 12px;
+      }
+
+      .MuiIconButton-root {
+        padding: 2px;
+        border-radius: 0 3px 3px 0;
```
fiveminus1: Can remove this

site/src/component/Review/ReviewCard.scss
```diff
@@ -175,57 +219,96 @@
...
+        color: var(--mui-palette-text-secondary);
+
+        svg {
```
fiveminus1: Same comment here about targetting MuiSvg class

site/src/component/Review/ReviewCard.scss
```diff
@@ -175,57 +219,96 @@
...
+          font-size: 14px;
+        }
+
+        &:hover {
+          background-color: transparent;
+        }
+      }
+
+      &.colored-vote {
```
fiveminus1: You can handle this in JSX instead, a separate class isn't needed

site/src/component/Review/ReviewCard.scss
```diff
@@ -175,57 +219,96 @@
...
-  color: var(--mui-palette-text-primary);
+.review-menu {
+  .MuiList-root {
+    padding: 4px 0;
+  }
+
+  .MuiMenuItem-root {
+    font-size: 10px;
+    min-height: unset;
+    padding: 2px 10px;
```
fiveminus1: Sorry, let's go with 14 font and remove the min-height and padding

site/src/component/Review/ReviewCard.tsx
```diff
@@ -165,6 +207,17 @@ const ReviewCard: FC<ReviewCardProps> = ({ review, course, professor, children }
...
+    const [year, term] = quarter.split(' ');
+    const termMap: Record<string, string> = {
+      Fall: 'F',
+      Winter: 'W',
+      Spring: 'Sp',
+      Summer: 'Su',
+    };
+    return `${termMap[term] ?? term}${year.slice(-2)}`;
+  };
+
```
fiveminus1: Move this to `helpers/reviews.ts`

site/src/component/Review/ReviewCard.tsx
```diff
@@ -283,113 +351,87 @@ const ReviewCard: FC<ReviewCardProps> = ({ review, course, professor, children }
...
   if (review.attendance) tags.unshift('Mandatory attendance');
 
+  const displayReviewDate = (date: string | Date) => {
+    return new Date(date).toLocaleString('default', {
+      year: 'numeric',
+      month: '2-digit',
+      day: '2-digit',
+    });
+  };
+
```
fiveminus1: Can this be refactored to a helper too?

site/src/component/Review/ReviewCard.tsx
```diff
@@ -255,12 +327,8 @@ const ReviewCard: FC<ReviewCardProps> = ({ review, course, professor, children }
...
   };
 
-  const openReportForm = () => {
-    setReportFormOpen(true);
-  };
-
-  const upvoteClassname = review.userVote === 1 ? 'upvote colored-upvote' : 'upvote';
-  const downvoteClassname = review.userVote === -1 ? 'downvote colored-downvote' : 'downvote';
+  const upvoteClassname = review.userVote === 1 ? 'upvote colored-vote' : 'upvote';
+  const downvoteClassname = review.userVote === -1 ? 'downvote colored-vote' : 'downvote';
```
fiveminus1: Same comment as earlier, but this should be able to be handled via JSX I believe

fiveminus1 (changes requested):
One major comment about menu styling (sorry for going back on that) and some minor unnecessary styling comments. Should be good after these!

site/src/component/Review/ReviewCard.scss
jchau06: Removed gap: 0, flex-direction: column is needed.

site/src/component/Review/ReviewCard.scss
jchau06: Removed unnecessary white-space style. font-size and color are necessary.

site/src/component/Review/ReviewCard.scss
jchau06: Removed unnecessary border-radius, padding and color are necessary.

site/src/component/Review/ReviewCard.scss
```diff
@@ -16,31 +16,65 @@
...
+      color: var(--mui-palette-text-secondary);
+    }
+
+    .reviewcard-identifier {
+      font-size: 18px;
+      margin: 0;
+      a {
+        color: var(--mui-palette-text-primary);
+      }
+      word-break: break-word;
```
fiveminus1: I think you can remove this too

fiveminus1 (approved):
LGTM! Thanks for being so responsive and quick about the changes. One final comment about an unneeded style you can probably take a look and fix without me re-reviewing. @CadenLee2 might want another look?

EDIT: Upon a second look, there are some dead stylings that I just commented on. Please remove those, and ensure there aren't any more throughout the file. 

--
I do think this looks a little strange on mobile but it's not worth nitpicking and it's probably a unique case we can try to make look better later via new issue if we do think it's weird
Image

site/src/component/Review/ReviewCard.scss
jchau06: Removing unnecessary word-break, font-size, margin, and color targetting a is necessary.

CadenLee2 (approved):
Tested:
- Various screen sizes
- All buttons
- Light and dark theme
- Edited dates

Everything seems to work/look as expected, and code looks good!

Will wait for Jeremiah's final confirmation as well, then will be good to merge

fiveminus1:
Very nice! Thank you for all the work on this!

Comment for a future issue/PR: we might want to consider cutting tags off at 3 or so on mobile and having a "View more" button or something of the sort...these will take up too much space over time.
image

PR #1066, opened 2026-04-12
Out of sync clarifications

### LINKED ISSUE #1061: Roadmap out of sync clarifications
As a reminder, the "roadmap out of sync" modal appears when you save changes while logged out, then log back into your account. The choice users make with this modal is very important, but it's not indicated as such because it's so dismissable.

To clarify to users what decision they're making, we have discussed a few indicators that might be helpful:
- The number of planners/courses in each option
- When each option was last edited

Update the Roadmap Out of Sync modal to show this information, using the sketch below as reference

## Rough sketch
This shows the content, but make sure to **use MUI components and make the modal look standardized with our other modals; don't worry about too much custom styling**. If you have a question on the appearance, or a suggestion for an improvement, message on Discord and feel free to change anything.

Image

Main considerations here are:
- Saving your device roadmap is *very destructive* as it overrides your whole account. Hence red
- Saving your account roadmap is *somewhat destructive* as it discards the device roadmap. Hence blue
- Dismissing the modal *postpones the decision* as the modal will come back on reload.
  - The current modal is problematic because it visually implies that the device option is like a cancel/leave/dismiss

Eventually we should consider how to further improve this modal as a whole team, including with UI/UX input, but currently a lot of us are busy with the booth setup and other tasks

atile4:
## Description
- add details about local/synced roadmap into the out of sync modal

## Screenshots

| original | new |
| --- | --- |
|image|image|

## Test Plan
Make sure "This Device" and "My Account" buttons work the same way as before

**Note:** design is not finalized, will need info from team meeting discussions on the final design of the modal. Right now this PR just has a basic structure with all the info (planner/course counts, etc) that will be needed in the future.
## Issues

Closes #1061


site/src/app/roadmap/planner/PlannerLoaderModal.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  const rtf = new Intl.RelativeTimeFormat('en', { numeric: 'auto' });
+
+  if (Math.abs(seconds) < 60) return rtf.format(seconds, 'second');
+  if (Math.abs(minutes) < 60) return rtf.format(minutes, 'minute');
+  if (Math.abs(hours) < 24) return rtf.format(hours, 'hour');
+  return rtf.format(days, 'day');
+};
+
+// count planners + courses + lastEdited
```
CadenLee2, line 23: code comment can be removed

site/src/app/roadmap/planner/PlannerLoaderModal.tsx
```diff
@@ -0,0 +1,152 @@
...
+              loading={overrideLoading}
+              disabled={overrideLoading || accountLoading}
+              color="error"
+              variant="contained"
+              onClick={overrideAccountRoadmap}
+              sx={{ marginTop: 4, alignSelf: 'flex-start' }}
+            >
+              This Device
+            </Button>
+            <DialogContentText sx={{ marginTop: 2, fontSize: 12, color: 'red' }}>
```
CadenLee2: Color should be `var(--mui-palette-error-main)`

CadenLee2 (commented):
Tested this out (both options), works well and looks good! 2 small details

CadenLee2 (approved):
```
L       GGG   TTTTT   M M
L      G        T    M M M
L      G   G    T    M M M
LLLLL   GGGG    T    M   M
```

PR #1071, opened 2026-04-15
Preview Navbar

MajdKatabi:
## Issues

Closes #


site/src/component/ResultPreview/ProfessorPreview.tsx
```diff
@@ -47,27 +48,35 @@ const ProfessorPreviewContent: FC<{ data: ProfessorGQLData | null }> = ({ data }
...
-      <ResultPageSection title="🗓️ Schedule of Classes">
-        <Schedule professorIDs={data.shortenedNames} termsOffered={unionTerms(data.courses)} />
-      </ResultPageSection>
-
-      <ResultPageSection title="💬 Reviews">
-        <Review professor={data} terms={sortTerms(getProfessorTerms(data))} />
-      </ResultPageSection>
+      <div id="preview-details">
+        <SideInfo
+          className="professor-summary"
```
CadenLee2: For this `SideInfo` and the `ResultPageSection`s below (incl. in `CoursePreview.tsx` too), I suggest instead having `SideInfo` and `ResultPageSection` take in an optional `id` prop and use it to set the id of their respective top-level divs. This will reduce the amount of nesting/duplication

site/src/component/ResultPreview/ResultPreview.scss
```diff
@@ -58,4 +86,10 @@
     max-width: 1000px;
     margin-bottom: 32px;
   }
+
+  @media (max-width: globals.$mobile-cutoff) {
+    .preview-nav-bar {
+      display: none;
+    }
+  }
```
CadenLee2: There are non-mobile screen sizes where the navbar overlaps with the course name. These cases should be treated the same as mobile (i.e. should be based on the width of the container). Consider [container query](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries#naming_containment_contexts) and choose a cutoff width that works for all course names.

In the future this can be extended to provide the mobile navigations when we reach that cutoff

Image

site/src/component/ResultPreview/PreviewNavBar.tsx
```diff
@@ -0,0 +1,67 @@
...
+      setActiveSection(currentSection.id);
+    };
+
+    updateActiveSection();
+    scrollContainer.addEventListener('scroll', updateActiveSection);
+    return () => scrollContainer.removeEventListener('scroll', updateActiveSection);
+  }, []);
+
+  const scrollToSection = (id: string) => {
+    setActiveSection(id);
```
CadenLee2: The interaction is currently a bit odd because you click an option (e.g. "reviews") and it flashes blue for a second (because of this line) but then goes back to gray until the smooth scroll animation reaches it.

Ideally this flash shouldn't happen. The simplest solution is to remove this line so that the active section can entirely be dependent on the scroll. This does mean the clicked item won't turn blue immediately, but it might help save a lot of extra logic.

CadenLee2 (commented):
I recommend getting rid of the absolute position of the preview navbar, and instead letting it flow naturally with the rest of the flex items. Reference the image below (note how it's not exactly in the center, but instead has an equal margin on the left and right).

This is to reduce the likelihood of overlap with the course names on small screens (see code comment on media query)

Image

CadenLee2 (approved):
Tested out further, the short pages issue does exist but is minor enough that it can be addressed in a follow-up. LGTM

PR #1079, opened 2026-04-20
Fix plan index out of range crash

CadenLee2:
Fixes a recent crash users have been experiencing

## Description

Some users have a plan index in the database that is out of range given the number of planners they have. This caused the frontend to crash

## Test Plan

- [ ] In the dev database, manually edit your current plan ID to be out of range
  - [ ] Most stagings should crash
  - [ ] This staging should work

## Issues

Closes #


anthonyj33 (approved):
LGTM; Tested with index out of range and worked fine.

Negative index doesn't work but I don't see that ever being a problem

PR #1082, opened 2026-04-22
Fix saving undefined variable units

### LINKED ISSUE #1062: Saving edited undefined variable units
Steps to recreate:
1. Add a course with variable units (e.g. CS 199)
2. Add variable units to it and save. Saving should work
3. Edit the variable units, delete everything in the text box, and press the check mark
4. Save. It will fail (check the console)

This is because the `UnitsContainer` component sets the units to `NaN` if they cannot be converted to a float. To fix this, we should set the units to `undefined` if they are `NaN`. This will involve making `UnitsContainer` accept a `setUnits` function that takes either a number or undefined, and then ensuring that `MenuTile`'s behavior is not changed (i.e. it should still only allow a number).

fiveminus1:
## Description

Added handling for NaN parsed units to fix roadmap saving when a variable units course is saved with an empty input.

## Screenshots

N/A

## Test Plan

- [ ] Add course with variable units (CS 199)
- [ ] Save with a valid variable unit 
- [ ] Edit the variable units, delete everything in text box, then confirm
- [ ] Save your roadmap, should go through
- [ ] No other behavior changed

## Issues

Closes #1062

### DIFF
--- a/site/src/app/roadmap/CustomUnitsContainer.tsx
+++ b/site/src/app/roadmap/CustomUnitsContainer.tsx
@@ -8,7 +8,7 @@ import { IconButton } from '@mui/material';
 
 interface UnitsContainerProps {
   units: number | undefined;
-  setUnits?: (value: number) => void;
+  setUnits?: (value: number | undefined) => void;
   minUnits: number | undefined;
   maxUnits: number | undefined;
   source: string;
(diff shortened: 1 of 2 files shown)


CadenLee2 (approved):
Works in transfers menu (does not allow undefined) and roadmap (allows undefined). Saving works

Issue #1078, opened 2026-04-20
Return upvotes and downvotes from review endpoint


CadenLee2:
Currently the `getReviews` endpoint sums up votes into one `score`. To display upvotes and downvotes independently, the endpoint should return both upvotes and downvotes.

The frontend should then be adjusted (after #1065 is merged) to show both upvotes and downvotes properly, rather than using the score

PR #1087, opened 2026-04-24
Return upvote and downvote from reviews endpoint

JasonNguyen067:
## Description

Summed up the counts of upvotes and downvotes separately in the backend and displayed in review cards.

## Screenshots

Screenshot 2026-05-19 at 3 29 15 PM
shows both upvote and downvote separately instead of just one total count reflected for either upvote/downvote. For example if there was 3 upvotes and 2 downvotes before the change it would just show 1 total upvote due to the total score being 1 (3-2) now it will properly show the counts separately.

## Test Plan

Log into a acc, create a review put a upvote on one acc log onto local with a incognito browser and put a downvote and see the separate counts.

## Issues

Closes # [1078](https://github.com/icssc/peterportal-client/issues/1078)

### DIFF
--- a/api/src/controllers/reviews.ts
+++ b/api/src/controllers/reviews.ts
@@ -54,6 +54,8 @@ async function getReviews(
   const results = await db
     .select({
       review: review,
+      upvotes: sql`COALESCE(SUM(CASE WHEN ${vote.vote} = 1 THEN 1 ELSE 0 END), 0)`.mapWith(Number),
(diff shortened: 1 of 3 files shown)


CadenLee2 (approved):
Tested:
- Sorting continues to work
- Works on reviews with multiple upvotes + downvotes
  - Image

Code changes look good

Feel free to update from main then merge!

PR #1093, opened 2026-04-27
Search major minor Abbreviations

### LINKED ISSUE #1086: Search for major/minor by abbreviations
During testing, users frequently tried searching for majors by their initials (e.g. 'cs' for "Computer Science", 'bim' for "Business Information Management"). Though these are not always official major initials, users often expect them to be supported.

We can create a simple solution by going based on *capital initials from the full name* (i.e. "Computer Science and Engineering" -> cse). This can then be used as an additional way for users to search (e.g. perhaps it can be added to the value passed into the Select). Experiment with how the search best works.

In terms of layout or styling nothing should change, but when you search "cs", "bim", or other major initials, the majors show up (ideally first). The same should be done for minors.

In the future, API may be able to serve us some related information, but many of these abbreviations may be informal. This will require some research and is out of scope here

Image

atile4:
## Description

- added filtering by abbreviations to the major/minor search dropdown

## Screenshots
image

image

## Test Plan

- check that searching by a major/minor's capital letters will give that major/minor
- check that searching for minors doesn't include the "Minoring in" as an abbreviation
  - ex: "MS" doesn't query "Minoring in Statistics"

## Issues
Closes #1086

### DIFF
--- a/site/src/app/roadmap/catalog/MinorSelector.tsx
+++ b/site/src/app/roadmap/catalog/MinorSelector.tsx
@@ -1,5 +1,5 @@
-import { FC, useCallback, useEffect, useState } from 'react';
-import { Autocomplete, TextField } from '@mui/material';
+import { FC, useCallback, useEffect, useMemo, useState } from 'react';
+import { Autocomplete, FilterOptionsState, TextField } from '@mui/material';
 import trpc from '../../../trpc';
 import { normalizeMajorName } from '../../../helpers/courseRequirements';
 import { addMinor, removeMinor, setMinorList, MinorRequirements } from '../../../store/slices/courseRequirementsSlice';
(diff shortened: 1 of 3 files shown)


site/src/app/roadmap/catalog/MinorSelector.tsx
```diff
@@ -97,6 +98,39 @@ const MinorSelector: FC = () => {
...
+  //   // list of filtered minors
+  //   const filtered = options.filter(
+  //     (option) =>
+  //       option.label.toLowerCase().includes(state.inputValue.toLowerCase()) &&
+  //       !abbrFiltered.some((a) => a.value.id === option.value.id),
+  //   );
+
+  //   // abbreviated matches first
+  //   return [...abbrFiltered, ...filtered];
+  // };
```
anthonyj33: Let's clean up this comment before merging

anthonyj33 (commented):
Code looks good, other than 1 small comment below

From staging testing, this works as expected. I did notice that some things didn't work how _**i personally**_/intuitively expected them to (e.g. try searching for "cs" -- computer science is not the first result), however all of these small things can't really be resolved cleanly AFAIK.

One small change that might be nice is treating `/` as a separator (and any other similar characters, if there are any I missed)

site/src/helpers/selector.ts
```diff
@@ -0,0 +1,50 @@
...
+  if (input) {
+    for (const [abbr, fullName] of Object.entries(abbreviations)) {
+      if (abbr.startsWith(input)) {
+        abbrMatches.push(...fullName);
+      }
+    }
+  }
+
+  const abbrFiltered = options.filter((option) =>
+    abbrMatches.some((term) => option.label.toLowerCase().includes(term.toLowerCase())),
```
CadenLee2: Would suggest changing `term` here to `fullName` just so it's more clear what's going on

CadenLee2 (commented):
Looks good to me as well, just one more comment in addition to Anthony's

Also:
> try searching for "cs" -- computer science is not the first result

This unfortunately is something we'll have to keep, but I think it's alright because people will at least see it relatively high up in the list. I agree that it's not what I expect too tho

> might be nice is treating / as a separator

I don't think this is too relevant, particularly because the few majors/minors with `/` likely won't go by their abbreviations anyway

anthonyj33 (approved):
LGTM!

anthonyj33:
> This unfortunately is something we'll have to keep, but I think it's alright because people will at least see it relatively high up in the list. I agree that it's not what I expect too tho

Yep, nothing we can do realistically:
> however all of these small things can't really be resolved cleanly AFAIK.

Issue #1076, opened 2026-04-18
Mobile popup improvements


CadenLee2:
As per UI/UX feedback, `MobilePopup` can use some improvements to clarify its role:
1. Slightly decrease the height of the popup/increase the space above it to indicate that there isn't some glitch with the top bar
2. Increase the border radius to something slightly more visible like `16px`
3. Fix a bug where scrolling for the first time scrolls the page behind (reach out to Charlie to see how to replicate. If this can't be replicated, this fix is not necessary for issue completion)
4. Allow a *swipe down* to close/dismiss the popup (test with devtools mobile touch emulation)
    - Likely a more involved change, but should be able to find examples online

<img width="300" height="auto" alt="Image" src="https://github.com/user-attachments/assets/3de736a4-fc21-41be-be83-b7f2dc38871e" />

PR #1096, opened 2026-04-28
Mobile popup improvement

JasonNguyen067:
## Description

Decreased height of the pop-up. Updated border radius of the popup, fixed scroll logic, and added a swipe down animation to close the popup.

## Screenshots
Before:
Screenshot 2026-04-27 at 9 17 30 PM

After:
Screenshot 2026-04-27 at 9 13 09 PM

## Test Plan

Click on add course, verify that it's the correct height with border radius, try scrolling to see just the pop-up scrolling and not the background and swipe down on the pop-up to see if it closes. 

## Issues

Closes #[1076](https://github.com/icssc/peterportal-client/issues/1076)


site/src/app/roadmap/MobilePopup.tsx
```diff
@@ -16,6 +16,42 @@ const MobilePopup: FC<MobilePopupProps> = ({ show, onClose, className, id, child
...
+    };
+
+    element.addEventListener('touchstart', handleTouchStart);
+    element.addEventListener('touchend', handleTouchEnd);
+    element.addEventListener('touchmove', handleTouchMove);
+
+    return () => {
+      element.removeEventListener('touchstart', handleTouchStart);
+      element.removeEventListener('touchend', handleTouchEnd);
+      element.removeEventListener('touchmove', handleTouchMove);
```
CadenLee2: Also, I notice that tapping the background (area above the popup) does close it, but trying to drag it down via the background doesn't work. Could you change the element used here to something that allows dragging on the whole screen?

site/src/app/roadmap/MobilePopup.tsx
```diff
@@ -16,6 +16,42 @@ const MobilePopup: FC<MobilePopupProps> = ({ show, onClose, className, id, child
...
+      element.style.transform = '';
+      element.style.transition = '';
+    };
+
+    const handleTouchMove = (e: TouchEvent) => {
+      const delta = e.touches[0].clientY - touchStartY.current;
+      if (delta > 0) {
+        element.style.transform = `translateY(${delta}px)`;
+      }
+    };
```
CadenLee2: This causes an issue when trying to scroll up. Choose a major, scroll down by swiping up, then try to scroll back up by swiping down. The swipe down action then problematically both scrolls and starts dragging the popup.

One solution might be only allowing the swipe down action when the `.mobile-popup` div (see tsx below and note that that's what provides the scrolling) scroll position is at the very top. Maybe still allow swiping down from the background, but just not from inside the popup.

Image

Image

CadenLee2 (commented):
When discussing in our meeting, the reasoning for changing the height of the mobile popup was to make it clear that it's a popup rather than some bug with the top bar. As such I think it might be more natural to have it *slightly* less tall to look more natural/independent from the top bar (see screenshot for an idea), but lmk what you think
Image

site/src/app/roadmap/MobilePopup.tsx
JasonNguyen067: Tapping the background to close the pop-up already existed before my changes. I implemented the drawdown on the actual pop-up. I'll take a look regarding adding the dragging effect on the whole screen.

JasonNguyen067: Imo it looks like unorganized if it's completely disconnected from the top bar, but I can go out and change it if you want, I think it looks better just slightly separated.

CadenLee2 (commented):
There are still some weird visual interactions, I think fixing these will boil down to fixing just a few details in the logic when it comes to borders/edge cases.

See the three I noticed here:
1. When partway scrolled down, the cancel button gets offset weirdly if you try to collapse the menu from *above*
2. When partway scrolled down, the menu glitches when you try to scroll back up and reach the top

https://github.com/user-attachments/assets/8a90c82e-a5da-4b60-858a-4dd160090961

3. When you try to collapse the menu from above then change your mind and move your finger back up, the collapsing fails
https://github.com/user-attachments/assets/fa671f27-aebe-4c99-80be-008b5c3ca0fa

anthonyj33 (approved):
LGTM!

There were a few strange behaviors I noticed on iOS but they were pretty minor and would be annoying to resolve so we can get this merged

PR #1100, opened 2026-04-29
Course Popover Revision

### LINKED ISSUE #1075: More useful course popover
In preparation for reviews potentially being integrated into the popover in the future:
1. Clamp the description and prerequisites to four lines
    - Potentially with an expand more button, though this shouldn't be too necessary
2. Above the Recent Offerings table, include a section for GPA information
    - TBD what info exactly/how to format it

## Current popover
Image

CLiang531:
## Description

- Clamped course description and prerequisite information to 4 lines
- Refactored grade fetching and statistic-calculating code from `Pie.tsx` and `GradeDist.tsx` into `gradeDist.ts` under `helpers/`; also replaced `Chart.tsx` code to reference the same statistic-calculating code in `gradeDist.ts`
- Included average GPA in course popover, displaying N/A when grade data is unavailable and "Loading..." when still computing

## Screenshots

image
image
image

## Test Plan

1. Hover over a course that has a long description and/or long prerequisite list, ensure that they display at most 4 lines before they cut off with '...'
2. Check that while "Average GPA" is being computed, it displays "Loading..."
3. Find a course without grade data (i.e. CS 130) and check that "Average GPA" displays N/A not NaN or something else
4. Look at both a course that has prerequisite warning and one that doesn't to ensure popover appearance looks good in both instances
5. Click into a course to double check that grade distribution charts are unaffected by refactor

## Issues

Closes #1075

### DIFF
--- a/site/src/component/CourseInfo/CourseInfo.tsx
+++ b/site/src/component/CourseInfo/CourseInfo.tsx
@@ -30,7 +30,7 @@ export const CourseBookmarkButton: FC<CourseProp> = ({ course, disabled = false 
 
 export const CourseSynopsis: FC<CourseProp> = ({ course, clampDescription = 0 }) => {
   return (
-    <p className="course-synopsis" style={clampDescription ? { WebkitLineClamp: clampDescription } : {}}>
+    <p className="course-info-clamp" style={clampDescription ? { WebkitLineClamp: clampDescription } : {}}>
       <b className="title">{course.title}</b>
       <span className="description">{course.description}</span>
     </p>
@@ -41,7 +41,7 @@ export const PrerequisiteText: FC<CourseProp> = ({ course, clampDescription = 0 
   if (!course.prerequisiteText) return <></>;
 
   return (
-    <p style={clampDescription ? { WebkitLineClamp: clampDescription } : {}}>
+    <p className="course-info-clamp" style={clampDescription ? { WebkitLineClamp: clampDescription } : {}}>
       <b>Prerequisites:</b> {course.prerequisiteText}
     </p>
   );
--- a/site/src/component/CourseInfo/CourseInfo.tsx
+++ b/site/src/component/CourseInfo/CourseInfo.tsx
@@ -9,9 +9,8 @@ import { IconButton } from '@mui/material';
 import BookmarkBorderIcon from '@mui/icons-material/BookmarkBorder';
 import BookmarkIcon from '@mui/icons-material/Bookmark';
 import WarningAmberIcon from '@mui/icons-material/WarningAmber';
-import { GradesRaw } from '@peterportal/types';
-import { fetchGradeDistDataRaw } from '../GradeDist/GradeDist';
-import { getAggregateClassGradeData } from '../GradeDist/Pie';
+import { fetchGradeDistData } from '../GradeDist/GradeDist';
+import { getAggregateGradeData, GradesAggregate } from '../../helpers/gradeDist.ts';
 
 interface CourseProp {
   course: CourseGQLData;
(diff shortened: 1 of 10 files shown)


site/src/component/CourseInfo/CourseInfo.tsx
```diff
@@ -35,11 +37,11 @@ export const CourseSynopsis: FC<CourseProp> = ({ course, clampDescription = 0 })
   );
 };
 
-export const PrerequisiteText: FC<CourseProp> = ({ course }) => {
+export const PrerequisiteText: FC<CourseProp> = ({ course, clampDescription = 0 }) => {
   if (!course.prerequisiteText) return <></>;
 
   return (
-    <p>
+    <p style={clampDescription ? { WebkitLineClamp: clampDescription } : {}}>
```
CadenLee2: Though this clamping works above, it doesn't work here because the `p` doesn't have the correct webkit styles (as opposed to the `.course-synopsis` above.

note that `.course-synopsis` has:
```
  display: -webkit-box;
  -webkit-box-orient: vertical;
  overflow: hidden;
  text-overflow: ellipsis;
```

You can consider even just using `.course-synopsis`, maybe renaming it
Image

CadenLee2 (commented):
Code changes look great and works well on staging!
- Tested popover loading
- Current grade dist functionality in course previews didn't break

One bug with clamping prereqs.

Also, after checking the database, we have 2.4k saved courses in prod, with a few users having a lot of saved courses. Imo if we remove the save button from the popover in order to prevent it from blocking hovers, it'll be okay for those who do save courses regularly because they'll probably be initially confused then try clicking on a course, where they'll find the feature.

Lmk your thoughts on this ^

CLiang531:
> Code changes look great and works well on staging!
> 
> * Tested popover loading
> * Current grade dist functionality in course previews didn't break
> 
> One bug with clamping prereqs.
> 
> Also, after checking the database, we have 2.4k saved courses in prod, with a few users having a lot of saved courses. Imo if we remove the save button from the popover in order to prevent it from blocking hovers, it'll be okay for those who do save courses regularly because they'll probably be initially confused then try clicking on a course, where they'll find the feature.
> 
> Lmk your thoughts on this ^

Yeah I think that makes sense since it'll be a lot more helpful for the majority of people who don't use saved courses to not be constantly blocked by the popover when trying to scan through them quickly. For the people who do have saved courses, I think that process also makes sense. Do you think it would be more helpful to make the save button be a little more visible though? Right now, it's next to the share button, which is labeled and has a lighter background while the bookmark is just a symbol next to it. I initially didn't see it when I was looking for it, so I was wondering if it would be more visible if the button matched the share button format. For people who don't know about the functionality, it could also be a way to expose them to it and let them know that the symbol is for saving, since I don't think it's specified elsewhere.

CadenLee2 (commented):
> Do you think it would be more helpful to make the save button be a little more visible though? Right now, it's next to the share button, which is labeled and has a lighter background while the bookmark is just a symbol next to it

I agree with that (styling the save button to look like the share button) The only concern is that on mobile devices or small screens it'll have to shrink so the header doesn't overflow (see image, that's not enough room for a full button. Can be done with something like a container query).

When the course is unsaved it can say "Save", but when the course is saved I'm not sure if we want it to say "Unsaved", "Saved", or something else.

I think this improvement is in the scope of this PR, so feel free to experiment with it and remove the bookmark from the popover

Image
Image

CadenLee2 (approved):
Tested some more, the issue mentioned in the meeting is fixed and everything else looks good! Saving courses works well, incl. on mobile

PR #1108, opened 2026-05-03
Revert mobile drag

CadenLee2:
We have gotten several reports of struggling to scroll on mobile, because users try to swipe but it ends up dragging the card instead. This PR reverts the change that caused this, so that now only the handle can be used to drag.

Drag enhancements are continuing and we can evaluate how to improve the mobile experience in the future (such as with whole-card drag but with a delay, as suggested in a meeting) but for now this is reversion is necessary.

## Description

Reverted commit d98b2e159b4df59127c481aff9c38c36b3375f7f.

## Screenshots

N/A (just view the interaction in staging)

## Test Plan

- [ ] Open mobile emulation and try to scroll on a large roadmap. Should work fine
- [ ] Drag the courses by the drag handle. Should continue to work

## Issues

Closes #

### DIFF
--- a/site/src/app/roadmap/planner/Course.tsx
+++ b/site/src/app/roadmap/planner/Course.tsx
@@ -112,10 +112,7 @@ const Course: FC<CourseProps> = (props) => {
   }, [userChosenUnits]);
 
   return (
-    <div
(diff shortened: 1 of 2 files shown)


anthonyj33 (approved):
LGTM

PR #1112, opened 2026-05-03
Quarter Mismatch Warning

### LINKED ISSUE #1105: Warning for quarter mismatch
Some classes are consistently offered in only certain quarters, and not frequently offered in others.

For instance, BIO SCI 9K is offered only in Spring. Users currently must check each course manually by hovering/clicking the course to ensure that it'll likely be offered. However, we can automatically inform them of a mismatch.

Mismatch detection could look like:
- If a course in the roadmap has not been offered in its quarter for the past (two?) years, it is considered "mismatched"
- A mismatched course can have a yellow warning icon similar to the red prerequisite one
- Hovering that icon shows the schedule of classes and says something like "Typically not offered in Winter"

The offered terms can be found from CourseGQLData, so you shouldn't have to do any additional API calls

Image

Image

CLiang531:
## Description

- Added orange (mui-palette-warning-main) warning symbol next to course cards in roadmap if that course has not been offerred in that quarter of the planner year and also for the previous two years
- The computation to determine mismatch starts on the planner year, but if the planner year it is set to the future, it takes the current school year. This is so that past taken courses aren't looking at most recent offerings and incorrectly displaying a warning while courses planned in the future are considering recent offerings and not future nonexistent data.
- Course popover also displays warning under offerings table with text `Typically not offered in [quarter]`

## Screenshots

image
image
image

## Test Plan

1. Add a course that is not offerred for the current academic year and the two previous academic years (for that quarter) to the roadmap. Verify that it displays an orange warning symbol next to the title (to the right of the red prerequisite warning if it exists).
2. Hover over the added course and check that the warning with `Typically not offered in [quarter]` text displays under the offerings table
3. Play around with the start year to ensure that even if it is set far into the future it looks at recent offerings for warning display, and if set into the past (when it used to be offerred in that quarter or was offerred a year or two back starting from that year) it correctly references the past offerings.

## Issues

Closes #1105

### DIFF
--- a/site/src/app/roadmap/planner/Quarter.tsx
+++ b/site/src/app/roadmap/planner/Quarter.tsx
@@ -207,10 +207,15 @@ const Quarter: FC<QuarterProps> = ({ yearIndex, quarterIndex, data }) => {
           });
 
           const searchPrevYears = 2;
-          const termYear = quarterIndex === 0 ? startYear - 1 : startYear;
+          const currentDate = new Date();
+          const currentStartYear =
+            currentDate.getMonth() >= 8 ? currentDate.getFullYear() : currentDate.getFullYear() - 1;
+          const useYear = Math.min(currentStartYear, startYear);
+          const termYear = quarterIndex === 0 ? useYear : useYear + 1;
           let mismatch = true;
-          for (let i = 0; i < searchPrevYears; i++) {
+          for (let i = 0; i <= searchPrevYears; i++) {
             const term = `${termYear - i} ${quarterTitle}`;
+            console.log(term);
             if (course.terms.includes(term)) {
               mismatch = false;
               break;
--- a/site/src/app/roadmap/planner/Quarter.tsx
+++ b/site/src/app/roadmap/planner/Quarter.tsx
@@ -211,11 +211,10 @@ const Quarter: FC<QuarterProps> = ({ yearIndex, quarterIndex, data }) => {
           const currentStartYear =
             currentDate.getMonth() >= 8 ? currentDate.getFullYear() : currentDate.getFullYear() - 1;
           const useYear = Math.min(currentStartYear, startYear);
-          const termYear = quarterIndex === 0 ? useYear : useYear + 1;
+          const termYear = data.name === 'Fall' ? useYear : useYear + 1;
           let mismatch = true;
           for (let i = 0; i <= searchPrevYears; i++) {
-            const term = `${termYear - i} ${quarterTitle}`;
-            console.log(term);
+            const term = `${termYear - i} ${data.name}`;
             if (course.terms.includes(term)) {
               mismatch = false;
               break;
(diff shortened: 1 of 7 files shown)


site/src/app/roadmap/planner/Quarter.tsx
```diff
@@ -206,6 +206,18 @@ const Quarter: FC<QuarterProps> = ({ yearIndex, quarterIndex, data }) => {
             }
           });
 
+          const searchPrevYears = 2;
+          const termYear = quarterIndex === 0 ? startYear - 1 : startYear;
```
CadenLee2: `quarterIndex` may be 0 for a year where the user has unchecked Fall (note that it's created in `Year.tsx` and is independent from the recent offerings `quarterIndexMap`). We should instead compare the quarter name here

site/src/app/roadmap/planner/Quarter.tsx
```diff
@@ -206,6 +206,18 @@ const Quarter: FC<QuarterProps> = ({ yearIndex, quarterIndex, data }) => {
...
+          let mismatch = true;
+          for (let i = 0; i < searchPrevYears; i++) {
+            const term = `${termYear - i} ${quarterTitle}`;
```
CadenLee2: Summer seems to fail because `quarterTitle` does not match the actual name returned by API for summer quarters:
Image
`quarterTitle` is set above in this file and looks like `Summer 10 Week`:
```ts
const quarterTitle = quarterDisplayNames[data.name];
```
So we should use `data.name` instead, which matches the API response of `Summer10wk`

CadenLee2 (commented):
Works well and looks good; I found a few small bugs and have one suggestion

I'm thinking it could be a bit more clear for users if the mismatch warning showed up above the recent offerings table / below the prereqs warning, because that way your eyes will immediately be led down into the list of recent offerings. Lmk what you think
Image

CLiang531:
> I'm thinking it could be a bit more clear for users if the mismatch warning showed up above the recent offerings table / below the prereqs warning, because that way your eyes will immediately be led down into the list of recent offerings. Lmk what you think

I thought about that placement too, but I feel like since prerequisite's warning is underneath the prerequisite text itself it might be inconsistent to then move this warning up. Since we're planning on including Average GPA as well in my other PR, it might even look like that warning is about the GPA instead. We can ask about this in the next meeting to see what other people think.

Also I just realized I did not push my commit for the most up-to-date computation of the year but now it should actually reflect this part of the PR description + the comparing with quarter name edit you suggested
>The computation to determine mismatch starts on the planner year, but if the planner year it is set to the future, it takes the current school year. This is so that past taken courses aren't looking at most recent offerings and incorrectly displaying a warning while courses planned in the future are considering recent offerings and not future nonexistent data.

CadenLee2 (approved):
Tested summer quarters and re-tested the original functionality, the fix works well

> I feel like since prerequisite's warning is underneath the prerequisite text itself it might be inconsistent to then move this warning up

Makes sense, as discussed in meeting I'm good moving ahead with how it is currently

PR #1114, opened 2026-05-04
fixed broken popover corequisite checks

### LINKED ISSUE #1059: Broken popover corequisite checks
Valid corequisites work fine with on the roadmap (i.e. red triangle doesn't show up), but the popover still treats them as prerequisites
Image

charlieweinberger:
## Description

Previously, corequisite courses would show an error saying the prerequisite wasn't met, even if the paired course was currently being taken in the same quarter (as corequisites are). Previously, only past courses taken were considered. Now, the current quarter's classes are also considered.

## Screenshots

| Before | After |
| --- | --- |
| Screenshot 2026-05-04 at 12 06 41 PM | Screenshot 2026-05-04 at 12 06 19 PM |

## Test Plan

- [ ] Make sure corequisite error bug is fixed
- [ ] Make sure existing prerequisite checking still works properly

## Issues

Closes #1059


CadenLee2 (approved):
Fix works, tested with PHYSICS 7C and with typical prerequisites

PR #1115, opened 2026-05-05
Gracefully handle intl failure

CadenLee2:
For reasons whose steps to replicate are currently unknown (will be investigated once the reporting user is available), a non-finite value is passed into `Intl.RelativeTimeFormat`, causing the entire app to crash due to a client-side error.

## Description

Handle non-finite numbers

## Test Plan

- [ ] Site doesn't break. Steps to replicate the original issue will be made available when they are found

## Issues

Closes #


anthonyj33 (approved):
LGTM and we can address the root cause of this soon

PR #1119, opened 2026-05-06
fix major specialization dropdown not loading properly

### LINKED ISSUE #1095: Specialization dropdown does not load properly
Steps to replicate:
1. Choose a major with a specialization (while logged in)
2. Choose a specialization
3. Reload the page. The specialization requirements will show up properly, but the dropdown will not load the correct result, and instead will be blank.

This is likely due to the MUI Autocomplete's value not updating when its options update after they load. A similar issue was faced in #1007 and was resolved by providing a default value, but in this case the solution may not be that simple

Image

ellisl10:
## Description
The major specialization dropdown does not properly load. It shows the requirements, but not the name. 

## Screenshots

Before
Screenshot 2026-05-06 at 1 37 15 PM

## Test Plan

1. Select a major with specialization
2. Reload the page
3. The specialization requirements and name should be properly shown

## Issues

Closes #1095

### DIFF
--- a/site/src/app/roadmap/catalog/MajorCourseList.tsx
+++ b/site/src/app/roadmap/catalog/MajorCourseList.tsx
@@ -140,6 +140,9 @@ const MajorCourseList: FC<MajorCourseListProps> = ({ majorWithSpec, onSpecializa
   );
 
   const toggleExpand = () => setOpen(!open);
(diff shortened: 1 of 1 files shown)


CadenLee2 (approved):
LGTM

PR #1118, opened 2026-05-05
Convert native inputs to MUI

fiveminus1:
## Description

Converts native `<input>` fields to use MUI components.

## Screenshots

Before|After
:-:|:-:
image|image
image|image
image|image

## Test Plan

- [ ] Test all modified inputs, functionality should be same 
- [ ] Styling looks good on mobile/desktop

## Issues

Closes #1074

PR #1120, opened 2026-05-06
Custom card improvements

CoreyZhong:
Reformatted custom cards to match AP style editing and improved visibility
## Description

- 

## Screenshots
Before:
image
After:
image
Edit Mode:
image

## Issues

Closes #

### DIFF
--- a/site/src/app/roadmap/catalog/CustomCourseCard.scss
+++ b/site/src/app/roadmap/catalog/CustomCourseCard.scss
@@ -1,9 +1,8 @@
 .custom-card {
-  box-shadow: 0 0 2px 0 #60616680;
   background: var(--mui-palette-overlay-overlay1);
-  border: 1px solid var(--border-secondary);
-  margin-left: -1px;
-  margin-right: -1px;
+  border: 2px solid var(--mui-palette-divider);
+  margin-left: -2px;
+  margin-right: -2px;
   height: 73px;
   font-size: 16px;
   padding: 12px;
--- a/site/src/app/roadmap/catalog/CustomCourseCard.scss
+++ b/site/src/app/roadmap/catalog/CustomCourseCard.scss
@@ -108,4 +108,8 @@
       font-size: 20px;
     }
   }
+
+  .course-edit-btn.MuiIconButton-root .MuiSvgIcon-root {
+    font-size: 18px;
+  }
 }
--- a/site/src/app/roadmap/catalog/CustomCourseCard.scss
+++ b/site/src/app/roadmap/catalog/CustomCourseCard.scss
@@ -9,13 +9,9 @@
   position: relative;
   color: var(--mui-palette-secondary-main);
 
-  display: grid;
-  grid-template-areas:
-    'A B'
-    'A C';
-  grid-template-columns: min-content auto;
-  grid-template-rows: 1fr 1fr;
-  align-items: center;
+  display: flex;
+  flex-direction: row;
+  align-items: stretch;
 
   .title {
     font-size: 14px;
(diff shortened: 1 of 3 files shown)


site/src/app/roadmap/catalog/CustomCourseCard.scss
```diff
@@ -43,6 +35,17 @@
     }
   }
 
+  &__main {
+    flex: 1 1 auto;
+    min-width: 0;
+    display: grid;
+    grid-template-areas:
```
CadenLee2, line 42: When it comes to CSS conventions, we don't typically use this ([BEM](https://css-tricks.com/bem-101/)) convention. Though it has some benefits, for consistency the CSS class names should match how it's done in other files.

For instance, because we already have this wrapped in `.custom-card`, the `&__main` (producing `.custom-card__main`) is not necessary. We can safely just use `.main` (allowing the className to become `"main"`).

The same applies throughout this file's class names

site/src/app/roadmap/catalog/CustomCourseCard.scss
```diff
@@ -77,6 +84,17 @@
...
 
+  .units-input[type='number']::-webkit-outer-spin-button,
+  .units-input[type='number']::-webkit-inner-spin-button {
+    -webkit-appearance: none;
+    margin: 0;
+  }
+
+  .units-input[type='number'] {
+    appearance: textfield;
+    -moz-appearance: textfield;
```
CadenLee2: 1. Note that once #1118 gets merged, this will no longer be necessary. There may be a small merge conflict or there may not be, but this should be removed once that happens (assuming 1118 gets merged first)
2. `webkit-outer-spin-button` is [non-standard](https://devdoc.net/web/developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-outer-spin-button.html). Was there a specific reason for including this rule?

site/src/app/roadmap/catalog/CustomCourseCard.tsx
```diff
@@ -21,23 +23,41 @@ export const CustomCourseCard: FC<CustomCourseCardProps> = ({ course, handleUpda
...
+  useEffect(() => {
+    if (inRoadmap || editing) return;
+    setNewName(course.courseName);
+    setNewUnits(course.units);
+    setNewDescription(course.description);
+  }, [course.courseName, course.units, course.description, course.id, editing, inRoadmap]);
+
+  const onDelete = useCallback(
+    async (e?: React.MouseEvent) => {
+      e?.stopPropagation();
```
CadenLee2: What's the reason for stopping propagation here?

CadenLee2 (commented):
The new interaction works and looks a lot better!

I have some code suggestions, and 2 overall suggestions:
1. When the user presses the "new custom card" button, ideally the card should be in the edit mode by default because it may look unclear/like a bug when it's just completely empty. This could be challenging, but one possible (kinda funny but it would work) solution is to make any card with an empty title be in edit mode by default
2. Based on some discussion from meetings, we should change the placeholder that says "Course" to say something more generic like "Title" so users understand that it can be used for anything
Image
Image

CoreyZhong:
> The new interaction works and looks a lot better!
> 
> I have some code suggestions, and 2 overall suggestions:
> 
> 1. When the user presses the "new custom card" button, ideally the card should be in the edit mode by default because it may look unclear/like a bug when it's just completely empty. This could be challenging, but one possible (kinda funny but it would work) solution is to make any card with an empty title be in edit mode by default
> 2. Based on some discussion from meetings, we should change the placeholder that says "Course" to say something more generic like "Title" so users understand that it can be used for anything
> 
> Image Image

I added the change where empty titles are in edit mode by default. Only thing is if titles were intentionally left empty, they would show in edit mode on refresh which might look weird. Although this is a minor issue that probably won't affect many people

CadenLee2 (approved):
LGTM!

As for cards with an empty title, as mentioned in meetings that'll be fine as users are expected to provide a title anyway

Issue #1008, opened 2026-02-27
Ability to store notes with courses/bookmarks


Awesome-E:
**What's a feature you would like to see?**
It would be helpful to be able to store notes with a certain course for scenarios like "want to take with a certain prof", "taking for GE III", or "taking instead of CS 171"

**Does this require data from the school, i.e. grade distributions, that Anteater API does not currently provide?**
No

**How should it look like or function for the user?**
I'm thinking this can be associated with Bookmarks/Saved Courses to some degree, perhaps a flow that's similar to adding user notes on Discord, but with Course Bookmarks. If associated with bookmarks, users would ideally be able to see all of their notes from the saved courses menu as well

A couple of things this would require:
- Thinking about how this interacts with the popover -- where would the text entry go
- How would the note show in the saved courses menu on desktop and on mobile?

**Describe alternatives you've considered**
marking notes directly on the roadmap, though this feels like it might need more UI around it and is also not persistent across roadmaps.

**Additional context**
N/A

PR #1122, opened 2026-05-09
Course notes endpoints

JasonNguyen067:
## Description

get, add, edit, and delete for the Coursenotes backend.

## Test Plan

I tested the add and delete via a mock query. Changes to the table were reflected in the drizzle. 

## Issues

Part 2 of #1008 (need to wait for frontend designs)

### DIFF
--- a/api/src/controllers/courseNotes.ts
+++ b/api/src/controllers/courseNotes.ts
@@ -1,0 +1,73 @@
+import { router, userProcedure } from '../helpers/trpc';
+import { z } from 'zod';
+import { TRPCError } from '@trpc/server';
+import { db } from '../db';
+import { courseNotes } from '../db/schema';
+import { and, eq } from 'drizzle-orm';
+
+export const courseNotesRouter = router({
+  get: userProcedure.input(z.object({ courseId: z.string() })).query(async ({ ctx, input }) => {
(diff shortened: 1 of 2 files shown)


api/src/controllers/courseNotes.ts
```diff
@@ -0,0 +1,73 @@
...
+        ).length
+      ) {
+        throw new TRPCError({ code: 'NOT_FOUND', message: 'No notes found for this course' });
+      }
+
+      await db
+        .update(courseNotes)
+        .set({ content: content })
+        .where(and(eq(courseNotes.userId, userId!), eq(courseNotes.courseId, courseId)));
+    }),
```
CadenLee2: This pattern (also reused below) makes two trips to the database, meaning it'll take twice as long to load.

Try to condense this logic into one database statement: consider just keeping the bottom statement, but adding [.returning](https://orm.drizzle.team/docs/update#returning) to return the updated row. You can test and see what is returned or thrown when no row exists, then make sure you're handling that case properly

CadenLee2 (commented):
Looks good, just one comment on performance for `edit` and `delete`

CadenLee2 (approved):
Tested out the change and the 404 is returned in the correct circumstances. Everything else looks good!

PR #1124, opened 2026-05-10
Add new tooltip for courses not offered in recent years

### LINKED ISSUE #1123: Better course offering tooltip for courses not offered in recent years
Currently, when a course hasn't been offered in a particular quarter (e.g., Fall), the tooltip displays a warning like "Typically not offered in Fall." However, this message can be misleading when the course hasn't been offered at all in recent years. See image below for more details.

Image

This should instead have a different tooltip saying "Not offered in recent years" in a slightly darker shade of orange/closer to red.

training4usaco:
## Before this PR
The quarterMismatch tooltip would display when a course hasn't been offered for a while. This is misleading if the course was previously offered in a specific quarter. This is shown in more detail in [this issue](https://github.com/icssc/peterportal-client/issues/1123).

## After this PR
- Added a check in Quarter.tsx to see if the course hasn't been offered in any quarters in the past `searchPrevYears` years. 
- Changed quarterMismatch to termMismatch and made it either a `QuarterName` for quarter mismatch or string `'RecentYears'` for courses not offered recently.
- This is then parsed in TermMismatchText to display the new tooltip when appropriate.

## Screenshots
Misleading tooltip before PR:
image

Example of new tooltip after PR:
image

## Test Plan
- Add courses that are not offered in recent years to ensure that the new tooltip is showing correctly (writing 39b, compsci 113).
- Add courses that actually have quarter mismatch to ensure the old tooltip is showing correctly (compsci 116 in spring, econ 105c in fall).
- Add courses that have been offered in the appropriate quarters to ensure no tooltips are showing (ics 33 in any quarter).

## FLUPs
While testing, I noticed that if the user has not satisfied the prereqs for a class and one of the course availability tooltips also applies, it will show both. While this is not the biggest issue, it seems rather clunky (see image below).
image

## Issues
Closes #1123

### DIFF
--- a/site/src/app/roadmap/planner/Course.scss
+++ b/site/src/app/roadmap/planner/Course.scss
@@ -46,6 +46,10 @@
     color: var(--mui-palette-warning-main);
   }
 
+  .warning-tertiary {
+    color: var(--mui-palette-warning-dark);
+  }
+
   .course-card-top {
     display: flex;
     justify-content: space-between;
--- a/site/src/app/roadmap/planner/Course.scss
+++ b/site/src/app/roadmap/planner/Course.scss
@@ -46,10 +46,6 @@
     color: var(--mui-palette-warning-main);
   }
 
-  .warning-tertiary {
-    color: var(--mui-palette-warning-dark);
-  }
-
   .course-card-top {
     display: flex;
     justify-content: space-between;
--- a/site/src/app/roadmap/planner/Quarter.tsx
+++ b/site/src/app/roadmap/planner/Quarter.tsx
@@ -228,7 +228,7 @@ const Quarter: FC<QuarterProps> = ({ yearIndex, quarterIndex, data }) => {
               }
               return false;
             });
-          const quarterMismatch = mismatch && !notRecentlyOffered ? quarterTitle : undefined;
+          const termMismatch = mismatch ? (notRecentlyOffered ? 'RecentYears' : data.name) : undefined;
 
           return (
             // addMode="drag" somehow fixes the issue with tapping a course after adding on mobile
@@ -240,8 +240,7 @@ const Quarter: FC<QuarterProps> = ({ yearIndex, quarterIndex, data }) => {
                 if (revision.edits.length > 0) dispatch(reviseRoadmap(revision));
               }}
               requiredCourses={requiredCourses}
-              quarterMismatch={quarterMismatch}
-              notRecentlyOffered={notRecentlyOffered}
+              termMismatch={termMismatch}
               onDelete={() => removeCourseAt(index)}
               addMode="drag"
               openPopoverLeft
--- a/site/src/app/roadmap/planner/Quarter.tsx
+++ b/site/src/app/roadmap/planner/Quarter.tsx
@@ -220,7 +220,15 @@ const Quarter: FC<QuarterProps> = ({ yearIndex, quarterIndex, data }) => {
               break;
             }
           }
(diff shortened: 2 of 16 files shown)


training4usaco:
@CadenLee2 Hi! Could I get a review for this PR.

site/src/app/roadmap/planner/Course.scss
```diff
@@ -46,6 +46,10 @@
     color: var(--mui-palette-warning-main);
   }
 
+  .warning-tertiary {
+    color: var(--mui-palette-warning-dark);
+  }
+
```
CadenLee2: Design-wise, these warnings both indicate the same level of concern for the user (the course is likely not offered when the user wants it to be), so they should be the same color.

From the user's perspective, the only thing that's changing here is the warning message itself ("typically not offered in" vs "not offered in recent years")

site/src/app/roadmap/planner/Quarter.tsx
```diff
@@ -220,7 +220,15 @@ const Quarter: FC<QuarterProps> = ({ yearIndex, quarterIndex, data }) => {
...
-          const quarterMismatch = mismatch ? quarterTitle : undefined;
+          const notRecentlyOffered =
+            mismatch &&
+            !course.terms.some((term) => {
+              for (let i = 0; i <= searchPrevYears; i++) {
+                if (term.startsWith(`${termYear - i} `)) return true;
+              }
+              return false;
+            });
+          const quarterMismatch = mismatch && !notRecentlyOffered ? quarterTitle : undefined;
```
CadenLee2: Rather than needing to pass two props `notRecentlyOffered` and `quarterMismatch` into the Course, since the two perform the same role (see color comment above), we can actually condense them into just `quarterMismatch`.

The idea is:
- Change the type of quarterMismatch to `QuarterName | 'RecentYears'`
- In this line, instead of setting quarterMismatch to the `quarterTitle` (which is a display title like "Summer 1"), use the `data.name` (which is of type `QuarterName`).

Then we can:
- Remove `notRecentlyOffered` props below/elsewhere, and remove `NotRecentlyOfferedText`
- In `QuarterMismatchText`, we can import the helper `quarterDisplayNames` and use `quarterDisplayNames[quarterMismatch]` to convert it to a display quarter title. If quarterMismatch is insead `'RecentYears'`, we can change the whole text to say "Not offered in recent years"

This helps minimize the code changes and also strengthens the type safety of quarterMismatch.

If you do feel that quarterMismatch is now a misleading name, you can consider renaming it to something more general like offeringMismatch or termMismatch (but I don't have a strong opinion on this)

CadenLee2 (commented):
I tested out your changes, and everything works as expected! This is a big improvement for those classes like WRITING 39B.

1. About your comment on both warnings showing up at once, I think this is acceptable because ideally a user's long-term roadmap should have no warnings, so that's a sign that something needs to change
2. I have one comment about the visuals (see code comment)
3. I have one comment about the code (see code comment). This is a slightly larger restructure but should significantly simplify things, lmk if you have questions

Thanks for taking the time to work on the feature and address the feedback!

training4usaco:
Just finished making the recommended changes! (also not sure if you want me to resolve previous comments or not so sorry about that). I removed the unnecessary style that I added and changed quarterMismatch to termMismatch and made it type QuarterName | 'RecentYears'.

CadenLee2 (approved):
Changes look good and the functionality continues to work! This is good to be merged

PR #1126, opened 2026-05-10
feat: support Sign in with Apple, add accounts table 

KevinWu098:
## Summary

Adds Apple Sign In support and introduces an `account` table to properly store multiple auth identities per user.

### Schema

- New `account` table with `(user_id, provider)` PK, `provider_account_id` (unique), `created_at`, `updated_at`. `provider` is a Postgres enum (`GOOGLE`, `APPLE`).
- Migration backfills all existing `user.google_id` values into `account` as `GOOGLE` provider rows.
- `google_id` column kept on `user` for now (compat writes in `successLogin`; will be dropped in a followup PR).

### Auth

- `successLogin` rewritten: finds or creates user by email, upserts an `account` row. Profile data (name/picture) only overwritten when incoming values are non-empty, preventing Apple's sparser profile from clobbering Google data.
- `providerFromSub` helper derives `GOOGLE`/`APPLE` from the OIDC `sub` prefix.
- `GET /auth/google` now accepts `?provider=google|apple` and forwards it to `icssc/auth`.

### External API

- `external.roadmaps.getByGoogleID` queries the `account` table by `provider_account_id` instead of `user.google_id`.

### UI

- "Sign In" replaced with two buttons ("Sign In with Google" / "Sign In with Apple") in `ProfileMenuButtons` and `Profile` `AuthButton`.
- Silent SSO (`AutoSignIn`) unchanged -- `prompt=none` is provider-agnostic.

### Depends on

- [icssc/auth Apple Sign In support](https://github.com/icssc/auth) (server-side Apple callback + `?provider=apple` on `/authorize`)
- [AntAlmanac Apple Sign In (PR #1744)](https://github.com/icssc/AntAlmanac/pull/1744)

## Test Plan

- [ ] Sign in with Google -- existing flow unchanged
- [ ] Sign in with Apple -- new account row created with `APPLE` provider
- [ ] Sign in with Apple using same email as existing Google account -- same user, two account rows, Google profile preserved
- [ ] Silent SSO (`prompt=none`) works for returning users
- [ ] External roadmap API still resolves users by provider ID
- [ ] Migration applies cleanly (backfill populates `account` from `google_id`)


api/drizzle/meta/0030_snapshot.json
```diff
@@ -1,5 +1,5 @@
 {
-  "id": "58a55613-eb11-4de3-97c2-b464cf9f637c",
```
KevinWu098, line 2: note: it appears that 0030 had the same id as 0027, but they had different contents. my guess is that this is the result of a bad merge conflict, but it is breaking because drizzle will throw on a non-sequential snapshot chain.

the ids themselves aren't important, so ive updated 0030 to resolve this issue

anthonyj33 (commented):
Having an issue w/ saving:
- Sign in with Apple
- sign out
- sign in with Google using **same email** so they are linked 
- change something
- save once —> fails
- save again —> doesn’t fail (at least doesn’t say it did)

===

Also:
- The Sign in with Apple prompt says sign in to "auth", can we change that to ICSSC?
- It might just be my account but profile picture is blank when signing in with Apple

KevinWu098:
> Having an issue w/ saving:
> 
> * Sign in with Apple
> * sign out
> * sign in with Google using **same email** so they are linked
> * change something
> * save once —> fails
> * save again —> doesn’t fail (at least doesn’t say it did)
> 
> ===
> 
> Also:
> 
> * The Sign in with Apple prompt says sign in to "auth", can we change that to ICSSC?
> * It might just be my account but profile picture is blank when signing in with Apple

1. can you repro?
2. updated
3. this is because apple doesn't provide profiles

CadenLee2 (approved):
Approving as we are unable to reproduce the saving error, and all other feedback is addressed

PR #1133, opened 2026-05-13
Animate Roadmap Height

Ascent817:
## Description

Add a new hook to smoothly animate roadmap heights when classes are dragged around, changing its height.


anthonyj33 (approved):
LGTM!

PR #1134, opened 2026-05-13
Add always active filters

### LINKED ISSUE #1110: Always show search filters
## Motivation
Currently, search filters only appear when the user has searched a query that has and is showing course results. This is because the search filters only apply to courses.

However, this creates significant frustrating or misleading interactions:
- Users may be unaware that search filters exist
- When changing their query, users often end up losing their search filters. Our custom logic is not good enough to support this (e.g. someone wants lower-div math courses, then changes to ics and loses the lower-div constraint)
Image

We can instead persist the search filters and simply indicate whether they are taking effect. Possible momentary confusion in the case of professors is worth heavily increased usability.

## Logic
1. When the user has no search:
    - Show the filters
    - The filters are changeable, but changing them does nothing until they search.
2. When the user has a search and *course results are being shown*:
    - Show the filters
    - Changing the filters re-runs the search completely
3. When the user has a search and *instructor results are being shown*:
    - Show the filters, but *grayed out* because they are not taking effect
    - Changing the filters re-runs the search completely
3. When the user has a search and *no results were found*:
    - Show the filters
    - Changing the filters re-runs the search completely

Our logic for auto-determining whether to show course or professor results should persist.

Image

## Design justification
It makes sense that changing the filters re-runs the search completely every time, because the filters are now paired with the query rather than with the result type. This is indicated by the result type now being *below* the filters, as it will re-generate every time.

Ascent817:
# Description

Always show search filters on desktop and mobile, with the results header moved below them.
Add a disabled state and helper message when instructor results are active.
Re-run both course and instructor searches on filter changes, recompute the view index, and avoid firing searches when the query is empty.

# Screenshots

image
image
image

# Test Plan

Manual: With no query, verify filters are visible and selectable without triggering a search.
Search for a query that yields instructor results first; confirm filters are disabled with message.
Change filters while on course results and confirm both result types refresh and the view index updates.
Verify the header appears below filters and padding looks correct when no header is shown.

# Issues
Closes #1110


CadenLee2 (commented):
This is a much better interaction!

Styling/behavior things I noticed:

1. Odd spacing here
Image

2. I like the disabled opacity, but what if we have it styled as "disabled" without actually disabling user interaction? Users might get frustrated being unable to change the filters on the instructors screen, even if it does nothing. See 3 for a case of this

3. You can lock the results into only instructors by changing the filters:
    - Search "math"
    - Change dept filter to something not matching (like AC ENG)
    - Because there are no resulting courses, only instructor results show up. You can't currently change the filters from there (see 2 above), meaning the only way to go back is to delete the query. I think that first we should update 2), then also maybe add an explainer in this situation like "No matching courses; showing instructors".
Image

Ascent817:
@CadenLee2 I've addressed your comments, let me know how it looks!

CadenLee2 (commented):
Looking better, though I did find a bug:
- Select a filter with no query in the box (e.g. choose LowerDiv)
- Enter a query (e.g. "math")
- The filter got reset when it shouldn't have

The same thing happens when you change the query to something substantially different like "math" to "klefstad". It's possibly related to some old logic? Maybe check where `setCourseFilters` is being called

Ascent817:
There's a ton of old logic I removed that I believe is no longer relevant, and I fixed the issue with filters resetting. My changes are a little aggressive, so let me know if you want me to keep more of the old behavior here.

site/src/hooks/search.ts
CadenLee2: Lint warning (see the other 2 as well; try reinstalling eslint and seeing if you're able to see this in your editor, or try `pnpm lint` and make sure you're able to see the error)

site/src/store/slices/searchSlice.ts
```diff
@@ -82,6 +74,11 @@ export const searchSlice = createSlice({
         ...action.payload.results,
       ] as (typeof state)[typeof index]['results'];
     },
+    // A search that failed (network error, API error — not an abort) must clear the in-progress
```
CadenLee2: Funny em dash lol, consider all 3 of these sets of comments and see if they're necessary/if you can clean them up a bit. I'm thinking the abort error one for instance is definitely helpful, but can be simplified to something like "Aborts are expected when a newer search supersedes this one"

site/src/component/SearchFilters/SearchFilters.scss
```diff
@@ -5,6 +5,22 @@
   flex-wrap: wrap;
   gap: 8px;
 
+  &.has-top-padding {
+    padding-top: 8px;
+  }
+
+  // Dim the controls to signal filters apply to course (not the shown instructor) results, while
+  // keeping them interactive so users can still change filters from the instructor view.
```
CadenLee2: Unnecessary comment

CadenLee2 (commented):
Everything works smoothly while testing, glad we were able to get rid of a lot of the complex old logic!

Code changes are mostly good, just saw some lint warnings and had a few quick thoughts on code comments

site/src/hooks/search.ts
Ascent817: Yep for some reason the linter wasn't running on commit, running it manually picked up the issues

site/src/store/slices/searchSlice.ts
Ascent817: Makes sense, I rewrote them to be terser

site/src/component/SearchFilters/SearchFilters.scss
Ascent817: Removed

CadenLee2 (approved):
Ran through the same tests and and looked through the code again, everything seems good.

This'll be great to get merged!

PR #1142, opened 2026-05-15
Group missing prerequisites

### LINKED ISSUE #590: Store Missing Prerequisites in Groups
Right now, missing prerequisites are stored as just strings of the courses that need to be completed. This does not account for "OR" cases, where only one of them need to be completed, i.e ICS46 OR CSE46. It currently will say "ICS46, CSE46" in missing prerequisite texts, but there should be some indication that only one actually needs to be completed.

training4usaco:
## Description
- Previously, every prerequisite would be listed in the course pop over, but there was no clear way to tell when one of two classes could be taken to satisfy the prereq (like an OR clause)
- Now, prerequisites are grouped with parenthesis, with an or separating courses within a "prereq group."
- `validateOrPrerequisites` now returns a set of a single string, containing a concatenation of the course ids separated by "|". 
- This is then later parsed in `IncompletePrerequisiteText`

## Screenshots
Before:
image
After:
image

## Test Plan
- Test courses with one group of prereqs (ics 33)
- Test courses with multiple groups of prereqs (compsci 117)
- Test courses without grouped prereqs (ics 32)

## Issues

Closes #590


site/src/helpers/planner.ts
```diff
@@ -508,16 +508,17 @@ const validateAndPrerequisite = ({ prerequisite, ...input }: ValidationInput<Pre
...
-    missing.forEach((course) => required.add(course));
+    missing.forEach((course) => required.push(course));
   }
 
-  return required;
+  const unique = [...new Set(required)];
+  return new Set([unique.join('|')]);
 };
 
 /**
```
CadenLee2, line 524: Because we are now returning strings that aren't necessarily valid course IDs (e.g. `MATH 3A|I&C SCI 6N`), we should indicate that in this JSDoc.

It says `@returns A set of all the prerequisites that are missing`, but this is misleading as a reader may think a "prerequisite" is always one course. This should be updated to `@returns A set of all missing prerequisite strings, with "or" groups separated by '|'`

CadenLee2 (commented):
After testing with several courses (ICS 33 prereqs, CS 117), I can confirm:
- Prerequisite checking continues to work properly
- The new or display functionality works properly

Codewise, this change works. However, I do recommend updating some documentation. See code comment

CadenLee2 (approved):
Testing continues to work, and updated documentation looks reasonable

PR #1149, opened 2026-05-20
Group Courses by Department

### LINKED ISSUE #1106: Group GEs by department
In large blocks of GE requirements (like the image below), users get lost and can't really make use of the list. For people who might want to look by department, the list should be grouped by department.

Each department will be collapsed by default, but can be expanded in a similar way to other requirement groups.

Very scuffed concept example, don't follow this exactly but use it as a reference for the general idea. **Note**: see the requirement text: we want to make sure users don't get confused and think they have to complete three entire departments. Maybe we change "Complete 3 of the following" in this case to "Complete 3 courses" or something more clear, this can be discussed
Image

When developing, make sure to work with the existing requirements components and structure.
- Ideally there should be an elegant way of determining whether to group or not (perhaps just pass down a prop, but maybe even just check for some combination of the length/number of items and variety of departments).
- Example of something that should be grouped:
Image

- If you're determining whether to group entirely programmatically across all requirements, consider that the following (CS list of electives) probably shouldn't be grouped by department because there are only a few departments anyway:

Image

MajdKatabi:
## Issues

Closes #1106


CadenLee2 (commented):
Reminder of our discussion in meeting
- [x] Border radius (saw this is done, looks good)
- [ ] Logic to unify the departments with H for honors, and probably M for (? forgor) as well. Remember that this logic only applies if both the `dept` and `deptH` departments exist (e.g. `I&CSCI` and `I&CSCIH` both exist -> collapsed into `I&CSCI`)

I also notice that while in dark theme the opened sections have a clear background, in light theme it seems that color matches the sidebar's background color and it becomes invisible. Maybe consider a different gray? Not sure what I think about it
Image

site/src/app/roadmap/catalog/ProgramRequirementsList.tsx
```diff
@@ -46,6 +46,10 @@ import MenuTile from '../transfers/MenuTile';
 import { setShowMobileCreditsMenu } from '../../../store/slices/transferCreditsSlice';
 import ClickableDiv from '../../../component/ClickableDiv/ClickableDiv';
 
+const DEPARTMENT_GROUPING_COURSE_THRESHOLD = 30;
+const COURSE_ID_DEPARTMENT_OVERRIDES = ['IN4MATX'];
+const DEPARTMENT_MERGE_SUFFIXES = ['H', 'M'];
```
CadenLee2, line 51: Fyi I recently learned there's also "E" (as in "BIO SCI E198"). However, we may not want to add it because there are departments that are differentiated (without spaces) purely by the E.

E.g. "ENGR CE" vs "ENGRCEE" would be merged when they shouldn't be.

Image

So the current merge suffix list shouldn't change, but I'm adding this comment just so we're aware

CadenLee2 (approved):
Tested the theming, and the grouping in several places (GEs, CS requirements). The code changes look good!

PR #1150, opened 2026-05-20
Preview Navbar Active Section Small Courses

### LINKED ISSUE #1137: Broken navbar on small courses
Steps to recreate:
1. Go to a small class like [CSE 104](https://antalmanac.com/planner?course=CSE104). This class has not been offered lately and has no reviews,
2. Click on the reviews section in the navbar. It will take you down to reviews, but will not change color

Possible solution for `PreviewNavBar.tsx`:
- Add state `targetedSection` in addition to just `activeSection`
- When the user clicks a button (`scrollToSection`) we not only try to scroll into view, but also set `targetedSection`
- On the `"scrollend"` event (see [MDN docs on scrollend](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollend_event)), set `activeSection` to `targetedSection`
  - Not sure how this works when you're already at the bottom and change your selection, but you can test this
  - scrollend is newly available in browsers, so old browsers might not work. However, it should be okay because adding an event listener for a nonexistent event won't cause issues; the bug will just still exist for those browsers.

Image

MajdKatabi:
## Issues

Closes #1137

### DIFF
--- a/site/src/component/ResultPreview/PreviewNavBar.tsx
+++ b/site/src/component/ResultPreview/PreviewNavBar.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { Button } from '@mui/material';
 import ArticleOutlinedIcon from '@mui/icons-material/ArticleOutlined';
 import BarChartIcon from '@mui/icons-material/BarChart';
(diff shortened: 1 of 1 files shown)


CadenLee2 (approved):
Tested:
- Clicking then scrolling
- Scrolling then clicking
- Clicking two items quickly

Everything works, logic looks reasonable

PR #1167, opened 2026-05-27
Review form validates terms / instructors

### LINKED ISSUE #1072: Review form should validate terms/instructors to prevent invalid reviews
(note: reopening from #747 since this was OOS and not addressed via #1042)

**Describe the bug**
When reviewing a course/instructor, the selected quarter and year do not necessarily determine the available instructors nor vice versa. If you enter an invalid combination (i.e. review an instructor for a term they did not teach), this will cause issues when attempting to edit this review later (see https://github.com/icssc/peterportal-client/pull/679#issuecomment-2874341227, but note this may be outdated to the behavior of the current form; may edit this issue later for specific cases)

**To Reproduce**
(here is one instance)
1. Review CS 175 with an invalid term and instructor (i.e. Monica Ann Alfaro Welling in Spring 2024)
2. Attempt to edit the review and observe that no year or quarter are present

(sidenote: probably to be filed as a separate issue, but attempting to edit the year also does not work and just resets it to the current selection)

**Expected Behavior**
Selecting a quarter + year should filter the available instructors for the course, and selecting an instructor should filter the available quarter + years. The solution may be more nuanced than this, however.

**Actual Behavior**
Invalid reviews shouldn't be allowed to be submitted in the first place

**Additional context**
TBD once investigated further

jchau06:
## Description
This PR addresses a bug where users can review an invalid combination of a term and an instructor, causing significant issues when trying to edit the review later. Now, selecting a quarter + year / year should filter the available instructors for the course or the available courses based on instructors. **Users can still change the year and quarter after filtering, allowing users to select courses or instructors from different terms.** On personal reviews, users can only adjust the term of the selected professor + course taken.

- useEffects fetch instructor data based on course and term / fetch course based on instructor data and term.
- Course / instructor data gets refetched whenever users change the term selected
- Filter works in every review context (reviewing professors, courses, personal reviews, and editing reviews)
## Screenshots
Before (reviewing professor):
image

After (reviewing professor):
image

Before (reviewing course):
image

After (reviewing course):
image

Before (on Your Reviews page):
image

After (on Your Reviews page):
image

## Test Plan

1. Open the review form for a course 
2. Input the year / year + quarter and see if instructors have been filtered.
3. Open the review form for a professor
4. Input the year / year + quarter and see if courses have been filtered.
5. Open the edit review form on Your Reviews page
6. Observe if available terms have been filtered based on review's professor and course. 
7. Observe if filtering is consistent when editing review for course and professor. 
## Issues

Closes #1072


site/src/component/ReviewForm/ReviewForm.tsx
```diff
@@ -81,38 +79,174 @@ const ReviewForm: FC<ReviewFormProps> = ({
...
+    if (yearTaken && availableTermsForProfessor.length > 0) {
+      const validQuarters = getQuarters(availableTermsForProfessor, yearTaken);
+      if (!validQuarters.includes(quarterTaken)) {
+        setQuarterTaken('');
+      }
+    }
+  }, [yearTaken, availableTermsForProfessor, quarterTaken]);
+
+  // clear instructor/course when year changes
+  useEffect(() => {
```
CadenLee2: Can we combine these into one useEffect that clears quarters, and then using the quarter+year combo clears professors?

When doing so also be careful about properly using setState (e.g. recall that quarterTaken won't have the new value yet after calling setQuarterTaken)

CadenLee2 (commented):
There's currently a bug where you can keep a bad year-quarter-prof combination if you:
1. Choose a valid prof, quarter, year combination
2. Change the year to *a year where the quarter and prof were both offered, but where the prof wasn't offered in that quarter* (see image below, where Pattis was offered in Spring 22 but not Fall 22, but this still doesn't get reset).
3. The same is true of changing quarters in general: you can change between Fall/Winter/Spring in the screenshot and it won't reset the prof
Image

Clearing the quarter and clearing the prof when you clear a year should be related and work together.
I think taking this into consideration could help tighten up the useEffects too (see comment)

site/src/component/ReviewForm/ReviewForm.tsx
```diff
@@ -81,38 +79,209 @@ const ReviewForm: FC<ReviewFormProps> = ({
...
+      });
+      termsToUse = Array.from(allTerms).sort();
+    }
+    // course context but NO instructor selected yet — show all terms from all instructors in that course
+    else if (courseProp && !instructor && Object.keys(professorTermsMap).length > 0) {
+      const allTerms = new Set<string>();
+      Object.values(professorTermsMap).forEach((terms) => {
+        terms.forEach((term) => allTerms.add(term));
+      });
+      termsToUse = Array.from(allTerms).sort();
```
CadenLee2: I'm seeing a lot of duplication here. The only differences in all three statements are:
1. The conditions of the ifs
2. Which terms map is being added to allTerms

This is a pretty satisfying opportunity to practice refactoring. The way I'd approach this is:
1. Extract out the repeated logic of constructing, populating `allTerms`, and converting it to a sorted array, based on a given term map. We can call this `uniqueTermsSorted`, and that helps us clear out all the repeated `allTerms ... Object.values(<something here>).forEach ... .sort();` code
2. The nested forEach for populating the set can be simplified down to simply A) flatten the values, B) turn it into a set, C) turn it back into an array and sort it. We get much more readable and shorter code like this (note that this means we don't need comments anymore!)
```ts
  useEffect(() => {
    const uniqueTermsSorted = (chosenMap: Record<string, string[]>): string[] => {
      const termsFromMap = Object.values(chosenMap).flat();
      const allTerms = new Set(termsFromMap);
      return Array.from(allTerms).sort();
    }

    let termsToUse: string[] = [];
    if (courseProp && instructor) {
      termsToUse = uniqueTermsSorted(professorTermsMap);
    } else if (professorProp && course) {
      termsToUse = uniqueTermsSorted(professorCourseTermsMap);
    } else if (courseProp && !instructor && Object.keys(professorTermsMap).length > 0) {
      termsToUse = uniqueTermsSorted(professorTermsMap);
    } else if (reviewToEdit && !courseProp && !professorProp) {
      // personal context: show only terms for this specific instructor + course
      termsToUse = professorTermsMap[reviewToEdit.professorId] ?? [];
    }
    // fallback
    else {
      termsToUse = terms;
    }

    setAvailableTermsForProfessor(termsToUse);
  }, [instructor, course, courseProp, professorProp, professorTermsMap, professorCourseTermsMap, terms, reviewToEdit]);

```
3. Notice that the first `if` and the second `else if` have the *exact* same contents. We can use some logic (thanks prof gassko) to simplify down the two conditions into one
  - `courseProp && instructor`, or `courseProp && !instructor && Object.keys(professorTermsMap).length > 0`
  - As one statement: `(courseProp && instructor) || (courseProp && !instructor && Object.keys(professorTermsMap).length > 0)`
  - Simplified: `courseProp && (instructor || Object.keys(professorTermsMap).length > 0)`

This gives us the final refactored code of:
```ts
  useEffect(() => {
    const uniqueTermsSorted = (chosenMap: Record<string, string[]>): string[] => {
      const termsFromMap = Object.values(chosenMap).flat();
      const allTerms = new Set(termsFromMap);
      return Array.from(allTerms).sort();
    }

    let termsToUse: string[] = [];
    if (courseProp && (instructor || Object.keys(professorTermsMap).length > 0)) {
      termsToUse = uniqueTermsSorted(professorTermsMap);
    } else if (professorProp && course) {
      termsToUse = uniqueTermsSorted(professorCourseTermsMap);
    } else if (reviewToEdit && !courseProp && !professorProp) {
      // personal context: show only terms for this specific instructor + course
      termsToUse = professorTermsMap[reviewToEdit.professorId] ?? [];
    } else {
      termsToUse = terms;
    }

    setAvailableTermsForProfessor(termsToUse);
  }, [instructor, course, courseProp, professorProp, professorTermsMap, professorCourseTermsMap, terms, reviewToEdit]);
```

Note that as a refactor, none of the logic has changed at all, but it's a lot more readable. It's up to you whether you want to use this exact code that I wrote (I haven't fully tested it so be careful) or if you want to make your own improvements to the original code, but either way it definitely should be refactored.

site/src/component/ReviewForm/ReviewForm.tsx
```diff
@@ -81,38 +79,209 @@ const ReviewForm: FC<ReviewFormProps> = ({
...
+            try {
+              const professorData = await searchAPIResult('instructor', professorProp.ucinetid);
+              if (professorData?.courses[courseId]?.terms) {
+                const terms = professorData.courses[courseId].terms as string | string[];
+                const termsArray = Array.isArray(terms) ? terms : terms.split(',').map((t: string) => t.trim());
+                courseTermsMap[courseId] = termsArray;
+              }
+            } catch (e) {
+              console.error(`Failed to fetch course ${courseId} terms:`, e);
+            }
```
CadenLee2: Note (like in my other comment) that there's a LOT of duplicated code.

You should be able to follow similar steps as my other comment to:
1. Identify what's unique between both code (whether it's a course or prof, and how we get the ucinetid). This should go outside of your helper function
2. Make a helper function (in this case, that gets prof or course map data based on a netID). Call it so that the maps get updated properly

site/src/component/ReviewForm/ReviewForm.tsx
```diff
@@ -81,38 +79,209 @@ const ReviewForm: FC<ReviewFormProps> = ({
...
+            setInstructorName(instructor.name);
+            // Also populate professorTermsMap for this specific course
+            if (instructor?.courses[reviewToEdit.courseId]?.terms) {
+              const courseTerms = instructor.courses[reviewToEdit.courseId].terms as string | string[];
+              const termsArray = Array.isArray(courseTerms)
+                ? courseTerms
+                : courseTerms.split(',').map((t: string) => t.trim());
+              setProfessorTermsMap({ [reviewToEdit.professorId]: termsArray });
+              setProfessorTermsMap({ [reviewToEdit.professorId]: termsArray });
+            }
```
CadenLee2: Are the two duplicate lines intentional?

CadenLee2 (commented):
Functionality looks good now! I did notice that the code can be cleaned up significantly, so consider ways that it can be refactored; I've left one pretty long example just to show my thought process.

Typically, when you see a lot of duplicated logic or feel the need to document cases through code comments, it's a sign that significant refactors could make things a lot simpler

CadenLee2 (approved):
Went through the same test steps as before and everything seems to be working; code looks good!

PR #1169, opened 2026-05-28
Improved Grade Distributions

### LINKED ISSUE #1121: Improve grade distribution
Update the grade distribution to match the new designs. This may require some new logic.

## Design
Reference the Figma for the full design. Note that the design is not 100% polished, so some details will be at the discretion of developers (the goal is consistency with existing components/designs). Places where you may need to fill in the gaps with reasonable assumptions:
- What to do in edge cases (e.g. there is no review for a certain prof, there are no tags)
- How to handle the exact "since last quarter" logic. If this is complex or difficult, we can leave it out for now and make an additional issue
- The overall "overview" for each professor shows upvotes/downvotes in the design, but we do not support that in our database, so that will be a future issue.
- For colors, we can update the MUI theme, or we can discuss if different colors would be preferred

Image

fiveminus1:
## Description

Implements the improved grade distributions views from #1121. This includes new cards for average GPA, quality, and difficulty, but does not include the instructors view. The pie chart is now removed, and we're also replacing nivo charts with recharts, since nivo seems to be poorly maintained.

RFC on mobile display, loading skeletons, and light mode styling.

## Screenshots

Old|New
:-:|:-:
image|image

## Test Plan

- [ ] Accurate data
- [ ] Selections work 
- [ ] Styling's good 

## Issues

Closes #1121


site/src/component/GradeDist/Chart.tsx
```diff
@@ -14,127 +13,37 @@ interface ChartProps {
...
+    { grade: 'F', count: aggregateGradeData.gradeFCount, fill: getCssVariable('--mui-palette-chart-f') },
+    { grade: 'P', count: aggregateGradeData.gradePCount, fill: getCssVariable('--mui-palette-chart-pass') },
+    { grade: 'NP', count: aggregateGradeData.gradeNPCount, fill: getCssVariable('--mui-palette-chart-noPass') },
+  ];
+
+  return (
+    <ResponsiveContainer width="100%" height="100%">
+      <BarChart data={data}>
+        <XAxis dataKey="grade" />
+        <Tooltip
```
CadenLee2, line 35: Somewhat odd tooltip behavior: it always spawns in from the top left corner then moves toward your mouse, meaning if you quickly mouse out then back into the chart, it flickers/teleports.
- If there's a way to easily fix it, let's go with that
- Otherwise, probably just disable the animation so it always instantly follows the user's mouse

CadenLee2 (commented):
Reminder of changes discussed in our meeting. Hopefully these are each pretty small:
1. Change "Winter 2026" to "W26" for space to prevent the layout shift.
2. Change the comparison direction and make it a delta rather than absolute, so that in Winter we can see "red down 0.2 from Fall" which means Winter's GPA was 0.2 less than Fall's

    - Image

3. Possibly change the background color in light theme. #1149 will add a better overlay3 color once it's merged, so perhaps we can go with that.
4. Change all of these emojis to the corresponding icons. We might as well change this on the professor preview page too; that allows us to remove the `react-twemoji` dependency! If you'd like, we can put this off for a different PR

    - Image

5. Consider running an accessibility check (I believe we were talking about the chart color contrast here)
6. Make the grade distribution chart + three info stats on the left a fixed height. This means that case 1) the down/up arrows are there, and case 2) the down/up arrows are not there, should have the same height.

UI/UX will work on a mobile redesign at some point (indeterminate), but in the meantime on small screens we should do a small fix to get
  - 1) instructor and quarter selectors either full width or on the same row, and
  - 2) these three statistics on the same row if possible. That might look awkward though so we could also consider putting them in a 2x2 grid with one slot empty, or having the GPA be full width above then the quality/difficutly in a row below it. Message me if this description is confusing lol.
Image

site/src/component/GradeDist/Chart.tsx
fiveminus1: Seems to be a known issue with custom `content` and I don't care enough for the animation, so we're disabling it

fiveminus1:
@CadenLee2 

> Possibly change the background color in light theme. https://github.com/icssc/peterportal-client/pull/1149 will add a better overlay3 color once it's merged, so perhaps we can go with that.

Used overlay2 to match reviewcards. Let me know what you think

> Change all of these emojis to the corresponding icons. We might as well change this on the professor preview page too; that allows us to remove the react-twemoji dependency! If you'd like, we can put this off for a different PR

Let's do this in another PR since it'll remove a dependency and span throughout some other components

> Consider running an accessibility check (I believe we were talking about the chart color contrast here)

I don't remember if we said we wanted the colors to be the same throughout the modes, but I changed up the colors to something that should be more WCAG-compliant on the contrast for both light/dark overlay2 card backgrounds. If we can separate it into light/dark again maybe we can brainstorm on the colors a little more, but if you like the current colors then we can keep them.

CadenLee2 (commented):
Generally looks good.

Responding to your comments:
> Used overlay2 to match reviewcards. Let me know what you think

Sounds good

> Let's do this in another PR since it'll remove a dependency and span throughout some other components

That sounds good with me; we can do the nivo removal and the icon switch in another PR (issue: #1201)

> I changed up the colors to something that should be more WCAG-compliant on the contrast for both light/dark overlay2 card backgrounds

These look good to me!

As for things to fix:
- Specifically on this screen size (small enough but not mobile) things get cooked
  - Image
- On some classes where the data can't be found, it shows an infinite loading animation rather than a proper error message
  - image

Fixing these should be quick, so I can help do that soon

anthonyj33 (approved):
LGTM! Everything we discussed has been resolved. This is a huge improvement over the current version, very cool 🎉

PR #1170, opened 2026-05-29
Add PostHog Events

### LINKED ISSUE #1159: Integrate PostHog custom events
In much the same way that AntAlmanac Scheduler maintains a list of custom events and sends those to PostHog, Planner should also use custom events so we can improve our analytics for specific features.

We should include a minimal set of events (more can be integrated after this change is merged):
- Create custom card
- Import a transcript
- Save a roadmap while logged out

We should follow a very similar implementation and structure to Scheduler. For now, we can keep autocapture on, but will likely disable it once we get better custom event coverage.

Ascent817:
Add custom PostHog events like Scheduler does, so we get better analytics than from just Autocapture.

## Description

Add analytics.ts to enumerate and manage logging of custom events
// TODO: More to come

## Test Plan

Verify that all custom events fire and are visible from PostHog dashboard.

## Issues

Closes #1159


Ascent817:
We hit the usage limits for our PostHog organization. I confirmed that the custom events work on my own development PostHog organization. We'll probably need to turn off Autocapture to be able to use custom events like Scheduler does without exceeding our limits every month.

PR #1173, opened 2026-05-30
Toolbar height navbar fix

### LINKED ISSUE #1158: Fix course preview toolbar height with navbar
Due to the padding of the preview toolbar, when the navbar is present on desktop it becomes too tall and looks inconsistent with the left sidebar.

Reduce the padding so that it can match the Figma design (below)
## Screenshots
What it currently looks like:
Image

Versus what it should look like:
Image
To correspond to the Figma:
Image

jchau06:
## Description
This PR makes a change to the padding of the toolbar when the navbar is present on desktop. Previously, the navbar changed the height of the toolbar, making it inconsistent with the left section's toolbar. Now, on both sides, the toolbars match heights. 

1. Changed padding of header in full screen.
2. Added media query to prevent vertical stacking issues at certain width range. 
## Screenshots

Before: 
image

After: 
image

## Test Plan

1. Search for a course.
3. Observe that in full-screen / desktop screen size, the toolbars on each side match in height.

## Issues

Closes #1158


site/src/component/ResultPreview/CoursePreview.tsx
```diff
@@ -31,6 +31,7 @@ interface PreviewTitleProps {
 const PreviewTitle: FC<PreviewTitleProps> = ({ isLoading, courseId, courseData }) => {
   const wrapContent = (content: ReactNode) => <p className="preview-title">{content}</p>;
   const shortenText = useMediaQuery('(max-width: 480px)');
+  const hidePreviewingText = useMediaQuery('(min-width: 800px) and (max-width: 860px)');
```
CadenLee2: 1. `shortenText` already exists here and handles the case of hiding "Previewing " properly, so we can just use that rather than adding a new change in the final return. This means either updating the one `shortenText` media query logic to include an or, or doing two separate media queries and or'ing them.
2. The hardcoded `800px` is duplicated because we already have mobile logic elsewhere. This can be removed by using the `useIsMobile` hook. So you should be able to do something like this:
```ts
  const isMobile = useIsMobile();
  const shortenTextMobile = useMediaQuery('(max-width: 480px)');
  const shortenTextDesktop = useMediaQuery('(max-width: 860px)');
  const shortenText = isMobile ? shortenTextMobile : shortenTextDesktop;
```

site/src/component/ResultPreview/CoursePreview.tsx
```diff
@@ -46,7 +47,12 @@ const PreviewTitle: FC<PreviewTitleProps> = ({ isLoading, courseId, courseData }
     return wrapContent(formattedCourseId);
   }
 
-  return wrapContent(<>Previewing {formattedCourseId}</>);
+  return wrapContent(
+    <>
```
CadenLee2: This change should be unnecessary if you implement the above

CadenLee2 (commented):
The height fix works, I have a code suggestion for hiding the previewing text

CadenLee2 (approved):
Testing still works, code changes address comments.

LGTM!

PR #1174, opened 2026-05-30
Saved library toast

### LINKED ISSUE #1161: Saved to library toast
When the user saves a course to their library from a preview (i.e. top right save button), they get no indication of what this did. As the library is relatively hidden, we have two choices:
1. Automatically navigate to the library tab
2. Show a toast indicating where the course was saved to.

Choice 1 may be obtrusive, especially if someone is trying to save many courses from a requirements list, for instance, so we'll go with choice 2.

When a course is saved (but not when it's unsaved), the toast should say: "Saved to your Catalog's Library". Consider making this toast *on click* navigate to the library tab; this may take some discussion to decide.
Image

jchau06:
## Description
This PR addresses an issue where a user can press the save / bookmark buttons and the user would not get an indication of what it did (saved the course to the library). This PR shows a toast indicating that the course has been saved to their catalog's library. The toast is also clickable and will navigate the user directly to the library tab, showing their saved courses.  

1. Extended redux state for toast actions in roadmapslice.ts
2. Updated toast component to be clickable + integrated ClickableDiv
3. Added toast into roadmap page
4. Integrated toast dispatch into CourseBookmarkButton, allowing toast to appear when the bookmark buttons are clicked

**Edit:** This PR also fixes a bug involving CompletionHint. Before, CompletionHint called the useCompletionCheck hook but CompletionHint was a function, violating the rule of hooks. Re-rendering the component would cause the hook count to change and an error would arise. Now, CompletionHint is a component so hook count stays consistent between re-renders. 
## Recordings

https://github.com/user-attachments/assets/264174ce-a341-4194-8715-0ac490bf6f4b

https://github.com/user-attachments/assets/74a5049c-d903-45df-88b4-c14754415e77

## Test Plan

1. Search a course
6. Either press the bookmark icon on the course hit item or the save button on the top right.
7. Observe if a toast shows up.
8. Click on the toast to navigate to the library tab. 

## Issues

Closes #1161


site/src/app/roadmap/catalog/ProgramRequirementsList.tsx
```diff
@@ -371,7 +371,8 @@ const CourseRequirement: FC<CourseRequirementProps> = ({ data, takenCourseIDs, s
           {showLabel && (
             <p className="requirement-label">
               <b>
-                Complete {label} of the following{CompletionHint(data, takenCourseIDs)}
+                Complete {label} of the following
+                <CompletionHint data={data} takenCourseIDs={takenCourseIDs} />
               </b>
```
CadenLee2, line 376: Why was this file touched in the diff?

I saw the commit mentioning that this fixes a bug. If the bug is relevant to this PR then that's fine, but I do suggest mentioning it in the PR description

site/src/app/roadmap/RoadmapPage.tsx
```diff
@@ -50,6 +52,15 @@ const RoadmapPage: FC = () => {
 
   const previewDepth = usePreviewDepth();
   const handleCloseToast = () => dispatch(setShowToast(false));
+
+  // handle toast click based on action specified
+  const handleToastClick = () => {
+    if (toastAction === 'library') {
+      dispatch(setSelectedSidebarTab(1)); // go to CourseCatalog tab
+      dispatch(setSelectedTab('Library')); // go to library within CourseCatalog
```
CadenLee2: The comments seem unnecessary here since the logic is quite clear from the code (which is good!)

site/src/app/roadmap/catalog/ProgramRequirementsList.tsx
```diff
@@ -383,12 +384,31 @@ const CourseRequirement: FC<CourseRequirementProps> = ({ data, takenCourseIDs, s
...
-  if (showCourseCount) return ` • (${completedCount}/${data.courseCount})`;
-  if (showUnitCount) return ` • (${completedCount}/${data.unitCount} units)`;
+
+  if (showCourseCount)
+    return (
+      <>
+        {' '}
+        • ({completedCount}/{data.courseCount})
+      </>
+    );
```
CadenLee2, line 403: Note that typically small code style changes like this shouldn't be included in the PR unless they come from Prettier, or if again they fix a relevant issue.

If this is from Prettier it's fine, but otherwise I recommend restoring the changes to this section

CadenLee2 (commented):
Behavior looks great and code is generally good! I had a few small code hygiene concerns

site/src/app/roadmap/catalog/ProgramRequirementsList.tsx
jchau06: Updated the comment to describe the bug being fixed.

site/src/app/roadmap/catalog/ProgramRequirementsList.tsx
jchau06: This is both a Prettier formatted change and because CompletionHint is a component now, it has to return JSX instead of a string.

CadenLee2 (approved):
Code changes address comments, and the updated description looks good. Testing continues to work! 💵

PR #1177, opened 2026-06-02
MUI hydration fixes

CadenLee2:
Due to incomplete MUI + Next.js integration, we were seeing hydration errors and other warnings, including:
- `<style data-emotion ...>` errors
- `The pseudo class ":first-child" is potentially unsafe when doing server-side rendering`

We also sometimes experienced related errors with MUI autocompletes:
- `MUI: A component is changing the uncontrolled value state of Autocomplete to be controlled`

Note that this PR will require team members to rerun `pnpm install`

## Description

1. Followed the start of the [MUI + Next.js guide](https://mui.com/material-ui/integrations/nextjs/) to install `@mui/material-nextjs`, which puts necessary styles in the head rather than the body and prevents these kinds of issues. This also explicitly allows us to use `:first-child` in our theme where previously it raised the error. Note that AntAlmanac Scheduler has already followed this step, and thus our codebase is also becoming more consistent in the process.
2. Fixed autocomplete control logic

## Screenshots

| Before | After! |
| --- | --- |
| image | image |

## Test Plan

- [ ] Staging console should not show hydration or MUI-related errors
- [ ] Running locally should not show hydration or MUI-related errors

## Issues

Closes #

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -139,6 +139,9 @@ importers:
 
   site:
     dependencies:
+      '@emotion/cache':
+        specifier: ^11.14.0
+        version: 11.14.0
       '@emotion/react':
         specifier: ^11.14.0
         version: 11.14.0(@types/react@18.3.14)(react@18.3.1)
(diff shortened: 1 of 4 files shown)


anthonyj33 (approved):
LGTM! Finally getting rid of the hydration error :o

PR #1180, opened 2026-06-03
Custom Card Suggestion in Search

### LINKED ISSUE #1151: Add custom cards suggestion in search
When the user searches for something that doesn't exist (e.g. "antalmanac"), the no results screen should suggest that they add a custom course, perhaps with a button or link. When clicked, it will take them to the "library" tab of the catalog.

This should work on mobile as well

Image

MajdKatabi:
## Issues

Closes #1151


CadenLee2 (approved):
Works well on desktop and mobile, lgtm!
1. As for the button text, the feature is called "custom cards" rather than "custom courses". However, I do think saying "custom course" makes sense in this context because the user was specifically searching for courses.
2. For a future task, on mobile it's currently not possible to add a custom course to your roadmap. However, that'll be fixed when we get the new course-adding flow, which comes in the future

PR #1181, opened 2026-06-03
Reordering roadmaps

### LINKED ISSUE #1155: Reordering roadmaps
Scheduler supports reordering schedules, but Planner doesn't support reordering roadmaps.

## Requirements
Add a drag handle to the left of each roadmap like in Scheduler, and allow them to be rearranged

## Implementation
Consider how this will work with the current edit/revision system. You may need to create a new edit type and handle it properly on the frontend and backend (if this becomes complicated, reach out to a lead). Or, it may be simple enough to do with the existing edits.

| Scheduler | Planner (current) |
| --- | --- |
| Image | Image |

DerrickThrower:
Drag and drop feature for roadmap planners allowing users to reorder their roadmaps. 

## Description

Used react sortable for the roadmaps and styled a  grab handle for the cursor as seen with the drag handle. Created PlannerOrderEdit type to track changes in the planner order.

## Screenshots

https://github.com/user-attachments/assets/79413744-e66c-47df-ad8f-5a5bc31efd2c

## Issues

Closes #1155


site/src/helpers/sortable.ts
```diff
@@ -38,3 +39,10 @@ export const programRequirementsSortable: SortableOptions & Partial<ReactSortabl
   revertOnSpill: true,
   group: { name: 'courses', pull: 'clone', put: false },
 };
+
+export const planSortable: SortableOptions & Partial<ReactSortableProps<RoadmapPlan>> = {
+  animation: 150,
+  fallbackTolerance: 4,
```
CadenLee2, line 47: Any reason for not following the `...baseSortable` pattern here? The main difference is that this code won't have `baseSortable`'s `forceFallback: true` and `fallbackOnBody: true`. If you believe that those should be false, it may be more clear to explicitly write them in anyway; otherwise, you can just switch to using `...baseSortable` and remove these two properties

site/src/store/slices/roadmapSlice.ts
```diff
@@ -109,23 +109,29 @@ export const roadmapSlice = createSlice({
     // Modifying the Roadmap
 
     reviseRoadmap: (state, action: PayloadAction<RoadmapRevision>) => {
+      const currentPlannerId = state.plans[state.currentPlanIndex]?.id;
       const currentIndex = state.currentRevisionIndex;
       state.revisions.splice(currentIndex + 1, state.revisions.length, action.payload);
       restoreRevision(state.plans, state.revisions, currentIndex, currentIndex + 1);
       state.currentRevisionIndex++;
+      const newIdx = state.plans.findIndex((p) => p.id === currentPlannerId);
+      state.currentPlanIndex = newIdx !== -1 ? newIdx : Math.min(state.currentPlanIndex, state.plans.length - 1);
```
CadenLee2: Seeing duplication here. Extracting the logic from the reducers fully with `state` might be hard, but you can make a function that takes in the right parameters:
```ts
const findValidIndex = (plans: RoadmapPlan[], currentPlannerId: number, currentPlanIndex: number): number => {
  const newIdx = plans.findIndex((p) => p.id === currentPlannerId);
  return newIdx !== -1 ? newIdx : Math.min(currentPlanIndex, plans.length - 1);
}
```
and then call it here as well as in the other two places.

CadenLee2 (commented):
Behavior looks good! Tested:
- Undo and redo
- Saving and loading works properly
- Combinations of reordering, deletion, and adding new roadmaps

Two small code comments

site/src/helpers/sortable.ts
DerrickThrower: forceFallback and fallbackOnBody cause conflicts between React and SortableJS thus breaking the feature. I'll explicitly set these properties to false.

CadenLee2 (approved):
Testing continues to work, changes lgtm! 💵

PR #1188, opened 2026-06-12
Fix materials querying for summer

CadenLee2:
Anteater API has updated the low-cost course materials list, but summer materials are not showing as we query for the precise quarter (e.g. `Summer2`) rather than the generalized quarter name used on the spreadsheet (e.g. `Summer`)

## Description

All `*Summer*` quarters become `Summer` when querying. In the future we can consider making the enhancement of differentiating by section codes, but because materials are listed by professor, this is unlikely to make much of a difference.

## Screenshots

image

## Test Plan

- Go to the [affordable initiatives site](https://www.lib.uci.edu/affordable-initiatives/course-materials) and find a few courses with low-cost materials during a summer session. Search for those on staging and ensure they render properly

## Issues

Closes #


anthonyj33 (approved):
Tested on a few courses and code looks good!

PR #1189, opened 2026-06-12
Improve schedule of classes loading time

CadenLee2:
The schedule of classes previously fetched course materials after fetching term data, slowing down loading time when switching quarters

## Description

Now both requests happen concurrently

## Test Plan

- [ ] Switching between quarters in the schedule of classes for a course (incl. with low-cost materials, like ANTHRO 2A) should be slightly faster and should continue to work

## Issues

Closes #


anthonyj33 (approved):
LGTM 📈

Issue #1154, opened 2026-05-22
Support catalogue years: backend


CadenLee2:
See #1144 for context.

Add the following:
1. Store the catalog year associated with a user and their major or minor, in a new table. Figure out how we can best associate it with either a major or a minor that a user is taking, whether this is by maintaining a string or using some Postgres feature. Make the migration a **separate PR**, as usual (message lead about running the migration).
  - The default catalog year for a newly added major/minor should probably match the most recent year that's been offered; figure out how to do this. May be okay to handle a "lack of a specified catalog year" in the frontend instead by defaulting to the latest year.
  - The default catalog year for an existing user that doesn't have one yet should be determined based on the first year of their first roadmap
    - This might require making a script. Reference the other scripts in `api/scripts` and you can debug by printing the expected output but don't actually do modifications to the dev database until talking with a lead
2. Endpoints to set and retrieve the user's catalog years based on their majors. Possibly consider also changing the major endpoints to also return their catalog year? TBD (again, talk to lead)

PR #1190, opened 2026-06-16
Integrate catalog years into major/minor endpoints

anthonyj33:
## Description

Integrated catalog years into major and minor endpoints.

**Notes:**
- `getSavedMajors` includes `catalogYear?`.
- `saveSelectedMajors` takes `majors: [{ majorId, specializationId?, catalogYear? }]`.
- `getSavedMinors` includes `catalogYear?`.
- `saveSelectedMinor` takes `minors: [{ minorId, catalogYear? }]`.
- Saves replace the full user major/minor list.
- Missing or `null` catalog years are not stored.
- Removing a major/minor also removes its catalog year.

## Screenshots

N/A

## Test Plan

- Ran DB-backed tRPC test
- Verified empty reads for new/anonymous users
- Verified anonymous writes fail
- Saved majors/minors with and without catalog years
- Verified API responses and DB rows
- Verified re-save clears stale catalog-year rows
- Verified clearing selections removes catalog-year rows

## Issues

#1154

### DIFF
--- a/api/src/controllers/programs.ts
+++ b/api/src/controllers/programs.ts
@@ -6,7 +6,7 @@ import { publicProcedure, router } from '../helpers/trpc';
 import {
   MajorProgram,
   MajorSpecialization,
-  MajorSpecializationPair,
+  SavedMajorProgram,
   MinorProgram,
   ProgramRequirement,
   SavedMinorProgram,
@@ -30,8 +30,8 @@ const getAPIProgramData = async <T extends ProgramType>(programType: string): Pr
   return response;
 };
 
-const zodMajorSpecPairSchema = z.object({
-  pairs: z.array(
+const zodMajorProgramSchema = z.object({
+  majors: z.array(
     z.object({
       majorId: z.string(),
       specializationId: z.string().optional(),
@@ -41,8 +41,7 @@ const zodMajorSpecPairSchema = z.object({
 });
 
 const zodMinorProgramSchema = z.object({
-  minorIds: z.array(z.string()),
-  catalogYears: z
+  minors: z
     .array(
       z.object({
         minorId: z.string(),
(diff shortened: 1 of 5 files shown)


api/src/controllers/programs.ts
```diff
@@ -34,14 +35,35 @@ const zodMajorSpecPairSchema = z.object({
...
       specializationId: z.string().optional(),
+      catalogYear: z.number().int().nullable().optional(),
     }),
   ),
 });
 
 const zodMinorProgramSchema = z.object({
   minorIds: z.array(z.string()),
+  catalogYears: z
+    .array(
```
CadenLee2, line 45: We can get rid of the `minorIds` since this is redundant, and maybe rename `catalogYears` to `minors`; consider renaming majors too

api/src/controllers/programs.ts
```diff
@@ -34,14 +35,35 @@ const zodMajorSpecPairSchema = z.object({
...
+        minorId: z.string(),
+        catalogYear: z.number().int().nullable().optional(),
+      }),
+    )
+    .optional(),
 });
 
+function dedupeCatalogYears<T extends { catalogYear?: number | null }>(
+  rows: T[],
+  getId: (row: T) => string,
```
CadenLee2: Should be unnecessary due to schema (already unique on userId and major/minor)

types/src/courseRequirements.ts
```diff
@@ -32,4 +32,11 @@ export type ProgramRequirement<T extends ReqType = ReqType> = RequirementSchema
 export interface MajorSpecializationPair {
```
CadenLee2: Rename since it's no longer a pair: `SavedMajorProgram`

CadenLee2 (commented):
A few initial points on code, will test further later

CadenLee2 (approved):
Tested adding catalog years to major/minors; code looks correct and previous feedback is addressed.

Note that the `ReviewForm.tsx` change doesn't appear on local git diff so it seems to be a GitHub moment

PR #1194, opened 2026-06-18
Migrate to oxlint and oxfmt

### LINKED ISSUE #1192: Migrate to Oxlint and Oxfmt
[AntAlmanac Scheduler](https://github.com/icssc/AntAlmanac/pull/1522) has migrated to Oxlint and Oxfmt, and seen improved performance. For consistency and performance, Planner should do the same.

This should be done in two steps, as Scheduler did:
1. Update to oxlint/oxfmt
2. Format the entire codebase with oxfmt

We should ensure that oxlint properly runs with `pnpm lint` (and GitHub Actions and Husky use it properly). Eslint can be removed.

anthonyj33:
## Description

Migrated to Oxlint and Oxfmt instead of ESLint and Prettier, respectively. Updated lint and format scripts, which also updates GitHub actions workflows. Updated the config for lint-staged so that Husky behaves properly.

Ran oxfmt to format the codebase (literally only one change) and manually fixed oxlint complaints.

## Screenshots

No screenshots but wow its so fast now

## Test Plan

1. Run `pnpm format` and `pnpm lint`. Verify they run and that the codebase is compliant.
2. Check that there are no remaining references to ESLint or Prettier
3. Check that GitHub actions lint workflow works properly and that husky pre-commit hook works properly.

## Issues

Closes #1192


CadenLee2 (approved):
Just say "super fast"!

Tested:
- Husky hook
- Local lint and format (initially got an error from the local `.next` folder, but I think that was just out of date)

We can eventually update our `.ox*rc.json` configs to match Scheduler, but for now let's keep it as is so we don't mess up existing PRs.

Issue #1144, opened 2026-05-18
Support catalogue years


CadenLee2:
Anteater API was forced to venture through the perilous mines of [catalog years](https://github.com/icssc/anteater-api/pull/344), and is close to finished. It will soon be our turn to face the terror of catalog years.

## Context
Each school year (2023-24, 2024-25, etc.), requirements change. Students who entered the university in one year (say 2023) may have different requirements from students who entered in a different year (say 2026).

But it's not just based on the year you entered. For instance, the new CS requirements (check your email) are opt-in for current students who choose to update their DegreeWorks to a different catalog year. This means that students can *choose their catalog year*.

## Requirements
Bare minimum supported actions for a user:
- Choose a catalog year (presumably for your account, but maybe specific to a roadmap)
- See the requirements *for that year* for majors/minors/etc

Nice-to-haves:
- Have your catalog year inferred by the start year of your roadmap or by an imported transcript?

## Implementation
1. Add catalog year to db (tied to either account or roadmap, depending on what we decide)
2. Support loading/saving catalog year in the frontend
3. Consider where to place the catalog year UI, and how to best inform the user of what it's used for/what catalog years actually mean. Implement this UI

CadenLee2:
Closed by #1204

PR #1197, opened 2026-06-19
Catalog Years Frontend

### LINKED ISSUE #1153: Support catalogue years: frontend
Reference #1144 for overall context.

In the frontend, we need:
1. Logic to load the user's catalogue year preferences from their account (if logged in). This should load when they open the page, and be saved in a Redux slice. For now just set up this structure, since the actual loading will be a tRPC route in a future issue
2. The actual frontend UI. This is TBD; consider messaging UI/UX about their ideas once you finish part 1

Edit: it turns out users can have a different catalog year for each major. This means that we need to associate catalog years with majors, and store/load them as such

atile4:
## Description
- updated `courseRequirementsSlice` to include catalog years
- update endpoint in `getRequiredCourses` to include optional parameter `catalogYear`
- include catalog year in fetches for course requirements
- added dropdown for catalog year input
  - inputs for dropdown have options for the past 5 years, including current year

## Screenshots
image image

## Test Plan
- Verify that different catalog years have different course requirements for **majors**
  - Note: Art major has different course requirements between 2025-2026 and 2022-2023 under "1 ART HIS Course"
- Verify that different catalog years have different course requirements for **minors**
- Verify changes reflect in backend
- Verify a catalog year persists across reloads

## Issues

Closes #1153

See #1144 for context


site/src/app/roadmap/catalog/MajorCourseList.tsx
```diff
@@ -15,11 +16,31 @@ import trpc from '../../../trpc';
...
 
+// default catalog year
+const now = new Date();
+const year = now.getFullYear();
+const month = now.getMonth();
+
+// catalog year increases when SOC is released for the next fall quarter during spring
+const startYear = month >= 4 ? year : year - 1;
+
+const DEFAULT_CATALOG_YEAR = `${startYear}${startYear + 1}`;
```
CadenLee2: Would prefer to see the scope here isolated at least to a function (e.g. `getDefaultCatalogYear`) for organization and so we don't pollute the namespace

site/src/app/roadmap/catalog/MinorCourseList.tsx
```diff
@@ -5,24 +5,46 @@ import {
...
+const DEFAULT_CATALOG_YEAR = `${startYear}${startYear + 1}`;
+
+const CATALOG_YEAR_OPTIONS = Array.from({ length: 5 }, (_, i) => {
+  const startYear = parseInt(DEFAULT_CATALOG_YEAR.slice(0, 4)) - i;
+  const endYear = startYear + 1;
+  return {
+    value: `${startYear}${endYear}`,
+    label: `${startYear}-${endYear}`,
+  };
+});
```
CadenLee2: Notice that this code is duplicated between here and `MajorCourseList.tsx`. Consider adding a `helpers/programs.ts` file and extracting into it both 1) the function for getting the default catalog year, and 2) the catalog year options constant

CadenLee2 (commented):
Behavior looks correct, though after testing I have some thoughts:
1. When hovering the text "catalog year" we should have some kind of info box (see figma near the top left of the mobile tab)
2. Although at first I thought API's "just choose the closest catalog year" solution would work, the incorrectness is pretty unclear here (in the case of us not having 2026-27 data yet for the CS major). What we might want to do in the meantime is have a warning like the example below.

Current:
Image

Better: a warning when API's returned requirements don't match.

This can say something like "Warning: 2026-2027 requirements are not yet publicly available. Currently showing 2025-2026"
Image

Also some small code comments

site/src/app/roadmap/catalog/MajorCourseList.tsx
```diff
@@ -144,23 +186,74 @@ const MajorCourseList: FC<MajorCourseListProps> = ({ majorWithSpec, onSpecializa
...
+          <div className="catalog-year-warning">
+            <WarningAmberIcon className="warning-icon" />
+            <p className="catalog-year-warning-text">
+              {formatCatalogYear(majorWithSpec.catalogYear ?? DEFAULT_CATALOG_YEAR)} requirements are not yet publicly
+              available. Currently showing {formatCatalogYear(fallbackCatalogYear)}.
+            </p>
+          </div>
+        )}
+        {/* <p>{fallbackCatalogYear} hi hi {DEFAULT_CATALOG_YEAR}</p> */}
+
```
CadenLee2: can be deleted

site/src/app/roadmap/catalog/MajorCourseList.tsx
```diff
@@ -144,23 +186,74 @@ const MajorCourseList: FC<MajorCourseListProps> = ({ majorWithSpec, onSpecializa
...
   return (
     <div className="major-section">
       <ClickableDiv className="header-tab" onClick={toggleExpand}>
         <h4 className="major-name">{major.name}</h4>
         <ExpandMore className="expand-requirements" expanded={open} onClick={toggleExpand} />
       </ClickableDiv>
       <Collapse in={open} unmountOnExit>
+        <Tooltip
+          title="Major, minor, and GE requirements from a specific catalog year"
+          placement="bottom-start"
```
CadenLee2, line 221: 1. Currently the tooltip can block the selector until you move your mouse out. Can we make the tooltip disappear if you stop hovering the header? Probably with `disableInteractive`
Image

2. Also, based on the new designs (check peterportal-uiux) I believe the tooltip should be specific to the type of requirement. In this case it would say "Major requirements from a specific catalog year", and the same thing with minors

CadenLee2 (commented):
Looks solid, I've got a few comments/concerns:
- Determining the default catalog year feels a bit funny, especially given that the default can be a year when API doesn't actually have any majors with it. This is good for now (since we're not merging into main yet) but we may want to talk to API about getting the current scraped catalog year
- We may want to talk to UI/UX about the design and see if the new designs in peterportal-uiux are preferred (out of scope of this PR tho)
- See code comments

CadenLee2 (approved):
Lgtm! 💵

PR #1198, opened 2026-06-27
Review verification UI

### LINKED ISSUE #1175: Restore review verification UI
After #1065 , review cards no longer have UI for verification in the admin menu (https://antalmanac.com/planner/admin/verify). This can be resolved simply by adding a "Admin: delete" and "Admin: verify" option in the three-dots menu of the review cards when on this screen.

If you're working on this, make sure your email is added as an admin email in your local env. Internal members can also be added on prod

Image

jchau06:
## Description

This PR adds options to the three dots menu of unverified reviews on the "Verify Reviews" screen for admin users. Admins can now delete unverified reviews and verify unverified reviews.

1. Inside the ThreeDotsMenu component, the pathname and user admin status are used to determine if a user can view the verify reviews page. 
2. Added Verify and AdminDelete helpers to component. 
3. If in verify reviews pages as an admin, those options will render in the TreeDotsMenu. 

## Recording / Screenshots

https://github.com/user-attachments/assets/76851f5a-4c3d-4bfd-9fe2-b4fdcc7e9bf2

image

> [!NOTE]
> Options in this recording include Edit and regular Delete since these are also my reviews in this local environment. Upon testing with the shared DB, only Report, Admin: Delete, and Admin: Verify appear as expected. 

## Test Plan

1. Ensure that logged in email is an admin email.
2. Navigate to /planner/admin/verify to open verify reviews page or navigate through profile.
3. Check if Admin: Verify and Admin: Delete are options in ThreeDotMenu. 
4. Check if button behavior is valid (proper deletion / verification). 

## Issues

Closes #1175


site/src/component/Review/ReviewCard.tsx
```diff
@@ -88,13 +92,29 @@ const ThreeDotsMenu: FC<AuthorEditButtonsProps> = ({ review, course, professor }
...
+  };
+
   return (
     <>
       <IconButton onClick={handleMenuOpen}>
         <MoreVertIcon />
       </IconButton>
       <Menu anchorEl={anchorEl} open={open} onClose={handleMenuClose} className="review-menu">
         <MenuItem onClick={openReportForm}>Report</MenuItem>
+        {isAdmin && isAdminVerifyPage && <MenuItem onClick={() => verifyReview(review.id!)}>Admin: Verify</MenuItem>}
```
CadenLee2, line 114: `review.id` should never be undefined, can we get rid of the `!`? Same goes for line 142 further down in this file, it might've been an artifact of some previous time when the id might've been undefined

CadenLee2 (approved):
Tested this locally and looked through the code, almost everything looks good. Will debug the actions failure further (the postinstall works locally, action works locally when using `act`, and actions seem to work fine on other branches, will lyk what I come up with)

PR #1199, opened 2026-06-28
ci: pin node to 22.23.1 due to fetch bug

CadenLee2:
Due to a regression in Node.js fetch (see https://github.com/nodejs/node/issues/63989) the API types postinstall script fails on current PRs.

## Description

The faulty version is a security fix introduced in 24.17.0 (for us, 22.23.0), where `ERR_STREAM_PREMATURE_CLOSE` gets thrown incorrectly. It's fixed in 22.23.1, so this version is now pinned. In the future we can remove the change (we really should think about upgrading node)

## Screenshots

Before: 💀 
image

After: (check actions below)

## Test Plan

- [ ] This PR should pass checks properly without errors

## Issues

Closes #

### DIFF
--- a/.github/actions/setup-pnpm/action.yml
+++ b/.github/actions/setup-pnpm/action.yml
@@ -5,7 +5,7 @@ runs:
     - name: Install Node.js
       uses: actions/setup-node@v4
       with:
-        node-version: 22
+        node-version: 22.23.1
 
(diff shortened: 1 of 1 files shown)


atile4 (approved):
PR passed checks without issue LGTM 💸

PR #1200, opened 2026-07-03
Fixing unintuitive mobile reviews button

### LINKED ISSUE #1182: Unintuitive mobile reviews button
Mobile devices suffer from a hidden reviews button when you scroll to the bottom, due to the floating navbar.

Furthermore, the review button has always been hard to find (including on desktop) since it's at the very bottom of the page and requires scrolling if there are many reviews.

1. The "add a review" button should be moved to the top of the reviews section. Consider talking to UI/UX about where would be most visible and reasonable.
2. Also consider that, were the "add a review" to be moved to the top, the navbar could end up hiding parts of the bottommost review (see screenshot). Consider how to mitigate this

Image

Image

jchau06:
## Description

This PR changes the structure of the reviews section so that the add review button is at the top of the section beneath the dropdowns and checkbox. The PR also adds more space at the bottom of the result preview to account for the pill navbar potentially covering up the bottommost review with the removal of the review button at the bottom.

#### With Reviews:
image

image

image

#### Without Reviews:
image

image

image

> [!NOTE]
> Ignore the app dock hovering lol

## Test Plan

1. Navigate to a given course or professor.
2. Observe frontend changes to the placement of the button above the reviews.
3. Change screen sizes and observe the additional space given for the pill navbar.
4. Also observe frontend changes for a reviews section that does not have reviews. 

## Issues

Closes #1182


site/src/component/ResultPreview/ResultPreview.scss
```diff
@@ -96,6 +96,10 @@
     @media (max-width: globals.$mobile-cutoff) {
       padding: 16px;
     }
+
+    @container (max-width: 940px) {
```
CadenLee2: Because `940px` has been repeated in this file, we can extract it out to an SCSS variable at the top of the file (e.g. `$show-navbar-cutoff: 940px`) and use it in the relevant places.

Notes: 1) a regular CSS variable won't work here because this is a media query, 2) it seems you have to wrap the variable in `#{ }`, i.e. `#{$myvar}` when referencing it in the query

site/src/component/Review/Review.tsx
```diff
@@ -202,6 +202,18 @@ const Review: FC<ReviewProps> = (props) => {
                 label="Show verified reviews only"
               />
             </div>
+
+            <Tooltip
+              title="You must be logged in to leave a review"
+              placement="top"
+              open={isLoggedIn ? false : undefined}
+            >
+              <span className="add-review-button">
```
CadenLee2: Kinda funny that the tooltip comes up even when the region outside the button is hovered. This is because the span takes up the full width, so if you reverse the nesting and put the tooltip inside the span it should be fixed
Image

CadenLee2 (commented):
Works well when testing (logged in and out, mobile and desktop, with reviews and without)

CadenLee2 (approved):
LGTM!

PR #1204, opened 2026-07-27
Catalog years

### LINKED ISSUE #1154: Support catalogue years: backend

CadenLee2:
> [!NOTE]
> Wait to merge this until Anteater API has the requirements for 2026-27 for most majors

Support major and minor catalog years. This is the product of several PRs.

## Description

Catalog years are now associated with majors and minors. The year can be chosen by a dropdown in the catalog, which then displays the correct data from Anteater API.

## Screenshots

Screenshot 2026-07-15 172702

## Test Plan

- [ ] Catalog years load and save properly for accounts
- [ ] A warning shows properly when the year's data is not yet available
- [ ] UI works well

## Issues

Closes #1154 #1153 #1144


laggycomputer:
image

anthonyj33 (approved):
LGTM! Fixed specializations and did extra testing 💸

PR #1208, opened 2026-08-07
Fix invalid prerequisite structure warnings

anthonyj33:
## Description
When inspecting the prerequisite parsing logic, some prereq nodes are plain tree objects with no prereqType. Update the logic to account for this and avoid undesired warnings.

## Test Plan
Verify that prereqs still parse correctly. Confirm no errors in console.


site/src/helpers/planner.ts
```diff
@@ -529,8 +532,13 @@ const validatePrerequisites = ({ prerequisite, ...input }: ValidationInput<Prere
...
 
-  if (prerequisite.AND) return validateAndPrerequisite({ prerequisite, ...input });
-  if (prerequisite.OR) return validateOrPrerequisite({ prerequisite, ...input });
+  // some prereq nodes are plain tree objects with no prereqType
+  const prerequisiteTree = prerequisite as PrerequisiteTree;
+
+  if (prerequisiteTree.AND) return validateAndPrerequisite({ prerequisite: prerequisiteTree, ...input });
+  if (prerequisiteTree.OR) return validateOrPrerequisite({ prerequisite: prerequisiteTree, ...input });
+
+  if (isEmptyPrerequisiteTree(prerequisiteTree)) return new Set();
```
CadenLee2: TypeScript already knows `prerequisite` satisfies a `PrerequisiteTree` thanks to the if statement above, so you should be able to just use it directly:
```ts
  if (prerequisite.AND) return validateAndPrerequisite({ prerequisite, ...input });
  if (prerequisite.OR) return validateOrPrerequisite({ prerequisite, ...input });

  if (isEmptyPrerequisiteTree(prerequisite)) return new Set();
```

site/src/helpers/planner.ts
```diff
@@ -521,6 +521,9 @@ const validateOrPrerequisite = ({ prerequisite, ...input }: ValidationInput<Prer
   return new Set([unique.join('|')]);
 };
 
+const isEmptyPrerequisiteTree = (prerequisite: PrerequisiteTree) =>
+  Object.keys(prerequisite as Record<string, unknown>).length === 0;
```
CadenLee2: Unnecessary `as Record` assertion

CadenLee2 (approved):
LGTM

